PR #1, opened 2018-10-29
Fix Maxwell-Boltzmann Velocity function

Ellipse0934:
Replacing by the Normal distribution function in Distributions.jl for performance and accuracy improvement.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -2,3 +2,4 @@ julia 0.7
 StaticArrays
 ProgressMeter
 BioStructures
+Distributions
--- a/src/Molly.jl
+++ b/src/Molly.jl
@@ -3,6 +3,7 @@ module Molly
 using StaticArrays
 using ProgressMeter
 using BioStructures
+using Distributions
 
 using LinearAlgebra: norm, normalize, dot, ×
 
--- a/src/setup.jl
+++ b/src/setup.jl
@@ -298,8 +298,7 @@ end
 
 # Generate a random velocity from the Maxwell-Boltzmann distribution
 function maxwellboltzmann(mass::Real, T::Real)
-    norm_dist = sum(rand(12)) - 6
-    return abs(norm_dist) * sqrt(T / mass)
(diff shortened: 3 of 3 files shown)


Ellipse0934:
```
function maxwellboltzmann(mass::Real, T::Real) # Currently in Molly
    norm_dist = sum(rand(12)) - 6
    return abs(norm_dist) * sqrt(T / mass)
end
function NEWmaxwellboltzmann(mass::Real, t::Real) # Changed function
    return rand(Normal(0,sqrt(8.3144598*t/mass)))
end
```
Benchmarks: using @benchmark in BenchmarkTools
```
BenchmarkTools.Trial:  # Old Function 
  memory estimate:  176 bytes
  allocs estimate:  1
  --------------
  minimum time:     49.129 ns (0.00% GC)
  median time:      50.660 ns (0.00% GC)
  mean time:        61.214 ns (15.26% GC)
  maximum time:     44.944 μs (99.81% GC)
  --------------
  samples:          10000
  evals/sample:     991

BenchmarkTools.Trial: # New Function 
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     4.364 ns (0.00% GC)
  median time:      4.789 ns (0.00% GC)
  mean time:        4.806 ns (0.00% GC)
  maximum time:     8.139 ns (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     1000
```
Also, I think the current function's implementation might be wrong 

> return abs(**_norm_dist_**) * sqrt(T / mass)

By taking the absolute value of the normal distribution the resultant of the normalization of the three velocities does not end up in a Maxwell Distribution. (I began with studying MD with no Chemistry background 2 days ago so please forgive me if I made a huge blunder!)

screenshot from 2018-10-30 00-44-08
Blue: Current function and Green: Chi(3) scaled by a factor of sqrt(RT/M)
I also plotted against my function which seems to agree with the scaled Chi(3) and various other sources on the Internet.

jgreener64:
Great work, thanks for fixing this and improving the implementation.

PR #5, opened 2020-06-16
Performance improvements

SebastianM-C:
Hi!
I noticed that there some performance improvements that can be made. For example due to `atoms::Vector{<:Any}` a lot of dynamic dispatch was occurring inside force calculations. Adding another type parameter solved the problem and provided a ~40x speedup for Lenard-Jones interactions. I think further improvements can be made.

### DIFF
--- a/src/forces.jl
+++ b/src/forces.jl
@@ -82,6 +82,8 @@ function force! end
     fdr = f * dr
     forces[i] -= fdr
     forces[j] += fdr
+
+    return nothing
 end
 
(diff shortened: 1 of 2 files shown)


jgreener64:
This was a regression introduced in https://github.com/jgreener64/Molly.jl/commit/97ffa4c3dd3581b832d8cbcd17b2b94306cab5fd, thanks a lot for the fix.

Shows I need to do performance testing as part of CI. I'll look into a simple benchmark suite.

Issue #11, opened 2020-06-25
Improving tests


SebastianM-C:
I noticed that the current tests only check if the simulation doesn't error, but not if the results are wrong. Since I'm refactoring several internal components, I think I should be careful not to change the results. To this end I was thinking to check for some general properties of the simulation to ensure that it is correct.
I was thinking of checking for the energy of the system. Do you have any other ideas?

jgreener64:
Yes, this is something we need. Since it's not thoroughly tested I'm not fully confident in the implementation as is. Some ideas in addition to the energy would be:
- Temperature, when using a thermostat.
- Sanity checks of no atoms too close and no atoms outside the bounding box.
- Test the radial distribution function is sufficiently close to what is expected.
- RMSD, for molecular examples.

About the energy, I was thinking of implementing this as an `EnergyLogger`. So users would add `EnergyLogger(10)` to the `loggers` to record the energy every 10 steps. The `energy` function would be defined for the in-built forces. Users could define their own alongside their custom `force` function, but if they don't it would default to returning 0.0. `log_property!` for the `EnergyLogger` would loop over all the interactions, sum up the energies and store them. Let me know if you have any input on that idea.

SebastianM-C:
Regarding the energy we could look at the
- value of the energy or of the energy / particle
- energy error (`E(t)` - `E_initial`)

For both we could look at the time evolution or at the final value only. If we look at the energy value, we would need some results to compare with and I'm not sure where to look for those. For the energy error I can use the results from NBodySimulator as guidance, but I'd appreciate if you could give me some references. 

The problem that I am also facing with the solver benchmarks in https://github.com/SciML/DiffEqBenchmarks.jl/pull/106 is that the energy (in the case of a Lennard Jones potential) has a quite large variation.

There are also some [tests in the NBodySimulator.jl repo](https://github.com/SciML/NBodySimulator.jl/tree/master/test) which I could try to port to Molly.

jgreener64:
Indeed, testing stochastic systems is difficult. I don't know any references off the top of my head but will keep an eye out.

Fixing the starting coordinates and velocities should lead to fixed output to within numerical error when not using a stochastic thermostat, we can use that to some degree.

Porting tests for NBodySimulator or elsewhere sounds like a good idea.

jgreener64:
Energy, forces and simulation are now compared against OpenMM. Of course there is always more work to do on the tests.

PR #12, opened 2020-07-06
Add energy function

SebastianM-C:
I'd like to expand the discussion from https://github.com/JuliaMolSim/Molly.jl/issues/11#issuecomment-650128455 about the energy computation before implementing. One of the important aspects that should be taken into account if the cutoff behavior.
I was thinking to embed the cutoff in the interaction struct and create some types for different cutoffs. The interactions could then use the cutoff type as a type parameter and we can dispatch on each of them to implement the various cutoffs. This would also imply adapting the forces signatures. 

Regarding implementation details, `sqdist_cutoff_nb` should go inside the interaction structs, and the `coulomb_const` too. I'd also like to understand the purpose of the `min` in
https://github.com/JuliaMolSim/Molly.jl/blob/6c2e249646f6937572c89b743724ae22ba7ed7a7/src/forces.jl#L56

The hardcoded `100` value implicitly makes an assumption about the units being used. I think it would be nice to have a more unit independent approach (maybe even Unitful compatibility, but that's something to discuss in a separate issue).
I'm not sure if `nl_only` is related to this discussion.

On the topic of cutoffs, I implemented a simple potential shift in this PR, and I looked in the GROMACS documentation at the switch function references, but I'm not sure how to choose the parameters in the context of the Lennard Jones potential. They seem to focus more on the Coulomb potential. I've posted more about this in the #nbody slack channel.


jgreener64:
> I was thinking to embed the cutoff in the interaction struct and create some types for different cutoffs
Yes, that sounds like a good idea.

> Regarding implementation details, `sqdist_cutoff_nb` should go inside the interaction structs, and the `coulomb_const` too
Yes they should, this links in with the implicit units discussion too. I think I briefly tried putting them in and there was a performance hit, but the performance bottleneck has likely changed since then. I guess you'll benchmark it and see anyway.

> I'd also like to understand the purpose of the `min`
This is effectively a cutoff, but for the minimum rather than the maximum distance. The 12th power repulsion meant that if things got too close they flew off ridiculously fast. This caps the force at low distances. Ideally this wouldn't be required at all if one sets up the simulation correctly and uses a sensible timestep. It could be a parameter in the same way as the other cutoff parameters, or you could try removing it, though I believe at least some MD softwares have a low distance cutoff like this.

> I'm not sure if `nl_only` is related to this discussion
I don't think it is directly, but if you are using a neighbour list you would want to make sure that the potential beyond the neighbour list cutoff distance was 0 as these pairs will be ignored during calculation. Alternatively, the neighbour list could not be used for potential energy calculations, though this would be slow if the energy was being logged regularly during the simulation.

> but I'm not sure how to choose the parameters in the context of the Lennard Jones potential
I can't help you much off the top of my head, sorry, and can't weigh in much on the related discussion on Slack.

Regarding the code, this PR seems to be pointing in the right direction.

SebastianM-C:
Regarding code organisation, I was thinking to make a new `interactions` folder and use files for each interaction type and put there the `struct` and the `force!` and `potential_energy` implementations corresponding to the different cutoffs.
Do you think that's a good idea?

> I don't think it is directly, but if you are using a neighbour list you would want to make sure that the potential beyond the neighbour list cutoff distance was 0 as these pairs will be ignored during calculation.

Regarding the approach for computing the total energy, I think that using something similar to how forces are computed, would help. Maybe also integrate this in ParticleAccelerations in the future. This way we can dispatch on different summation algorithms for the energy too. From what I understand, the Ewald summation was used for energies initially.

> This is effectively a cutoff, but for the minimum rather than the maximum distance.

Should this cutoff be composable somehow with the other cutoff methods? I will try to see if we can make it optional with 0 cost (maybe a static if or something).

SebastianM-C:
Regarding the minimum distance cutoff, I'm not sure how to proceed in the potential energy calculation. The integration of the `min` function could be in principle done by dividing the integration domain and finding the solution of the `F(r) < f_min` equation.
I tried with Mathematica and the r_min would be `Root[-48 \[Epsilon] \[Sigma]^12 + 24 \[Epsilon] \[Sigma]^6 #1^6 + 
   p #1^14 &, 2]` 
Do you think adding additional dependencies to handle the root finding for the polynomial equations is worth the trouble?

jgreener64:
> Regarding code organisation, I was thinking to make a new interactions folder and use files for each interaction type and put there the struct and the force! and potential_energy implementations corresponding to the different cutoffs.
Do you think that's a good idea?

Sure, why not.

> Regarding the approach for computing the total energy, I think that using something similar to how forces are computed, would help. Maybe also integrate this in ParticleAccelerations in the future.

Yes, ideally this could be abstracted somewhat so one is either summing over forces or energies using the same algorithm. It might be complicated a bit by the different return values in each case.

> Regarding the minimum distance cutoff, I'm not sure how to proceed in the potential energy calculation.

I guess for the energy calculation the minimum cutoff can be removed, as one would want to see that very high energy to diagnose what went wrong (e.g. reduce the timestep or don't start the particles overlapping). It's more just a way to stop one force calculation ruining a whole simulation by giving a particle near infinite velocity if it gets too close to another.

SebastianM-C:
> Yes, ideally this could be abstracted somewhat so one is either summing over forces or energies using the same algorithm. It might be complicated a bit by the different return values in each case.

I've also thought about this, but the main problem is that the force calculation is in place and the energy one out-of-place. I don't see any solution here, other than maybe also have out-of-place force calculations.

jgreener64:
> I don't see any solution here, other than maybe also have out-of-place force calculations.

I have nothing against that in principle, in fact the differentiable version I am working on switched the API to an out-of-place force function to allow it to be broadcasted over.

SebastianM-C:
I that case, I think it would be useful to have both in and out of place kernels. This could also help with the future integration with DiffEq IIP/OOP problems.

SebastianM-C:
I'm trying to compare the results with the ones from `NBodySimulator`, but I get different values for the potential energy. I think that there is an error in my implementation somewhere as the energy seems to increase very rapidly during the simulation.

I also noticed that the values for `r2` seem to differ in some cases. I'll look further into this.

The code I'm using:
```julia
using BenchmarkTools, NBodySimulator
using StaticArrays
using Molly

using NBodySimulator: gather_bodies_initial_coordinates
using NBodySimulator: obtain_data_for_lennard_jones_interaction, pairwise_lennard_jones_acceleration!
using NBodySimulator: gather_accelerations_for_potentials, gather_simultaneous_acceleration

const T = 289.0 # K
const kb = 8.3144598e-3 # kJ/(K*mol)
const ϵ = T * kb
const σ = 0.34 # nm
const ρ = 100
const N = 4
const m = 10.0
const L = (m*N/ρ)^(1/3)#10.229σ
const R = 0.5*L
const v_dev = sqrt(kb * T / m)

function nb_setup(τ, t)
    bodies = generate_bodies_in_cell_nodes(N, m, v_dev, L)
    
    lj_parameters = LennardJonesParameters(ϵ, σ, R)
    lj_system = PotentialNBodySystem(bodies, Dict(:lennard_jones => lj_parameters));
    
    pbc = CubicPeriodicBoundaryConditions(L)
    t1 = 0.0
    t2 = t
    simulation = NBodySimulation(lj_system, (t1, t2), pbc, kb)
end

function molly_setup(τ, t)
    simulation = nb_setup(τ, t)
    u0, v0, n = gather_bodies_initial_coordinates(simulation)

    atoms = [Atom(mass=m, σ=σ, ϵ=ϵ) for i=1:N]
    coords = [SVector{3}(u0[:,i]) for i=1:N]
    velocities = [SVector{3}(v0[:,i]) for i=1:N]
    general_inters = (LennardJones(),)
    
    Simulation(
        simulator=Molly.VelocityVerlet(),
        atoms=atoms,
        coords=coords,
        velocities=velocities,
        temperature=T,
        box_size=L,
        loggers=Dict("temp" => TemperatureLogger(100)),
        timestep=τ, # ps
        n_steps=t/τ
    )
end

function acceleration(simulation)

    u0, v0, n = gather_bodies_initial_coordinates(simulation)

    acceleration_functions = gather_accelerations_for_potentials(simulation)
    simultaneous_acceleration = gather_simultaneous_acceleration(simulation)

    function soode_system!(dv, v, u, p, t)
        @inbounds for i = 1:n
            a = MVector(0.0, 0.0, 0.0)
            for acceleration! in acceleration_functions
                acceleration!(a, u, v, t, i);
            end
            dv[:, i] .= a
        end
        for acceleration! in simultaneous_acceleration
            acceleration!(dv, u, v, t);
        end
    end

    return soode_system!
end

τ = 0.5e-3 # ps or 1e-12 s
t = 10_000τ

s = molly_setup(τ, t)
nb_sim = nb_setup(τ, t)

Molly.potential_energy(s) # -10.447007725262868
res = run_simulation(nb_sim)
NBodySimulator.potential_energy(res, 0.) # -9.076902159226412
```

Note that I'm using a patched version of NBodySimulator to try and have the same cutoff behavior.

jgreener64:
Cool I'll see if I can spot anything, and have a look at the PR more broadly, later in the week or on the weekend.

SebastianM-C:
I think I've found the problem. The summing was different, I copied it over from the forces. I changed it so that it is `i < j`, but I think that there maybe a problem in `NBodySimulator` though. For the `N=4` config above and with the following in NBodySimulator
```julia
function lennard_jones_potential(p::LennardJonesParameters, indxs::Vector{<:Integer}, coordinates, pbc::BoundaryConditions)
    e_lj = 0
    n = length(indxs)
    @inbounds for ind_i = 1:n
        i = indxs[ind_i]
        ri = @SVector [coordinates[1, i], coordinates[2, i], coordinates[3, i]]
        for ind_j = ind_i + 1:n
            j = indxs[ind_j]
            rj = @SVector [coordinates[1, j], coordinates[2, j], coordinates[3, j]]

            (rij, r, r2) = get_interparticle_distance(ri, rj, pbc)

            if !(r2 < p.R2)
                #r2 = p.R2
                continue
            end

            σ_rij_6 = (p.σ2 / r2)^3
            @show (r2, σ_rij_6) (i,j)
            σ_rij_12 = σ_rij_6^2
            e_lj += (σ_rij_12 - σ_rij_6)
        end
    end

    return 4 * p.ϵ * e_lj
end
```
and
```julia
 @inbounds function potential_energy(inter::LennardJones,
                                    s::Simulation,
                                    i::Integer,
                                    j::Integer)
    T = eltype(s.coords[i]) # this is not Unitful compatible
    sqdist_cutoff_nb = inter.sqdist_cutoff_nb
    i == j && return zero(T)
    dr = vector(s.coords[i], s.coords[j], s.box_size)
    r2 = sum(abs2, dr)
    r2 > sqdist_cutoff_nb && return zero(T)

    if iszero(s.atoms[i].σ) || iszero(s.atoms[j].σ)
        return zero(T)
    end
    σ = sqrt(s.atoms[i].σ * s.atoms[j].σ)
    ϵ = sqrt(s.atoms[i].ϵ * s.atoms[j].ϵ)

    invr2 = inv(r2)
    six_term = (σ ^ 2 * invr2) ^ 3
    @show (r2,six_term) (i,j)    

    return 4ϵ * (six_term ^ 2 - six_term) #- cutoff_energy(inter, σ, ϵ)
end
```

For `Molly.potential_energy(s)`
```
(r2, six_term) = (0.1357208808297453, 0.617921766400001)
(i, j) = (1, 2)
(r2, six_term) = (0.1357208808297453, 0.617921766400001)
(i, j) = (1, 3)
(r2, six_term) = (0.2714417616594906, 0.07724022080000012)
(i, j) = (1, 4)
(r2, six_term) = (0.2714417616594906, 0.07724022080000012)
(i, j) = (2, 3)
(r2, six_term) = (0.1357208808297453, 0.617921766400001)
(i, j) = (2, 4)
(r2, six_term) = (0.1357208808297453, 0.617921766400001)
(i, j) = (3, 4)
```

while for
```julia
res = run_simulation(nb_sim)
NBodySimulator.potential_energy(res, 0.)
```
I get
```
(r2, σ_rij_6) = (0.1357208808297453, 0.617921766400001)
(i, j) = (1, 2)
(r2, σ_rij_6) = (0.1357208808297453, 0.617921766400001)
(i, j) = (1, 3)
(r2, σ_rij_6) = (0.1357208808297453, 0.617921766400001)
(i, j) = (2, 4)
(r2, σ_rij_6) = (0.1357208808297453, 0.617921766400001)
(i, j) = (3, 4)
```

@ChrisRackauckas I think that this may be of importance to the NBodySimulator benchmarks.

ChrisRackauckas:
Is the jitter expected?

SebastianM-C:
From what I understand `NBodySimulator` misses some interaction terms in the sum, I haven't figured out why yet. I suppose that this will impact the energy computation, but not sure exactly how. I'm continuing the investigations.

jgreener64:
Judging from the observation that it is the high `r2` pairs that are missed, it is because they are being skipped by
```julia
            if !(r2 < p.R2)
                #r2 = p.R2
                continue
            end
```

SebastianM-C:
Should this not be equivalent to
```
r2 > sqdist_cutoff_nb && return zero(T)
```
?

SebastianM-C:
wait, the cutoffs may be different

SebastianM-C:
Yeah, that's the problem. Thanks @jgreener64 
The cutoff distance in Molly is hardcoded to 1 by default. Do you think we should use a σ dependent value instead? I saw in the literature that 2.5-3.5σ was considered.

jgreener64:
Yeah that could work.

That hard-coded value should go, we discussed above having it as a parameter for the interaction but it would also make sense as a multiple of σ.

SebastianM-C:
Since the value of σ is dynamically computed, we can only store a coefficient. I've pushed the changes of the Lennard Jones case. If that's ok, I can try to extend it in similar ways for the rest of the potentials. This should be ready for review, but I will still keep it marked as draft since the changes are implemented only for the Lennard Jones potential and thus not yet ready to be merged.

src/interactions/lennard_jones.jl
```diff
@@ -0,0 +1,88 @@
...
+    r2 > sqdist_cutoff && return zero(T)
+    
+    invr2 = inv(r2)
+    six_term = (σ2 * invr2) ^ 3
+    # @show (r2,six_term) (i,j)
+
+    return 4ϵ * (six_term ^ 2 - six_term) #- cutoff_energy(inter, σ, ϵ)
+end
+
+function cutoff_energy(inter::LennardJones, σ, ϵ)
```
jgreener64, line 83: Should this dispatch on the cutoff type?

src/interactions/lennard_jones.jl
```diff
@@ -0,0 +1,88 @@
...
+    
+    if iszero(s.atoms[i].σ) || iszero(s.atoms[j].σ)
+        return zero(T)
+    end
+    σ = sqrt(s.atoms[i].σ * s.atoms[j].σ)
+    ϵ = sqrt(s.atoms[i].ϵ * s.atoms[j].ϵ)
+    
+    σ2 = σ^2
+    sqdist_cutoff = inter.sqdist_cutoff_coeff * σ2
+    r2 > sqdist_cutoff && return zero(T)
```
jgreener64, line 74: Looks like a good way to do it, did you notice much slowdown compared to the hard-coded `const` before?

src/cutoffs.jl
```diff
@@ -0,0 +1,5 @@
+export ShiftCutoff
+
+struct ShiftCutoff <: AbstractCutoff
+    limit_force::Bool
+end
```
jgreener64, line 5: Add EOL character.

src/interactions/lennard_jones.jl
```diff
@@ -0,0 +1,88 @@
+"""
+    LennardJones(nl_only)
+
+The Lennard-Jones 6-12 interaction.
+"""
+struct LennardJones{C, T} <: GeneralInteraction
```
jgreener64, line 6: `{T, C}` to match `T` going first elsewhere.

src/interactions/lennard_jones.jl
```diff
@@ -0,0 +1,88 @@
...
+    
+    σ2 = σ^2
+    sqdist_cutoff = inter.sqdist_cutoff_coeff * σ2
+    r2 > sqdist_cutoff && return zero(T)
+    
+    invr2 = inv(r2)
+    six_term = (σ2 * invr2) ^ 3
+    # @show (r2,six_term) (i,j)
+
+    return 4ϵ * (six_term ^ 2 - six_term) #- cutoff_energy(inter, σ, ϵ)
```
jgreener64, line 80: Uncomment before merge, and delete the above `@show`.

src/loggers.jl
```diff
@@ -1,112 +1,140 @@
-# Loggers to record properties throughout a simulation
```
jgreener64, line 1: Why does it diff the unchanged lines? Did the line endings change? Fine as long as they end up unix-style.

src/loggers.jl
```diff
@@ -1,112 +1,140 @@
...
+end
+
+"""
+    EnergyLogger(n_steps)
+
+Log the energy of the system throughout a simulation.
+"""
+struct EnergyLogger{T} <: Logger
+    n_steps::Int
+    energy::Vector{T}
```
jgreener64, line 81: `energies` would be more in line with the other loggers.

src/types.jl
```diff
@@ -28,6 +28,8 @@ Custom specific interactions should sub-type this type.
 """
 abstract type SpecificInteraction <: Interaction end
 
+abstract type AbstractCutoff end
```
jgreener64, line 31: Add a docstring. Perhaps export this type?

src/interactions/lennard_jones.jl
SebastianM-C: I think I should give it a better name. It is a constant term for the energy at the cutoff point. This is can be used by multiple cutoff types. For example the shift cutoff or a more advanced switch and shift one. The constant term would be the same for all, so that's why there's no need for additional methods for the energy at the cutoff point.

src/interactions/lennard_jones.jl
SebastianM-C: No, this only adds a multiplication operation and those are quite fast. Most of the slowdown comes from the `if`s and the distance computation. I have some optimization ideas for those.

SebastianM-C:
Thanks for the review. I'll address the comments.
Before moving on to the next potentials I'd like to get the tests passing for this. I would have expected the energy error to be small in the test simulation (maybe my 0.01 was too small), but getting [values of the order of 3e8](https://travis-ci.com/github/JuliaMolSim/Molly.jl/jobs/362535409#L436) seems problematic. Do you have any ideas about what could cause that?

src/interactions/lennard_jones.jl
jgreener64: Makes sense.

jgreener64:
That test simulation is in the constant NVT canonical ensemble: it uses a thermostat, which will change the energy.

You could duplicate that test and remove the thermostat. You'd probably also want to set up more reproducible starting coordinates, e.g. on a grid or randomly placed but with no close clashes, otherwise there might be stochastic test failures when measuring bulk properties.

I should add that my physical values for things like LJ parameters, temperature etc. were chosen rather arbitrarily for the tests and examples. If you have values that are more physically-sound then feel free to change the tests.

src/loggers.jl
SebastianM-C: There were a few places where the line endings were CRLF and I changed them to LF everywhere.

src/interactions/lennard_jones.jl
SebastianM-C: I used this form to simply dispatch on `LennardJones{ShiftCutoff}` instead of `LennardJones{T, ShiftCutoff} where T`.

SebastianM-C:
@jgreener64 I tried to use the parameters from https://aip.scitation.org/doi/full/10.1063/1.4726728 as a test case with the following code
```julia
using Molly
using Test
using Statistics

temp = 1.
timestep = 0.005
n_steps = 10_000
box_size = 50.0
n_atoms = 2000
m = 0.8 * box_size^3 / n_atoms

s = Simulation(
    simulator=VelocityVerlet(),
    atoms=[Atom(attype="Ar", name="Ar", resnum=i, resname="Ar", charge=0.0,
                mass=m, σ=0.3, ϵ=0.2) for i in 1:n_atoms],
    general_inters=(LennardJones(),),
    coords=[box_size .* rand(SVector{3}) for i in 1:n_atoms],
    velocities=[velocity(10.0, temp) .* 0.01 for i in 1:n_atoms],
    temperature=temp,
    box_size=box_size,
    loggers=Dict("coords" => CoordinateLogger(100),
                 "energy" => EnergyLogger(100)),
    timestep=timestep,
    n_steps=n_steps
)

E0 = energy(s)
@time simulate!(s, parallel=true)

ΔE = energy(s) - E0
@test abs(ΔE) < 1e-2

Es = s.loggers["energy"].energies
maxΔE = maximum(abs.(Es .- E0))
@test maxΔE < 1e-2

@test abs(Es[end] - Es[1]) < 1e-2
@test std(Es.-Es[1])/n_atoms < timestep^2
```

but I'm seeing a strange behavior. The initial energy (`E0`) is different from the energies in the logger. The energy in the logged values seems to conserve.

src/interactions/lennard_jones.jl
SebastianM-C: done

jgreener64:
The energy is logged every 100 steps so the first energy is logged at step 100. There is a drop off between the start and step 100 which can be seen if you log the energy every step.

This drop off is because randomly placing the atoms puts some too close. If you choose the coordinates with a minimum distance between atoms using something like
```julia
function placeatoms(n_atoms, box_size, min_dist)
    min_dist_sq = min_dist ^ 2
    coords = SArray{Tuple{3}, Float64, 1, 3}[]
    while length(coords) < n_atoms
        new_coord = rand(SVector{3}) .* box_size
        okay = true
        for coord in coords
            if sum(abs2, vector(coord, new_coord, box_size)) < min_dist_sq
                okay = false
                break
            end
        end
        if okay
            push!(coords, new_coord)
        end
    end
    return coords
end
```
and then use `coords=placeatoms(n_atoms, box_size, 0.5)`, the energy is conserved and the tests pass.

Perhaps something like the above function should go in the package, and perhaps there should be a step 0 purely for the loggers so that such issues can be found.

SebastianM-C:
I noticed that the `iszero(s.atoms[i].σ) || iszero(s.atoms[j].σ)` check affects the performance since it will be almost always false. There is a small performance gain if it is removed and in the case where it would apply, the zero propagates to the cutoff check anyway and thus the output will be the same.

Another check that could be elided would be the `i==j` one if the for loop bounds are correctly adjusted to `for j in 1:i-1`. Do you think that's a good idea in general?

jgreener64:
> There is a small performance gain if it is removed.

The motivation was for systems where a non-negligible fraction of the atoms have `σ` equal to 0, for example the hydrogens in some water models. In the simulation of macromolecules these hydrogens can account for ~50% of atoms. I think we should keep the check for these cases.

> Another check that could be elided would be the `i==j` one if the for loop bounds are correctly adjusted

It used to be like that, then I changed the looping bounds in order to allow self-interactions such as in [this example](https://juliamolsim.github.io/Molly.jl/latest/docs.html#Agent-based-modelling-1). This is a bit of a hack, I admit, and is a gotcha for people implementing their own `force!` functions. However, there should be a way to implement self-interactions. One way would be to have a `SelfInteraction` type to add to `GeneralInteraction` and `SpecificInteraction`. But that is probably a change for another PR. I am of course open to suggesions on that part of the API, and it may be relevant to your work on ParticleAccelerations.jl too.

SebastianM-C:
> In the simulation of macromolecules these hydrogens can account for ~50% of atoms.

Oh, I see. I should have more benchmarks to test. Do you have a test case for this situation?
I think that we could find a way to elide the check by using some tricks. I'll think about it.

> However, there should be a way to implement self-interactions.

I think that it would be better to just have a different summation algorithm in ParticleAccelerations.jl. That is one that allows self-interactions and one that doesn't and select the appropriate via multiple dispatch (this is similar to how there are pairwise symmetric and general pairwise algorithms). This should provide both features and performance.

jgreener64:
> Do you have a test case for this situation?

I think the peptide simulation in the tests should work, 64% of the atoms are water hydrogens with `σ` equal to 0.

> I think that it would be better to just have a different summation algorithm in ParticleAccelerations.jl.

Makes sense, in this case there could be a method that only sums over the self interactions. I think it's rare someone would want the same functional form for a self and a pairwise interaction, hence why the current setup is a bit of a hack.

SebastianM-C:
I redesigned cutoffs to apply more generally to potentials and used some metaprogramming to conditionally remove checks for performance. Let me know what you think. Is it too complicated? I tried to use optionally generated functions, so that one can more easily see the normal code. The obvious downside of this metaprogramming approach is that it's harder to debug and maintain. I'm not sure if this complicates things with autodiff tools.
Regarding performance, I tested the Lennard Jones potential with 300 particles, and I think all these optimizations cut the acceleration function evaluation from something like ~650 μs to ~420 μs. I will do some more rigorous benchmarks, but after we settle with the implementation (i.e. this PR gets merged)

If the implementations seems ok to you, the remaining things would be
- [ ] Soft sphere
- [ ] Harmonic bond
- [ ] Harmonic angle
- [ ] Mie

jgreener64:
I like the cutoff design.

The metaprogramming stuff looks quite complicated and could become a maintenance issue. I'm not necessarily against it, but I'd need more time to get into it and fully understand it.

Since this PR is getting rather large and contains a few different things, perhaps we should try and finish up the potentials then merge everything apart from the metaprogramming. That can be discussed separately, maybe in its own PR? I'm also aware that things in master may change under you, like with the Mie potential, and rebasing/merging can be a pain. Let me know if it's not easy to split out the metaprogramming.

SebastianM-C:
One very simple way to split it out would be to just have the `if @generated` branch commented out. The code should have the same semantics.

jgreener64:
My preference would be to not have significant commented out blocks on master; if you removed those branches they would still be in the git history.

SebastianM-C:
I resolved the Mie potential conflict with the master branch, I hope it's fine now.
I also noticed several optimizations that could be done, but I'll try to reserve that for a separate PR.

SebastianM-C:
It looks like the energy conservation test are pushing the test time over the allowed CI time (50mins)... I'm not sure how to proceed.
I could try to reduce the time or number of particles, but then it would not be that meaningful.
One other option would be to switch to GH actions, which give 2000 minutes / month (https://docs.github.com/en/github/setting-up-and-managing-billing-and-payments-on-github/about-billing-for-github-actions) and we can set an eventual time limit on our own (https://github.community/t/limit-runtime/16364).

And this is just with the LJ case, I didn't add tests for other potentials yet.

For this PR we can check the tests locally, and decide what to do about this later.

SebastianM-C:
Tests pass locally. If you have some references, I can also try to add more tests related to energy conservation.
Or we could just add minimal ones to check if the function doesn't error and leave it for another PR.

src/interactions/gravity.jl
```diff
@@ -0,0 +1,53 @@
...
+    mi, mj = s.atoms[i].mass, s.atoms[j].mass
+    params = (inter.G, mi, mj)
+
+    potential(inter, r2, inv(r2), params)
+end
+
+
+@fastmath function potential(::Gravity, r2, invr2, (G, mi, mj))
+    (G * mi * mj) * √invr2
+end
```
jgreener64, line 53: Add EOL character.

src/cutoffs.jl
```diff
@@ -0,0 +1,91 @@
+export ShiftedPotentialCutoff, ShiftedForceCutoff, NoCutoff
+
+struct NoCutoff <: AbstractCutoff end
+
+cutoff_points(::Type{NoCutoff}) = 0
+
+struct ShiftedPotentialCutoff{F, T} <: AbstractCutoff
+    min_force::F
```
jgreener64, line 8: This should be named `max_force`.

test/runtests.jl
```diff
@@ -102,6 +103,57 @@ end
     end
 end
 
+@testset "Lennard-Jones gas energy conservation" begin
```
jgreener64, line 106: If this test times out CI you could put it in a new file `runtests_long.jl` and just test it locally. We don't want to limit our testing due to CI usage concerns, which will probably (hopefully!) become more of an issue as the package grows.

SebastianM-C: I think that the tests could be restructured a bit. I was thinking we can group (in a separate PR) either by potentials or by functionality (i.e. energy conservation).

SebastianM-C:
This should be good to merge now.

jgreener64:
Great, merge it yourself.

PR #13, opened 2020-07-08
Langevin Thermostat

moradza:
I am not sure about my implementation, as I am new to Julia. But, I have implemented the LE thermostat. Let me know what you think?

### DIFF
--- a/src/thermostats.jl
+++ b/src/thermostats.jl
@@ -4,6 +4,7 @@ export
     NoThermostat,
     apply_thermostat!,
     AndersenThermostat,
+    LangevinThermostat,
     velocity,
     maxwellboltzmann,
     temperature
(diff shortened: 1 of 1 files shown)


jgreener64:
Thanks for doing this.

My understanding of Langevin dynamics is that the stationary Gaussian process makes it trickier to implement than just drawing from the normal distribution, see for example https://arxiv.org/abs/1212.1244. I think a good implementation would look something like equations 19 and 20 from that paper.

Also, Langevin dynamics is more akin to a simulator (that acts as a thermostat as well) than to just a thermostat. So I think this should be implemented as a `Simulator`.

moradza:
I just saw your roadmap, probably you have implemented the method in the paper.
What I have implemented is a simple version of the OpenMM Langevin Integrator. https://github.com/choderalab/openmmtools/blob/master/openmmtools/integrators.py 
In terms of accuracy, probably you are right. Actually, in this paper "https://www.ncbi.nlm.nih.gov/pmc/articles/PMC6208357/pdf/nihms965548.pdf" they have analyzed the error for the above implementation.

jgreener64:
I haven't got round to it, though it is on the todo list. When I do get round to it I'll merge this and modify from there, unless you or someone else does it first.

Those are useful references, thanks.

jarvist:
I think having a Langevin thermostat is useful, and the generalised Langevin equation is the base for more sophisticated thermostats that enable biasing the simulation to a particular behaviour (https://arxiv.org/abs/1204.0822 - colored noise thermostats). 
These model thermostats are particularly useful when you have very small systems / expensive evaluation of forces, such as when using ab-initio (from electronic structure calculation) forces. 

Maybe it's even worth having a 'thermostats' development issue, so that we can coordinate what we want to implement?

jgreener64:
Yeah a version of this PR could go in as a general Langevin implementation and then other flavours could be added.

> Maybe it's even worth having a 'thermostats' development issue, so that we can coordinate what we want to implement?
Sure, feel free to open one for discussion.

jgreener64:
Sorry for not dealing with this back in 2020. However I am now about to push a simple Langevin implementation based off the `LangevinMiddleIntegrator` in OpenMM and your implementation here.

With your permission @moradza I will add you as a co-author on that commit? I think that's easier than reviving this PR, as the package has moved on a lot since then.

moradza:
It is fine on my side! I don't remember much actually haha, let me know if you need any help!

jgreener64:
Thanks, I have done that now so will close this PR.

Issue #16, opened 2020-08-17
GPU compatibility


SebastianM-C:
I think it would be useful to discuss here the design changes required for working on GPUs.

@jgreener64, could you detail a bit on what's needed for working on GPUs?
I was wondering if [KernelAbstractions.jl](https://juliagpu.gitlab.io/KernelAbstractions.jl/) could help.

jgreener64:
Yeah it's a good time to discuss it, and in the context of the package more broadly. We want an API where users can plug and play their own components, yet is:
- fast on CPU
- works CPU parallel
- fast on GPU
- differentiable.

This is clearly a big ask, and under the hood we will likely have to have multiple implementations. I think the DiffEq ecosystem has two versions, in-place and out-of-place, of some algorithms.

Personally, I am willing to sacrifice *some* speed to achieve the above aims. I doubt we will ever be able to match the likes of GROMACS for speed, but if we are in the same ballpark and can offer a powerful and flexible API then we have a useful contribution.

My current thinking is to have two implementations, which would be used like this:

|              | Standard | Differentiable |
| ------------ | -------- | -------------- |
| CPU          | A        | B              |
| CPU parallel | A        | -              |
| GPU          | B        | B              |

A is the current implementation on master, which is already 2 implementations in the sense that it branches based on `parallel`, but we'll ignore that. B would be a broadcasted, out-of-place version that allocates more memory but works with Zygote and makes use of GPU broadcasting without custom kernels. Of course we could write GPU kernels for everything, but that would be another implementation and if we can get decent performance with broadcasting I'd prefer to do that.

I have some rough work on the `gpu` branch that runs but is slower than the main CPU implementation, currently by a factor of 5x on CPU and 2.5x on GPU. Profiling suggests the time is in broadcasting but I'll have to dig further and probably ask some people who know more about Julia on GPUs. It is nearly differentiable, I think there is one adjoint missing. Consequently, this could be the foundation for implementation B.

There are some changes to the force functions to make broadcasting work. We would probably have to make the corresponding changes to implementation A to make the APIs match, but I think it's possible without sacrificing performance.

KernelAbstractions.jl is a possibility though I don't know too much about it, or how it works with autodiff.

Let me know what you think of the above thoughts.

SebastianM-C:
I am in favor of the in-place and out of place APIs. I think this would go well with a future DiffEq integration (i.e. replace the simulators with ODEProblems).

[...]

PR #17, opened 2020-08-26
Switch to out of place force function and add a GPU/almost differentiable implementation

jgreener64:
Here's a prototype for the API changes that would allow two implementations under the hood, see discussion in https://github.com/JuliaMolSim/Molly.jl/issues/16, plus a first go at a GPU implementation with broadcasting. That implementation is almost differentiable using Zygote, and should be with a bit more work. The out-of-place `force` functions would be defined as shown for the LJ case. Current tests pass.

This rather convoluted function allows testing the various combinations:
```julia
using Molly # unify branch
using CUDA
using BenchmarkTools

n_atoms = 400
mass = 10.0
box_size = 6.0
temp = 1.0
starting_coords = placeatoms(n_atoms, box_size, 0.2)
starting_velocities = [velocity(mass, temp) for i in 1:n_atoms]
starting_coords_f32 = [Float32.(c) for c in starting_coords]
starting_velocities_f32 = [Float32.(c) for c in starting_velocities]

struct AtomMin{T}
    charge::T
    mass::T
    σ::T
    ϵ::T
end

function runsim(nl::Bool, parallel::Bool, gpu_diff_safe::Bool, f32::Bool, gpu::Bool)
    n_atoms = 400
    n_steps = 200
    mass = f32 ? 10.0f0 : 10.0
    box_size = f32 ? 6.0f0 : 6.0
    timestep = f32 ? 0.02f0 : 0.02
    temp = f32 ? 1.0f0 : 1.0
    simulator = VelocityVerlet()
    thermostat = NoThermostat()
    specific_inter_lists = ()

    neighbour_finder = NoNeighbourFinder()
    general_inters = (LennardJones(false),)
    if nl
        neighbour_finder = DistanceNeighbourFinder(trues(n_atoms, n_atoms), 10, f32 ? 1.5f0 : 1.5)
        general_inters = (LennardJones(true),)
    end

    if gpu
        coords = cu(deepcopy(f32 ? starting_coords_f32 : starting_coords))
        velocities = cu(deepcopy(f32 ? starting_velocities_f32 : starting_velocities))
        atoms = cu([AtomMin(f32 ? 0.0f0 : 0.0, mass, f32 ? 0.2f0 : 0.2, f32 ? 0.2f0 : 0.2) for i in 1:n_atoms])
    else
        coords = deepcopy(f32 ? starting_coords_f32 : starting_coords)
        velocities = deepcopy(f32 ? starting_velocities_f32 : starting_velocities)
        atoms = [Atom(attype="Ar", name="Ar", resnum=i, resname="Ar", charge=f32 ? 0.0f0 : 0.0,
                        mass=mass, σ=f32 ? 0.2f0 : 0.2, ϵ=f32 ? 0.2f0 : 0.2) for i in 1:n_atoms]
    end

    s = Simulation(
        simulator=simulator,
        atoms=atoms,
        specific_inter_lists=specific_inter_lists,
        coords=coords,
        velocities=velocities,
        temperature=temp,
        box_size=box_size,
        neighbour_finder=neighbour_finder,
        thermostat=thermostat,
        timestep=timestep,
        n_steps=n_steps,
        gpu_diff_safe=gpu_diff_safe
    )

    c = simulate!(s, parallel=parallel)
    return c
end

runs = (
    ("default"        , [false, false, false, false, false]),
    ("nl"             , [true , false, false, false, false]),
    ("parallel"       , [false, true , false, false, false]),
    ("implementation2", [false, false, true , false, false]),
    ("float32"        , [false, false, false, true , false]),
    ("gpu"            , [false, false, true , false, true ]),
    ("gpu_float32"    , [false, false, true , true , true ]),
)
```
The `gpu_diff_safe` argument to `Simulation` can be used to explicitly use one of the implementations, but the GPU implementation is used by default if `coords` look like a `CuArray` so it is not generally required to set it.

The difference per coordinate at the end of the simulation compared to the first set seems reasonable:
```julia
final_coords_ref = runsim(runs[1][2]...)
for (name, args) in runs
    final_coords = runsim(args...)
    final_coords_f64 = [Float64.(c) for c in final_coords]
    diff = sum(sum(map(x -> abs.(x), final_coords_f64 .- final_coords_ref))) / (3 * n_atoms)
    println(rpad(name, 17), diff)
end
```
```
default          0.0
nl               0.0
parallel         1.952142151632567e-17
implementation2  0.0
float32          1.4312343463763101e-5
gpu              3.6191535879304126e-16
gpu_float32      1.4258215708955277e-5
```
Benchmarking each implementation:
```julia
for (name, args) in runs
    time = @belapsed runsim($(args)...)
    println(rpad(name, 17), round(time, digits=4), " s")
end
```
```
default          0.3268 s
nl               0.0541 s
parallel         0.1517 s
implementation2  0.7934 s
float32          0.3557 s
gpu              0.0234 s
gpu_float32      0.0226 s
```

On the master branch the default implementation gives 0.27 s so we should hunt down that 20% performance regression (the same amount of memory is allocated). The GPU implementation currently counts each force twice but is ~10x faster than the default implementation.

Still needed for this to be merged:
- Sort out double counting for the GPU implementation, possibly using a triangular matrix [can wait].
- Use of neighbour lists for the GPU implementation [can wait].
- Fix, or at least identify, the performance regression for the CPU implementation [done].
- Make specific interactions work [done].
- Modify other potentials [done].
- Expand tests [done].
- Update docs [done].

Needed but could come later:
- GPU implementation for the energy logger.
- GPU implementation for thermostats.
- Make the GPU implementation differentiable, currently the `coord` arrays are mutated to make the views work.

@SebastianM-C let me know if you have any broad comments at this stage, I'll get further before asking for a full review.


jgreener64:
The Zygote dependency makes the build fail on Julia 1.0. That dependency won't be necessary once Zygote gets mutation support, as the `Zygote.ignore()` blocks can be removed.

src/interactions/coulomb.jl
```diff
@@ -21,48 +21,46 @@ Coulomb(nl_only=false) = Coulomb(
...
-                                    j::Integer) where C
-    i == j && return
-    dr = vector(s.coords[i], s.coords[j], s.box_size)
+@inline @inbounds function force(inter::Coulomb{C},
+                                    coord_i,
+                                    coord_j,
+                                    atom_i,
+                                    atom_j,
+                                    box_size) where C
+    coord_i == coord_j && return zero(coord_i)
```
SebastianM-C, line 30: This causes a performance drop as we now compare floating point numbers instead of integers.

src/forces.jl
```diff
@@ -85,6 +85,42 @@ function accelerations(s::Simulation; parallel::Bool=true)
...
+    forces = zero(coords)
+
+    for inter in values(s.general_inters)
+        if inter.nl_only
+            forces -= reshape(sum(force.((inter,), coords_is, coords_js, atoms_is, atoms_js,
+                                            s.box_size), dims=2), n_atoms)
+        else
+            forces -= reshape(sum(force.((inter,), coords_is, coords_js, atoms_is, atoms_js,
+                                            s.box_size), dims=2), n_atoms)
+        end
```
SebastianM-C, line 99: Is the if necessary `if` both branches do the same thing?

SebastianM-C (commented):
> Fix, or at least identify, the performance regression for the CPU implementation.

I looked at the performance differences with regard to master and most of it seems to be due to the replacement of the `i==j` check with the one on `coords`. I think the rest may be due to performing the additions when the result is 0, but I'm not 100% sure.

SebastianM-C:
> Sort out double counting for the GPU implementation, possibly using a triangular matrix.

I think this would solve the performance difference on the CPU too. I'll try to look for some solutions.

src/interactions/coulomb.jl
jgreener64: Okay I'll probably just remove it then, and make a note to add a `SelfInteraction` type so that a couple of the examples keep working.

src/forces.jl
jgreener64: Not currently, I intend to replace this with something that actually uses the neighbours though.

SebastianM-C: If we give the coords and the atoms to the function, isn't that equivalent to using a neighbor list?
The caller of the function could just use a part of the `i`s and `j`s and here we could just keep the implementation generic.

src/interactions/coulomb.jl
SebastianM-C: The problem with removing it would be in the out of place case, where we must take care what we sum.

jgreener64: True, I was thinking of passing a diagonal array of ones as an extra broadcasting argument. Then a wrapper function returns `zero()` if it sees a one, representing an i-i interaction, otherwise it calculates the force as normal.

src/forces.jl
jgreener64: There would have to be some way of summing the forces afterwards though, at the minute the 2D broadcast returns things in the correct shape for summing. I did think of doing something with a `SparseVector`, similar to the specific interaction case.

SebastianM-C:
I think that we can leave out performance optimizations in the summation and deal with them after we move this part in the future NBodySummations.jl package. I think that we could also reuse the same summation code for both forces and potentials.

src/types.jl
```diff
@@ -220,8 +220,7 @@ function Simulation(;
                     timestep=0.0,
                     n_steps=0,
                     n_steps_made=[0],
-                    # Guess whether we are on the GPU without depending on CUDA.jl
-                    gpu_diff_safe=startswith(string(typeof(coords)), "CuArray"))
+                    gpu_diff_safe=isa(coords, CuArray))
```
SebastianM-C, line 223: This seems better. The previous version was a bit hacky.

jgreener64: I read it's okay to depend on CUDA.jl on systems without a GPU, provided you don't try and move anything to the GPU.

jgreener64: Though depending on CUDA.jl means we can't support Julia versions before I think 1.3...

SebastianM-C: I think that having julia 1.3 as minimum requirement is reasonable. Recent versions of DiffEqBase also dropped support for earlier julia versions, so I think we are fine.

SebastianM-C:
With the latest changes the performance problems of the CPU versions are almost solved and the performance seems to be improved in some cases. Using the LJ and peptide tests, I get
```
# Lennard-Jones

@time simulate!(s, parallel=false)
# commit b7d87b241e8422284f96dcd2dec2b2a8b310f2e8 (before #5): 4.007033 seconds (151.04 M allocations: 6.070 GiB, 16.34% gc time)
# commit 92308291a316d58c9e5569369edde7b3d4624831 (v0.1.0): 0.179878 seconds (34.52 k allocations: 29.653 MiB)
# commit 45c8a78fce5491373131a6dc1049c55fee9d34df (current master): 0.145809 seconds (33.14 k allocations: 4.652 MiB)
# commit 8b2b220dc5630aeaecb57459655186a56dbd8c74 (unify): 0.159678 seconds (33.05 k allocations: 4.649 MiB)
# commit d93d51fef9f285f3e0048108b7cfa4624ca9506d (unify): 0.148450 seconds (33.26 k allocations: 4.656 MiB)

@btime Molly.accelerations($s, parallel=false)
# old version
n = find_neighbours(s, nothing, s.neighbour_finder, 0, parallel=false)
@btime Molly.accelerations($s, $n, parallel=false)
# commit b7d87b241e8422284f96dcd2dec2b2a8b310f2e8 (before #5): 3.022 ms (151773 allocations: 6.20 MiB)
# commit 92308291a316d58c9e5569369edde7b3d4624831 (v0.1.0): 123.643 μs (1 allocation: 2.50 KiB)
# commit 45c8a78fce5491373131a6dc1049c55fee9d34df (current master): 121.729 μs (1 allocation: 2.50 KiB)
# commit 8b2b220dc5630aeaecb57459655186a56dbd8c74 (unify): 134.209 μs (1 allocation: 2.50 KiB)
# commit d93d51fef9f285f3e0048108b7cfa4624ca9506d (unify): 114.391 μs (1 allocation: 2.50 KiB)
```

Which shows a performance improvement. In the case of the peptide tests
```
# Peptide

@time simulate!(s, parallel=false)
# commit b7d87b241e8422284f96dcd2dec2b2a8b310f2e8 (before #5): 105.474686 seconds (4.20 G allocations: 162.335 GiB, 16.98% gc time)
# commit 92308291a316d58c9e5569369edde7b3d4624831 (v0.1.0): 6.870994 seconds (1.77 M allocations: 535.806 MiB, 0.65% gc time)
# commit 45c8a78fce5491373131a6dc1049c55fee9d34df (current master): 7.695775 seconds (1.77 M allocations: 233.411 MiB, 0.35% gc time)
# commit 8b2b220dc5630aeaecb57459655186a56dbd8c74 (unify): 8.374999 seconds (14.49 M allocations: 691.883 MiB, 0.83% gc time)
# commit d93d51fef9f285f3e0048108b7cfa4624ca9506d (unify): 7.855441 seconds (14.49 M allocations: 691.860 MiB, 0.99% gc time)

@btime Molly.accelerations($s, parallel=false)
# old version
n = find_neighbours(s, nothing, s.neighbour_finder, 0, parallel=false)
@btime Molly.accelerations($s, $n, parallel=false)
# commit b7d87b241e8422284f96dcd2dec2b2a8b310f2e8 (before #5): 903.236 ms (41561843 allocations: 1.60 GiB)
# commit 92308291a316d58c9e5569369edde7b3d4624831 (v0.1.0): 49.141 ms (16351 allocations: 1.29 MiB)
# commit 45c8a78fce5491373131a6dc1049c55fee9d34df (current master): 57.982 ms (16351 allocations: 1.56 MiB)
# commit 8b2b220dc5630aeaecb57459655186a56dbd8c74 (unify): 63.042 ms (36908 allocations: 3.39 MiB)
# commit d93d51fef9f285f3e0048108b7cfa4624ca9506d (unify): 57.547 ms (36908 allocations: 3.39 MiB)
```

Using more atoms as in your above script shows that there are still some differences to master
# master
default          0.2891 s
nl               0.0512 s
parallel         0.1007 s
implementation2  0.289 s
float32          0.2993 s

# unify
default          0.3094 s
nl               0.0504 s
parallel         0.104 s
implementation2  0.6478 s
float32          0.3349 s
gpu              0.0784 s
gpu_float32      0.0372 s

I'm not sure why the speedup due to the GPU is smaller in my case.

jgreener64:
Thanks, those are useful benchmarks. I think this just needs tests and docs, hopefully done in a day or two.

> I'm not sure why the speedup due to the GPU is smaller in my case.

I have noticed inconsistent benchmark results on the GPU, or it could be the make of GPU - I am using GeForce GTX 1660 SUPER 6 GB for these tests.

SebastianM-C:
Oh, right. And also different CPUs would give different reference values, so the speedup is highly system dependent.
I have a GTX 1050Ti 4GB, so it's not as performant as your card, which explains the difference.

src/interactions/mie.jl
```diff
@@ -31,34 +31,14 @@ Mie(m, n) = Mie(m, n, false)
...
-        r2 > sqdist_cutoff && return zero(coord_i)
-
-        if r2 < activation_dist
-            f = force_nocutoff(inter, r2, inv(r2), params)
-        else
-            f = force_cutoff(cutoff, r2, inter, params)
-        end
-    end
-
+    f = force_nocutoff(inter, r2, inv(r2), params)
```
SebastianM-C, line 41: The Mie potential doesn't have cutoffs by definition?
I saw a cutoff in the original PR
https://github.com/longemen3000/Molly.jl/blob/79336da46c8271379a1240f1f2be54779084c332/src/forces.jl#L84
and that's why I added the cutoff part.

test/runtests.jl
```diff
@@ -193,7 +193,9 @@ end
         box_size=box_size,
         neighbour_finder=DistanceNeighbourFinder(nb_matrix, 10),
         thermostat=AndersenThermostat(10.0),
-        loggers=Dict("temp" => TemperatureLogger(10)),
+        loggers=Dict("temp" => TemperatureLogger(10),
+                        "coords" => CoordinateLogger(10),
+                        "energy" => EnergyLogger(10)),
```
SebastianM-C, line 198: Should we also add some energy conservation tests?
I'm not sure how well does it conserve in more complex cases.

test/runtests.jl
```diff
@@ -221,14 +223,43 @@ end
...
     )
 
     @time simulate!(s, parallel=false)
 end
 
+@testset "General interactions" begin
+    n_atoms = 100
+    #general_inter_types = (LennardJones(true), SoftSphere(true), Mie(5, 10, true), Coulomb(true),
+    general_inter_types = (LennardJones(true), SoftSphere(true), Coulomb(true),
+                            Gravity(10.0, true))
```
SebastianM-C, line 239: Are neighbor lists also used in these kind of simulations or `Gravity` is using neighbor lists just for testing purposes?

test/runtests.jl
```diff
@@ -221,14 +223,43 @@ end
...
         n_steps=n_steps
     )
 
     @time simulate!(s, parallel=false)
 end
 
+@testset "General interactions" begin
+    n_atoms = 100
+    #general_inter_types = (LennardJones(true), SoftSphere(true), Mie(5, 10, true), Coulomb(true),
+    general_inter_types = (LennardJones(true), SoftSphere(true), Coulomb(true),
```
SebastianM-C, line 238: Should we also test without neighbor lists given that there are a few atoms? Or those interactions are tested in other places already?

docs/src/differentiable.md
```diff
@@ -9,7 +9,7 @@ The concept of using automatic differentiation to obtain exact gradients through
 There are some projects that explore differentiable molecular simulations, such as [Jax, M.D.](https://github.com/google/jax-md) and [DiffTaichi](https://github.com/yuanming-hu/difftaichi), or you can write your own algorithms in something like [PyTorch](https://pytorch.org).
 However Julia provides a strong suite of autodiff tools, with [Zygote.jl](https://github.com/FluxML/Zygote.jl) allowing source-to-source transformations for much of the language.
 The [differentiable](https://github.com/JuliaMolSim/Molly.jl/tree/differentiable) branch of Molly lets you use the power of Zygote to obtain gradients through molecular simulations.
-It is not yet merged because it is experimental, untested, slow, liable to change, has a slightly different API to the main package, and only works for some parts of the main package.
+It is not yet merged because it is experimental, untested, slow, liable to change, and only works for some parts of the main package.
```
SebastianM-C, line 12: The code branch that would be used for Zygote is the one you called `implementation2`?

test/runtests.jl
```diff
@@ -1,13 +1,23 @@
...
+    @info "The parallel tests will be run as Julia is running on $(nthreads()) threads"
+else
     @warn "The parallel tests will not be run as Julia is running on 1 thread"
+end
+
+if CUDA.functional()
+    @info "The GPU tests will be run as CUDA is available"
 else
-    @info "The parallel tests will be run as Julia is running on $(nthreads()) threads"
+    @warn "The GPU tests will not be run as CUDA is not available"
```
SebastianM-C, line 20: I think we could talk with the JuliaGPU org at some point about testing.

src/interactions/mie.jl
jgreener64: Yeah just added it back.

jgreener64: Yeah just added it back.

test/runtests.jl
jgreener64: Probably longer term, though protein cases usually use a thermostat so energy won't conserve (though temperature should).

test/runtests.jl
jgreener64: Just for testing, though I now added the non-neighbour list tests too.

test/runtests.jl
jgreener64: Added them.

docs/src/differentiable.md
jgreener64: Eventually yes, the out-of-place implementation will work with Zygote, though it doesn't yet. At that point the experimental tag on the differentiable stuff can be taken off. The current `differentiable` branch is an older version of the out-of-place implementation that is very slow but does work with Zygote. So it's a third implementation on an experimental branch that will hopefully be deleted soon.

jgreener64:
Just added tests similar to the code above to check that the various implementations give the same results.

@SebastianM-C let me know if you have any more comments on this PR, thanks.

test/runtests.jl
```diff
@@ -221,14 +223,50 @@ end
...
 end
 
+@testset "General interactions" begin
+    n_atoms = 100
+    general_inter_types = (
+        LennardJones(true ), SoftSphere(true ), Mie(5, 10, true ), Coulomb(true ), Gravity(10.0, true ),
+        LennardJones(false), SoftSphere(false), Mie(5, 10, false), Coulomb(false), Gravity(10.0, false),
+    )
+
+    for gi in general_inter_types
```
SebastianM-C, line 242: You could add a `@testset "$gi"` in front of the `for` in order to be easier to tell which test fails.

SebastianM-C (approved):
This looks good to merge.

jgreener64:
Thanks for the review.

PR #40, opened 2021-06-27
Update adjust_bounds by removing iteration

ehgus:
The `adjust_bounds` is optimized into a one-line using the `floor` in base julia.
Here is the simple benchmark.

benchamarked julia version : 1.6.1

```julia
julia> @time adjust_bounds(90001.0,3.0)
  0.000029 seconds
1.0

julia> @time adjust_bounds_new(90001.0,3.0) # this is the function proposed
  0.000000 seconds
1.0
```


jgreener64:
Clear improvement, thanks.

PR #44, opened 2021-11-27
added velocity autocorrelation

chemicalfiend:
Added VelocityLogger to loggers.jl and vacf (velocity autocorrelation function) to analysis.

### DIFF
--- a/src/analysis.jl
+++ b/src/analysis.jl
@@ -5,7 +5,8 @@ export
     displacements,
     distances,
     rdf,
-    energy
+    energy,
+	velocityautocorr
 
 """
     visualize(coord_logger, box_size, out_filepath; <keyword arguments>)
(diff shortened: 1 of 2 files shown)


src/loggers.jl
```diff
@@ -98,6 +99,41 @@ function log_property!(logger::EnergyLogger, s::Simulation, step_n::Integer)
...
+end
+
+function VelocityLogger(n_steps::Integer, dims::Integer=3)
+	return VelocityLogger(typeof(one(DefaultFloat)u"nm * s^-1"), n_steps; dims=dims)
+
+end
+
+function Base.show(io::IO, vl::VelocityLogger)
+	print(io, "VelocityLogger{", eltype(eltype(vl.velocities), "} with n_steps
+	", vl.n_steps, ", ", length(vl.velocities), " frames recorded for",
```
jgreener64, line 127: Move closing quote to previous line.

src/analysis.jl
```diff
@@ -130,3 +130,10 @@ function potential_energy(s::Simulation)
 
     return uconvert(s.energy_unit, potential)
 end
+
+function vacf(vl::VelocityLogger)
```
jgreener64, line 134: Maybe the name `velocityautocorr` is clearer? Also, export the function and add a docstring.

src/analysis.jl
```diff
@@ -130,3 +130,10 @@ function potential_energy(s::Simulation)
...
+	n_atoms = length(first(vl.velocities))
+	
+	return (1/n_atoms)*(dot(first(vl.velocities), last(vl.velocities)))
```
jgreener64, line 137: Might be worth having arguments to the function to specify the starting and ending indices, e.g. `vacf(vl, 10, 20)` would do `dot(vl.velocities[10], vl.velocities[20])`. The defaults could be the first and last frame, i.e. `1` and `length(vl.velocities)`, so `vacf(vl)` gives the above behaviour.

jgreener64 (changes requested):
Looks great, thanks for making these changes. I suggested a few small things before merging.

src/analysis.jl
chemicalfiend: Done. I changed the name to velocityautocorr too (I made it vacf at first since that's the compute command in LAMMPS), this is a much better name.

jgreener64:
:+1:

PR #46, opened 2021-12-24
docs.md: Simulating gravity; Specify `box_size` in calling `visualize` method

hsugawa8651:


jgreener64:
Good catch, thanks for this.

PR #56, opened 2022-03-09
implement remaining AtomsBase API functions

### LINKED ISSUE #55: Inconsistency with AtomsBase interface
There are a couple things that the `System` implementation is missing for it to be fully interchangable with other `AtomsBase` implementations.

1. `atomic_symbol(::System, ::Integer)` is not implemented
2. `getindex(::System, ::Integer)` returns a struct with insufficient fields to be useful in the way that iterating over other `AbstractSystem` is useful. Specifically this is a problem in my `Atomistic` and `DFTK` integration code where I have to have a special case for a `System` versus an `AbstractSystem`. This could be resolved simply by having `getindex` return an [`AtomView`](https://github.com/JuliaMolSim/AtomsBase.jl/blob/master/src/atomview.jl) (though any current usage of `getindex` would have to be modified).

jrdegreeff:
fixes #55

### DIFF
--- a/src/types.jl
+++ b/src/types.jl
@@ -141,8 +141,6 @@ The mass of an atom.
 """
 mass(atom::Atom) = atom.mass
 
-AtomsBase.atomic_mass(atom::Atom) = mass(atom)
-
 function Base.show(io::IO, a::Atom)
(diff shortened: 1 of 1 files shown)


jgreener64:
Looks good, thanks.

PR #59, opened 2022-03-27
Update atom i&j partners correctly

Ruibin-Liu:
When deleting and adding a bond between atom `i` and `j`, the previous actions:
```Julia
delete!(atom_i.partners, atom_j.i)
delete!(atom_j.partners, atom_j.i)
```
and
```Julia
 push!(atom_i.partners, atom_j.i)
 push!(atom_j.partners, atom_j.i)
```
would get atom `i`'s partners correctly, but only delete/add itself from atom `j`'s partners. Instead, we should change as below:
```Julia
delete!(atom_i.partners, atom_j.i)
delete!(atom_j.partners, atom_i.i)
```
and
```Julia
 push!(atom_i.partners, atom_j.i)
 push!(atom_j.partners, atom_i.i)
```

### DIFF
--- a/docs/src/examples.md
+++ b/docs/src/examples.md
@@ -46,13 +46,13 @@ function Molly.force(inter::BondableInteraction,
     # Break bonds randomly
     if atom_j.i in atom_i.partners && rand() < inter.prob_break
         delete!(atom_i.partners, atom_j.i)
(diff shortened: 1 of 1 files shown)


jgreener64:
You are correct, thanks.

PR #61, opened 2022-05-09
Doc fixes

leios:
I noticed that the gravity example should have the opposite sign, so here's a quick PR fixing that and a few typos. I also added a codeblock on how to create the SIR plot.

I have an intermediate commit that takes all the examples in the docs and puts them in an `examples/` directory, but I couldn't figure out how to get Documenter to import the blocks of code from a different file, so I deleted the directory in the latest commit.

Basically, if you would like compilable examples in a separate directory, I can revert the latest commit to make them available. Otherwise, I can squash everything as is.

Also: no pressure to merge this if you feel there is not enough substance to it!


docs/src/docs.md
```diff
@@ -392,6 +392,18 @@ visualize(sys.loggers["coords"], box_size, "sim_agent.mp4"; markersize=0.1)
...
+
+```
+using Plots
+
+sir_matrix = zeros(length(sys.loggers["SIR"].fracs_sir),3)
+for i = 1:101
+    sir_matrix[i,:] .= sys.loggers["SIR"].fracs_sir[i][:]
+end
+
+Plots.plot(sir_matrix)
```
jgreener64, line 404: Remove `Plots.`?

docs/src/docs.md
```diff
@@ -188,7 +188,7 @@ This example also shows the use of `Float32`, a 2D simulation and no specified u
 atoms = [Atom(mass=1.0f0), Atom(mass=1.0f0)]
 coords = [SVector(0.3f0, 0.5f0), SVector(0.7f0, 0.5f0)]
 velocities = [SVector(0.0f0, 1.0f0), SVector(0.0f0, -1.0f0)]
-pairwise_inters = (Gravity(nl_only=false, G=1.5f0),)
+pairwise_inters = (Gravity(nl_only=false, G=-1.5f0),)
```
jgreener64, line 191: Since the gravitational constant should be positive, this negative sign should go in line 32 and also line 52 of `src/interactions/gravity.jl`.

leios:
That works as well. I just double-checked and it seems to produce the correct visual from the example

PR #63, opened 2022-06-05
Fixed visualization issue with Makie.jl

chemicalfiend:
Changed the markerspace parameter on the scatter function to feed makie the required data type (looks like the TypeSpace style has been removed from Makie? I can't find the exact point in the version control where this happened but seems to be the case). Also added LinearAlgebra since norm wasn't working on the diatomics simulation without it.


jgreener64:
Great, I'll have a look next week.

jgreener64:
Fixes the issue but the diatomic case at https://github.com/JuliaMolSim/Molly.jl/blob/master/test/simulation.jl#L204-L206 still fails, I think due to the other call to `SceneSpace` at https://github.com/JuliaMolSim/Molly.jl/blob/master/src/makie.jl#L74?

`using LinearAlgebra` should not be required as it is already called at https://github.com/JuliaMolSim/Molly.jl/blob/master/src/Molly.jl#L26. Are you running this as part of a `dev Molly` source tree? That is the best way to develop as it lets you run the tests too.

chemicalfiend:
I'll look into it. I haven't been using a dev Molly source tree but I will to take care of the tests.

jgreener64:
:+1:

PR #64, opened 2022-06-14
Implementation of temperature REMD

JaydevSR:
Includes:

- Types `ReplicaWrapper` and `ReplicaSystem` for handling replica exchange simulations. `ReplicaSystem` is equivalent to `System` but containing multiple replicas at different temperatures. `ReplicaWrapper` is a wrapper for a particular replica.


src/types.jl
```diff
@@ -321,6 +321,78 @@ function System(;
...
+
+# A system to be simulated using replica exchange simulation
+mutable struct ReplicaSystem{D, G, T, A, AD, PI, SI, GI, RL, RT, RI, B, NF, L, F, E} <: AbstractSystem{D}
+    atoms::A
+    atoms_data::AD
+    pairwise_inters::PI
+    specific_inter_lists::SI
+    general_inters::GI
+    replica_list::RL
+    replica_temps::RT
```
JaydevSR, line 340: I was thinking of generalizing this field to contain any reaction coordinate (be it temperature, interactions for hamiltonian REMD, pH etc.) Does this sound like an improvement?

jgreener64 (commented):
Looks good. I agree that there should be a system type to hold the replicas. Since the parametric types are generic it could be made more general by just renaming `temp` to `data` or `info`.

However I think the temperature data should be stored at the level of the simulator rather than at the level of the system, in line with the other thermostats. In this case that would mean removing references to temperature in these two structs and having a `TemperatureREMD` simulator or similar, which would contain temperature data in its struct along with the thermostat to use. What do you think?

JaydevSR:
> However I think the temperature data should be stored at the level of the simulator rather than at the level of the system, in line with the other thermostats. In this case that would mean removing references to temperature in these two structs and having a `TemperatureREMD` simulator or similar, which would contain temperature data in its struct along with the thermostat to use. What do you think?

I was thinking about this as well as we will need different simulators for each replica with different coupling temperature for temp-REMD, so this will be better. We still need to maintain common indices between the simulator replicas and the system replicas so that there is no ambiguity. I will look into simplifying this.

JaydevSR:
I have added the supporting functions to the `ReplicaSystem` type in addition to the suggested edits. Many function for `System` and `ReplicaSystem` are identical. Should I use a union type for that to simplify the code?

jgreener64:
Good changes. Yes you can use a union type, possibly we could introduce an abstract type later.

I think the AtomsBase interface requires that the index given to `position` and `velocity` is the atom index, so the replica index plus atom index used here would be confusing. Possibly use the first replica for now in these functions?

JaydevSR:
For making the simulation, I first define a simulation type:
```julia
struct TemperatureREMD{N, T, S, SD}
    temps::StaticVector{N, T}
    indices::StaticVector{N, Int}
    simulators::SD
end
```

A constructor is defined separately for this type that I am omitting here right now. Then a new methods for the `simulate!` function:
```julia
function simulate!(;
    sys::ReplicaSystem,
    sim::TemperatureREMD,
    n_cycles::Int,
    cycle_length::Int;
    parallel=true
    )

    if sys.n_replicas != length(sim.indices)
        error("Number of replicas in ReplicaSystem and simulators in TemperatureREMD must match.")
    end
    for cycle=1:n_cycles
        for idx in sim.indices
            simulate!(sys.replicas[idx], sim.simulators[idx], cycle_length; parallel=parallel)  #flag1
        end

        if cycle != n_cycles
            n = rand(1:sys.n_replicas)
            m = mod(n + 1, sys.n_replicas)
            β_n, β_m = 1/sim.temps[n], 1/sim.temps[m]
            V_n, V_m = potential_energy(sys.replicas[n]), potential_energy(sys.replicas[m])  #flag2
            Δ = (β_m - β_n)*(V_n - V_m)
            if Δ <= 0 || rand() < exp(-Δ)
                # exchange coordinates
                sys.replicas[n].coords, sys.replicas[m].coords = sys.replicas[m].coords, sys.replicas[n].coords
                # scale velocities
                sys.replicas[n].velocities *= sqrt(β_n/β_m)
                sys.replicas[m].velocities *= sqrt(β_m/β_n)
            end
        end
    end
end
```
*Note: Right now the replica simulation is not in parallel.*

The main issue here is that `sys.replicas[i]` has type `IndividualReplica` for which many of other function will need new methods, but if it had the type `System` then all of this will work without any problem and we don't need to make any changes to other functions (`#flag1` and `#flag2`) or define new methods. But if we make all the replicas their own systems, then there is issue of redundant information in all of them. Any suggestions?

*Reference for T-REMD algorithm that I am using: [https://www.ncbi.nlm.nih.gov/pmc/articles/PMC6484850/](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC6484850/)*

jgreener64:
The simulator outline looks good. In order to follow the `simulate!(sys, sim, n_steps)` convention it might be better to store an `exchange_time` in `TemperatureREMD` and calculate `n_cycles` and `cycle_length` in `simulate!`.

You raise an interesting point about types. I think the most flexible approach might just be to have multiple `System`s. This opens up the possibility of setting up the systems differently too, which some people might find use for. Longer term each replica would need a different `box_size` anyway (barostats change the box size) and I think each replica should have its own neighbour finder since `CellListMapNeighborFinder` stores some neighbour info in the struct.

That leaves the `atoms`, `atoms_data` and interaction fields as duplicated, but since arrays are stored by reference each `System` can just point to the same memory for this. You could still have `ReplicaSystem` which just holds a list of `System`s and does relevant setup in the constructor.

How does that sound?

jgreener64:
(Accidentally clicked close sorry)

The simulator outline looks good. In order to follow the `simulate!(sys, sim, n_steps)` convention it might be better to store an `exchange_time` in `TemperatureREMD` and calculate `n_cycles` and `cycle_length` in `simulate!`.

You raise an interesting point about types. I think the most flexible approach might just be to have multiple `System`s. This opens up the possibility of setting up the systems differently too, which some people might find use for. Longer term each replica would need a different `box_size` anyway (barostats change the box size) and I think each replica should have its own neighbour finder since `CellListMapNeighborFinder` stores some neighbour info in the struct.

That leaves the `atoms`, `atoms_data` and interaction fields as duplicated, but since arrays are stored by reference each `System` can just point to the same memory for this. You could still have `ReplicaSystem` which just holds a list of `System`s and does relevant setup in the constructor.

How does that sound?

JaydevSR:
For the constructor, in order to support multiple sub simulators and couplings, I am having to write multiple checks:
```julia
function TemperatureREMD(;
    simulator, exchange_time, temps, dt, 
    kwargs...)
    S = simulator
    T = eltype(temps)
    N = length(temps)
    ET = typeof(exchange_time)
    temps = SA[temps...]
    indices = SA[collect(1:length(temps))...]

    coupling_required = false
    if simulator ∈ [VelocityVerlet, Verlet, StormerVerlet]
        coupling_required = true
        coupling = kwargs[:coupling]
    end

    if coupling_required && coupling ∈ [AndersenThermostat, BerendsenThermostat]
        couplings = [coupling(temps[i], kwargs[:coupling_const]) for i in indices]
    elseif coupling_required && coupling ∈ [RescaleThermostat]
        couplings = [coupling(temps[i]) for i in indices]
    elseif coupling_required
        error("Temperature coupling required for TemperatureREMD using $(simulator) simulator.")
    end

    if simulator==Langevin
        simulators = Dict(
            [i, simulator(dt=dt, temperature=temps[i], friction=kwargs[:friction],
            remove_CM_motion=kwargs[:remove_CM_motion])]
            for i in indices
        )
    elseif simulator ∈ [VelocityVerlet, Verlet]
        simulators = Dict(
            [i, simulator(dt=dt, temperature=temps[i], coupling=couplings[i],
            remove_CM_motion=kwargs[:remove_CM_motion])]
            for i in indices
        )
    elseif simulator == StormerVerlet
        simulators = Dict(
            [i, simulator(dt=dt, temperature=temps[i], coupling=couplings[i])]
            for i in indices
        )
    else
        error("Simulator type $(simulator) not supported.")
    end
    SD = typeof(simulators)

    return TemperatureREMD{N, T, S, SD, ET}(temps, indices, simulators, exchange_time)
end
```

But when new simulators are added, this function will also need to be changed. This does not seem good.

jgreener64:
I agree it doesn't scale well. This could just be left to the user for now, i.e. the constructor takes a list of simulators as an argument and checks that it is the same length as the list of temperatures. The docs would make clear that the simulators should be set up with the appropriate coupling/temperatures.

JaydevSR:
I was looking a bit into how to parallelize the whole replica simulations and would like to make a few points about that:
1. Right now the `parallel=true` versions of other functions use `Threads.@threads` macro, which is statically scheduled. So, If I use `Threads.@threads` for parallelizing replica simulation, then only one thread will be available for those functions.
2. My suggestion is we use the `@sync` and `@spawn` for all the parallelization, this will change:
    ```julia
    Threads.@threads for i=1:n
        do_some_work()
    end
    ```
    to:
    ```julia
    @sync for i=1:n
        Threads.@spawn do_some_work()
    end
    ```
3. Also, from julia version 1.8 onwards, the behavior of `Threads.@thread` macro is going to change. And it will be dynamically scheduled by default. See: https://github.com/JuliaLang/julia/blob/v1.8.0-rc1/NEWS.md#multi-threading-changes. So maybe these changes won't be necessary after the release of v1.8?

jgreener64:
I think target Julia 1.8 behaviour, it may be that we drop support for Julia 1.7 soon anyway depending on advances in the AD landscape. The parallel behaviour here doesn't have to mirror the existing behaviour though since it is more complex.

As you suggest it would be useful to be able to run each simulation on a certain number of threads, e.g. 16 cores could run 4 replicas on 4 cores each. Is this possible using the threads setup? The current `parallel` keyword argument may be insufficient here, I have considered changing it to `n_threads` across the whole package previously. Would that help?

jgreener64:
Also, sorry for that large breaking change recently. Hopefully there will be fewer breaking changes as the right API emerges!

src/types.jl
```diff
@@ -369,14 +370,14 @@ function ReplicaSystem(;
                 n_replicas,
                 boundary,
                 neighbor_finder=NoNeighborFinder(),
```
JaydevSR, line 372: Right now I am assuming neighbor finder to me immutable, so that same one can be passed to all the replicas. But the `CellListNeigbourFinder` won't work as it stores some data, as @jgreener64 mentioned earlier. Maybe the the neighbor finders can also be passed like loggers are being passed as a Tuple?

jgreener64: Yes sounds good.

src/types.jl
```diff
@@ -369,14 +370,14 @@ function ReplicaSystem(;
...
-                loggers=(),
+                loggers=Tuple(() for i=1:n_replicas),
```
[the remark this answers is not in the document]
jgreener64: `replica_loggers` or `loggers_list`?

JaydevSR:
I have added the simulator for TemperatureREMD. Right now I am using the `parallel` kwarg but I agree with you that its usage in this case is not ideal. 

The current `@sync` and `Threads.@spawn` method chooses a free thread to perform one cycle of the simulation and say if there are 4 replicas then 4 threads will be used leaving the other threads free for the other parallel calls. But it is not ensured that all the threads are distributed equally among all the replicas (which may be problematic if the number of nested parallel calls is close to the number of free threads due to which workload may not be evenly balanced). 

About using using an `n_threads` argument, in that case we can make use of the `Threads.foreach()` function (see: https://docs.julialang.org/en/v1/base/multi-threading/#Base.Threads.foreach) which seems useful in our case.

Also, I performed some initial simulations (akin to the Leonard Jones gas example in docs) to check the sanity of the simulator and here are some results, for 4 replicas of the system at 100 K, 105 K, 110 K and 115 K:
1. Temperature of each replica with steps temperature_of_each_replica
2.  X-coordinate of one of atoms with time for each replica (change in color denotes a replica exhange): x_coordiantes_of_atom_7_in_each_replica

Still I am a bit skeptical about the replica exchange part and I will check it once again to ensure that I am not doing something wrong.

src/types.jl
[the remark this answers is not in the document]
JaydevSR: Yes I initially had `replica_loggers` in mind. I will go with that.

JaydevSR:
@jgreener64 Okay, I will try making these changes. Also thanks for suggestions on test simulations.

JaydevSR:
It seems that making these changes using just the base library will be a bit more of a trouble than we need to handle. Instead JuliaFolds/FLoops.jl, already supports such a change in number of threads utilized. See: https://juliafolds.github.io/data-parallelism/howto/faq/#set-nthreads-at-run-time. This library will also support much more different flavors of thread based parallelization (via JuliaFolds/FoldsThreads.jl) which can come in handy for load balancing. Should I use this library if adding it as a dependency won't be a problem?

jgreener64:
Taking the dependency is fine so yes go ahead and try it out if it provides functionality beyond base Julia. Long term I think we will change `parallel` to `n_threads` anyway though.

I guess using FLoops.jl may also allow the CPU `forces` path https://github.com/JuliaMolSim/Molly.jl/blob/3bed491a9d183492262d5ec323a0f509923e51f8/src/force.jl#L221 to be written more concisely.

JaydevSR:
Okay then I will experiment with a bit and make a PR if I am able to get it to work as expected.

src/types.jl
```diff
@@ -454,17 +532,16 @@ function ReplicaSystem(;
     
     K = typeof(k_converted)
 
-    # TODO:  CellListNeighborFinder is mutable so can't be passed to all replicas.
-    replicas = Tuple(System{D, G, T, A, AD, PI, SI, GI, C, V, B, NF, L, F, E, K}(
+    replicas = Tuple(System{D, G, T, CU, A, AD, PI, SI, GI, C, V, B, NF, typeof(replica_loggers[i]), F, E, K}(
             atoms, atoms_data, pairwise_inters, specific_inter_lists,
-            general_inters, copy(coords), copy(velocities), boundary, neighbor_finder,
-            loggers[i], force_units, energy_units, k_converted) for i=1:n_replicas)
+            general_inters, copy(coords), copy(velocities), boundary, deepcopy(neighbor_finder),
```
JaydevSR, line 537: Right now a deepcopy of neighbor_finder is made in order to support `CellListNeighborFinder`. This can be made more efficient by defining a method of `Base.copy` for neighbor finders which passes the field that are not mutated by reference and then using this.

For example, for `DistanceNeighborFinder` it can be like:
```julia
Base.copy(nf::DistanceNeighborFinder{D, B, I}) = DistanceNeighborFinder{D, B, I}([getfield(x, k) for k ∈ fieldnames(T)]...)
``` 

whereas for `CellListNeighborFinder` we can have something like:
```julia
Base.copy(clnf::CellListMapNeighborFinder{N, T}) = CellListMapNeighborFinder{N, T}(
    clnf.nb_matrix, clnf.matrix_14, clnf.n_steps, clnf.dist_cutoff,
    deepcopy(clnf.cl), deepcopy(clnf.aux), deepcopy(clnf.neighbors_threaded)
)
```

Then any custom neighbor finder will need to define this method for it to be used in replica simulation.

src/simulators.jl
```diff
@@ -438,3 +438,110 @@ function B_step!(s, dt_eff, acceleration_vector, compute_forces::Bool, n_threads
...
+                    n_threads::Int=Threads.nthreads())
+    if sys.n_replicas != length(sim.simulators)
+        error("Number of replicas in ReplicaSystem and simulators in TemperatureREMD must match.")
+    end
+
+    if n_threads > sys.n_replicas
+        thread_div = equal_parts(n_threads, sys.n_replicas)
+    else # pass 1 thread per replica
+        thread_div = equal_parts(sys.n_replicas, sys.n_replicas)
+    end
```
JaydevSR, line 497: A static division of threads is done among all the replicas. I am not sure whether this can be made dynamic or not without much overhaul.

JaydevSR:
I ran the simulator on Leonard-Jones gas and it seem to work without any problems. What possible tests can I make that will better highlight the replica exchange at work that this basic simulation? After this I can write some unit tests to improve coverage.

src/simulators.jl
```diff
@@ -438,3 +438,110 @@ function B_step!(s, dt_eff, acceleration_vector, compute_forces::Bool, n_threads
...
+
+    # calculate n_cycles and n_steps_per_cycle from dt and exchange_time
+    n_cycles = convert(Int64, (n_steps * sim.dt) ÷ sim.exchange_time)
+    cycle_length = n_steps ÷ n_cycles
+    remaining_steps = n_steps % n_cycles
+
+    # scale to correct temperatures
+    for i in eachindex(sys.replicas)
+        sys.replicas[i].velocities .*= sqrt(sim.temperatures[i] / temperature(sys.replicas[i]))
+    end
```
JaydevSR, line 507: As the same velocities are assigned to each system during replica system setup, I am scaling the velocities to correct temperatures. Here the velocities can also be assigned using `random_velocities!`, but I am not sure which approach is correct from a simulation point of view? Also, should this be optionally controlled by a kwarg in case the simulation is a continuation of some already performed simulation?

src/types.jl
jgreener64: Not sure it's worth speeding this up as it is only called once, unless it is really slow.

src/simulators.jl
jgreener64: A static division is probably okay for now.

Yes CellListMap.jl uses a number of threads up to `Threads.nthreads()` when `parallel=true`, I opened an issue to discuss this (https://github.com/m3g/CellListMap.jl/issues/64).

src/simulators.jl
jgreener64: I think starting velocities should be determined when setting up the `ReplicaSystem`. As you say this works better with continuing simulations, and it is consistent with the other simulators. Perhaps that could take `replica_velocities` instead of `velocities`, and set up the relevant `random_velocities` when `replica_velocities` has its default value of `nothing`.

jgreener64:
Great. I think a Lennard-Jones simulation is okay for now but it would be good to see the graphs you had previously with a clearer temperature profile.

It also would be useful to log when the exchanges took place, then compare the number of exchanges over a simulation to an expected number. Perhaps the `ReplicaSystem` could have a `loggers` field, with the default being a new `ReplicaExchangeLogger` that records the time step, the indices exchanged and `Δ` for each exchange.

On the topic of loggers, I think `ReplicaSystem` shouldn't store `replica_loggers` in its `struct`. `replica_loggers` should be used in the constructor as is, then the loggers can be accessed with `replica_system.replicas[i].loggers`.

src/simulators.jl
jgreener64: Looks like the response has the answer and we can use something like `nbatches=(min(n, 8), n)` where `n` is the threads per replica.

src/types.jl
JaydevSR: Yeah seems fair. I will keep it this way.

src/simulators.jl
JaydevSR: As replica system is a general type so temperature values are not passed to that so we can't use `random_velocities` there. Instead If `replica_velocities` is `nothing`, we can assign velocities to be zero same as `System` type.

jgreener64: Ah yes, sounds good. `simulate!` could also have a keyword argument `assign_velocities`, default `false`, that randomizes them in the simulator. Then the user can either set them up manually with the `ReplicaSystem`, giving flexibility and consistency with the other simulators, or call `simulate!` with `assign_velocities=true`.

src/simulators.jl
JaydevSR: There is one problem though, the `nbatches` argument is passed to the constructor while defining `CellListNeighborFinder`, which is then used in the underlying cell list type. So we can not update it later say during the thread division in the replica exchange simulator.

JaydevSR:
I performed a test simulation on 3 replicas of Leonard-Jones gas on 8 threads, using `DistanceNeighborFinder` and a `Langevin` simulator. 
```julia
temp_vals = [120.0u"K", 160.0u"K", 200.0u"K"]
simulator = TemperatureREMD(
    dt=0.005u"ps",
    temperatures=temp_vals,
    simulators=[
        Langevin(
            dt=0.005u"ps",
            temperature=tmp,
            friction=0.1u"ps^-1",
        )
        for tmp in temp_vals],
    exchange_time=5u"ps",
);
```

The exchanges were logged with the help of the new `ReplicaExchangeLogger` and we have the following exchanges in form of `(step, exchanged indices)` after simulating for `20_000` steps:
```
 (1000, (2, 3))
 (5000, (2, 3))
 (6000, (1, 2))
 (7000, (3, 1))
 (8000, (2, 3))
```

These are plots for temperatures of each replica and the x-coordinates of the one of the atoms in these replicas:
1. Temperatures: 
temperature_of_each_replica

2. Coordinates: 
x_coordiantes_of_atom_10_in_each_replica

src/loggers.jl
```diff
@@ -473,3 +474,34 @@ function Base.show(io::IO, aol::AverageObservableLogger)
...
+                    indices::Tuple{Int,Int},
+                    delta::Float64,
+                    n_threads=Threads.nthreads())
+    push!(rexl.indices, indices)
+    push!(rexl.steps, step_n+rexl.end_step)
+    push!(rexl.delta, delta)
+    rexl.n_exchanges += 1
+end
+
+function finish_logs!(rexl::ReplicaExchangeLogger)
```
jgreener64, line 503: I think this should record the final step number regardless of recorded exchanges. Perhaps pass `n_steps` to this function and add that to the current `end_step`?

src/types.jl
```diff
@@ -387,21 +388,191 @@ function System(;
...
+mutable struct ReplicaSystem{D, G, T, CU, A, AD, PI, SI, GI, RS, B, NF, EL, F, E, K} <: AbstractSystem{D}
+    atoms::A
+    atoms_data::AD
+    pairwise_inters::PI
+    specific_inter_lists::SI
+    general_inters::GI
+    n_replicas::Integer
+    replicas::RS
+    boundary::B
+    neighbor_finder::NF
```
jgreener64, line 444: Doesn't need to be stored in the `struct` since it is copied to each system?

src/types.jl
```diff
@@ -387,21 +388,191 @@ function System(;
...
+    end
+    if n_cuarray == n_replicas && !isa(atoms, CuArray)
+        throw(ArgumentError("The velocities are on the GPU but the atoms are not"))
+    end
+
+    if length(replica_loggers) != n_replicas
+        throw(ArgumentError("There are $(length(replica_loggers)) loggers but $(n_replicas) replicas"))
+    end
+
+    if energy_units == NoUnits
```
jgreener64, line 532: Since this code block appears in `System` too it can be moved to a function that takes `k` and `energy_units`.

src/simulators.jl
```diff
@@ -438,3 +438,121 @@ function B_step!(s, dt_eff, acceleration_vector, compute_forces::Bool, n_threads
...
+        end
+
+        if cycle != n_cycles
+            n = rand(1:sys.n_replicas)
+            m = mod(n, sys.n_replicas) + 1
+            k_b = sys.k
+            T_n, T_m = sim.temperatures[n], sim.temperatures[m]
+            β_n, β_m = 1/(k_b*T_n), 1/(k_b*T_m)
+            V_n, V_m = potential_energy(sys.replicas[n]), potential_energy(sys.replicas[m])
+            Δ = ustrip((β_m - β_n)*(V_n - V_m))
```
jgreener64, line 524: `ustrip` not required?

src/simulators.jl
```diff
@@ -438,3 +438,121 @@ function B_step!(s, dt_eff, acceleration_vector, compute_forces::Bool, n_threads
...
+        end
+    end
+
+    for cycle=1:n_cycles
+        @sync for idx in eachindex(sim.simulators)
+            Threads.@spawn Molly.simulate!(sys.replicas[idx], sim.simulators[idx], cycle_length; n_threads=thread_div[idx])
+        end
+
+        if cycle != n_cycles
+            n = rand(1:sys.n_replicas)
```
jgreener64, line 518: I think all pairs should be considered each cycle with another loop otherwise the exchange rate is too low. Following the Gromacs policy (https://manual.gromacs.org/current/reference-manual/algorithms/replica-exchange.html) of exchanging odd pairs on odd cycles and even pairs on even cycles sounds like a good idea.

jgreener64 (commented):
Looks good, the temperature difference is much clearer there. Ideally turn that simulation into a test that looks for the temperatures and exchange count to fall within an expected range.

Not that it would block the PR, but it would be interesting to see how performance scales with replica count. For example, how long does 4 replicas on 8 cores for 1000 atoms take compared to 1 simulation on 2 cores?

src/simulators.jl
jgreener64: In that case it might be easier to leave it up to the user to specify when giving the neighbour finder to `ReplicaSystem`.

src/simulators.jl
JaydevSR: Here `Δ` is giving me units of `mol^-1`. So, the comparison below doesn't work. I think I should log the value of `Δ` with these units as well.

jgreener64: That `mol^-1` should be divided out, it's a little fiddly but the following should work. Replace `k_b = sys.k` with
```julia
if dimension(sys.energy_units) == u"𝐋^2 * 𝐌 * 𝐍^-1 * 𝐓^-2"
    k_b = sys.k * T(Unitful.Na)
else
    k_b = sys.k
end
```
and `sys::ReplicaSystem` in the function arguments with `sys::ReplicaSystem{D, G, T}` ... ` where {D, G, T}`.

This will probably change the acceptance rate in the examples.

src/simulators.jl
JaydevSR: I have changed the exchange scheme to the one above, please check if it is fine. After the change, the number of exchanges have increased from about 5-6 to 12-16 for the above simulation.

src/simulators.jl
JaydevSR: Okay I have done this, the number of exchanges also changed (increased). In the above example I am getting about 30 exchanges (from 40 attempts) maybe because the temperature difference is small.

src/simulators.jl
jgreener64: Looks good.

src/simulators.jl
```diff
@@ -438,3 +438,130 @@ function B_step!(s, dt_eff, acceleration_vector, compute_forces::Bool, n_threads
...
+            random_velocities!(sys.replicas[i], sim.temperatures[i]; rng=rng)
+        end
+    end
+
+    for cycle=1:n_cycles
+        @sync for idx in eachindex(sim.simulators)
+            Threads.@spawn simulate!(sys.replicas[idx], sim.simulators[idx], cycle_length; n_threads=thread_div[idx])
+        end
+
+        if cycle != n_cycles
```
jgreener64, line 517: Probably better to skip this check, since if `remaining_steps > 0` you would want to exchange before that.

JaydevSR:
**Benchmarks**

1. `System` with 1000 atoms and `DistanceNeighborFinder`, `LeonardJones` interactions using `Langevin` simulator for 5000 steps (2 threads): 
    ```77.253948 seconds (582.38 k allocations: 4.966 GiB, 0.87% gc time)```
2. `ReplicaSystem` with 4 replicas of above system with same simulator and steps (8 threads): 
    ```173.750242 seconds (2.18 M allocations: 20.255 GiB, 2.47% gc time, 0.45% compilation time)```

*Ratio of times: 2.25*

test/simulation.jl
```diff
@@ -451,6 +451,62 @@ end
...
+            for temp in temp_vals],
+        exchange_time=2.5u"ps",
+    )
+
+    rng = MersenneTwister()
+    @time simulate!(repsys, simulator, 10_000; assign_velocities=true, rng=rng);
+    @time simulate!(repsys, simulator, 10_000; assign_velocities=false, rng=rng);
+
+    efficiency = repsys.exchange_logger.n_exchanges / repsys.exchange_logger.n_attempts
+    @test efficiency > 0.4
```
JaydevSR, line 507: I am not sure about this. Is this a good test (what should be the expected efficiency)?

jgreener64: Not sure what it should be but checking a sensible range is fine. It's probably also worth testing that the temperatures stored in the temperature logger for each replica have their mean within some range of the specified temperatures.

test/simulation.jl
```diff
@@ -451,6 +451,62 @@ end
...
 end
 
+@testset "Temperature Replica Exchange" begin
+    n_atoms = 100
+    atom_mass = 10.0u"u"
+    atoms = [Atom(mass=atom_mass, σ=0.3u"nm", ϵ=0.2u"kJ * mol^-1") for i in 1:n_atoms]
+    boundary = CubicBoundary(2.0u"nm", 2.0u"nm", 2.0u"nm")
+    coords = place_atoms(n_atoms, boundary, 0.3u"nm")
+
+    pairwise_inters = (LennardJones(),)
```
jgreener64, line 461: `nl_only=true` or it won't use the neighbour list (also in the other test file).

src/loggers.jl
```diff
@@ -473,3 +474,34 @@ function Base.show(io::IO, aol::AverageObservableLogger)
...
             " samples collected for observable ", aol.observable)
 end
+
+mutable struct ReplicaExchangeLogger
+    n_replicas::Int
+    n_attempts::Int
+    n_exchanges::Int
+    indices::Vector{Tuple{Int,Int}}
+    steps::Vector{Int}
+    delta::Vector{Float64}
```
jgreener64, line 484: Ideally this would be `Vector{T}` with `T` as a type parameter, allowing `Float32` to be used.

jgreener64 (commented):
The benchmark results are interesting - it's not essential for this PR but it would be good to profile and see what is stopping that ratio from being lower. Make sure the neighbour list is being used as well (see other comment).

JaydevSR:
I did profiling on the `simulate!` function and the bottleneck seems to be the part where the replicas are being exchanged. Most of the part there is type unstable. I will try to make it type stable and see if there are any improvements.

JaydevSR:
It seems that I misinterpreted the profiles before. The actual issue with the drop in performance seems to do with task scheduling. Here are the flame graph of the profiles:
overall_flame_graph
find_neighbors_focus_flame_graph

In the second image we can see this more clearly, more than 60% of time is spent in the `wait()` in the call stack of `find_neighbors!`. This as far as I can think should be because of clashing scheduling of `@floops` and `@spawn`. I think I may be able to fix this by using one of the different executors for the parallel loop in `find_neighbors!`. See: https://juliafolds.github.io/FoldsThreads.jl/dev/#FoldsThreads.WorkStealingEx. I will see if this works.

JaydevSR:
So I experimented a bit to see if there can be an improvement or not. After trying different executors and a pure julia implementation also, the case with 2 threads 1 system and 8 threads 4 replicas, gives no improvement. But for a 1 thread 1 system and 4 threads 4 replicas, the ratio was improved a bit to about 1.6, as my machine supports at most 8 threads so I think the wait time was high in the first case because all threads were getting used. So, maybe it is best to use less than the maximum available threads for the replica simulation to avoid such clashes.

But still julia v1.8 will bring some changes to multithreading, so it will be interesting to see if that gives any improvement for nested threaded loops.

test/simulation.jl
```diff
@@ -451,6 +451,66 @@ end
...
+            Langevin(
+                dt=0.005u"ps",
+                temperature=temp,
+                friction=0.1u"ps^-1",
+            )
+            for temp in temp_vals],
+        exchange_time=2.5u"ps",
+    )
+
+    rng = MersenneTwister()
```
jgreener64, line 502: Is this required? If not then remove.

jgreener64 (commented):
Fair enough. Yes let's see how it changes when Julia 1.8 comes out.

I think this is ready to merge after some minor changes. Also add `TemperatureREMD` to the list of simulators in `docs/src/docs.md` line 679.

JaydevSR:
Yay 🎊 . 

Your suggestion seem excellent. This will help in generalizing the simulation code. I will go ahead and make these additions. Also, I think I can try implementing the Hamiltonian-REMD next, that would require the replicas to have different potentials. So, I will also need to make some changes to `ReplicaSystem` to allow for this.

PR #67, opened 2022-06-21
Add Langevin simulator for more general splittings

noeblassel:
Add support for more general splitting schemes for Langevin dynamics. 
# Changes
- [x] Add `LangevinSplitting` type and `simulate!` method to simulate.jl
- [x] Add docstring describing the simulator
- [x] Write test comparing the current Langevin and LangevinSplitting with identical RNG, in the case of a BAOA splitting.

### DIFF
--- a/src/simulators.jl
+++ b/src/simulators.jl
@@ -330,22 +330,19 @@ A Langevin simulator using a general splitting scheme, consisting of a successio
 - `temperature::temperatureType`: The equilibrium temperature.
 - `splitting::splittingType`: The splitting specifier. Should be a string consisting of the characters `A`,`B` and `O`. Strings with no `O`s reduce to deterministic symplectic schemes.
 - `remove_CM_motion::Bool=true`: Whether to remove the centre of mass motion at each simulation iteration.
-- `rseed::UInt32=round(UInt32,time())`: An optional seed for the random number generator. Defaults to nearest epoch time in seconds.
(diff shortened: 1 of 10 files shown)


src/simulators.jl
```diff
@@ -320,3 +321,106 @@ function simulate!(sys,
...
+- `splitting::splittingType`: The splitting specifier. Should be a string consisting of the characters `A`,`B` and `O`. Strings with no `O`s reduce to deterministic symplectic schemes.
+- `remove_CM_motion::Bool=true`: Whether to remove the centre of mass motion at each simulation iteration.
+- `rseed::UInt32=round(UInt32,time())`: An optional seed for the random number generator. Defaults to nearest epoch time in seconds.
+- `rng::rngType=MersenneTwister`: The random generator type for the simulation. The RNG is seeded with rseed.
+"""
+struct LangevinSplitting{dtType,frictionType,temperatureType,rngType,splittingType}
+    dt::dtType
+    friction::frictionType
+    temperature::temperatureType
+    rng::rngType
```
jgreener64, line 340: I think the Langevin variants should be consistent in how they use the random seed. Passing it to `simulate!` as a kwarg might be more flexible than having it in the struct, since a user may make repeated calls to `simulate!` with the same simulator and wish to control the random seed each time.

jgreener64 (commented):
Great. Also add it to the list of simulators in the documentation.

jgreener64:
:+1:

PR #68, opened 2022-06-22
Autocorrelation logger

noeblassel:
Add a logger type for general time correlations with docstring, and define a custom `Base.getproperty` method for this type to retrieve the computed correlation functions.


jgreener64 (commented):
Looks good, mainly style stuff.

PR #69, opened 2022-06-22
modify Dict loggers to NamedTuple throughout

noeblassel:
# Suggested changes:
Replace every instance of a `System`'s loggers as a `Dict` to a `NamedTuple`.
This is important for type-inference reasons, especially when the `System` has several different loggers. In the case of `TimeCorrelationLogger`, the speedup is significant.

# Cons
This introduces ugly syntax. For example, if
```julia
sys.loggers=(coords=CoordinateLogger(args...))
```
the syntax to access the recorded coordinate history is 
```julia
sys.loggers.coords.coords
```
This could be alleviated by having a common (possibly non-exported) type for all loggers which push some observation onto a history vector, like

```julia
struct GeneralObservableLogger {T,F}
    n_steps::Int64
    history::Vector{T}
    observable::F
end

function log_property!(logger::GeneralObservableLogger,s::System,neighbors=nothing,step_n::Integer=0;parallel::Bool=true)
    if (step_n % logger.n_steps) == 0
        obs=logger.observable(sys,neighbors;parallel=parallel)
        push!(logger.history,obs)
    end
end
```
and have `TemperatureLogger`, `CoordinateLogger`, etc... wrap a constructor for this type.
So logged observations would be accessed through
`sys.loggers.coords.history`
and so on.
This would also shorten `loggers.jl` by removing a lot of boilerplate code / make it easier for users to define custom loggers of this kind. I can implement this change if that sound good.

### DIFF
--- a/docs/src/differentiable.md
+++ b/docs/src/differentiable.md
@@ -73,7 +73,7 @@ simulator = VelocityVerlet(
 
 function loss(σ)
     atoms = [Atom(0, 0.0, atom_mass, σ, 0.2, false) for i in 1:n_atoms]
-    loggers = Dict("coords" => CoordinateLogger(Float64, 10))
+    loggers = (coords = CoordinateLogger(Float64, 10))
 
     s = System(
         atoms=atoms,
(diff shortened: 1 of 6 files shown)


jgreener64 (commented):
Looks great.

I think the argument order for `GeneralObservableLogger` should be `observable`, `T`, `n_steps` in line with other Julia functions.

Also I think we should have a function to access the stored values, perhaps `Base.values`, and use this throughout. It can have a default definition falling back to getting the `history` field. This makes it easier to implement lazy computation on custom loggers later.

PR #74, opened 2022-06-29
add FENE bond and cosine angle.

ehsanirani:
Hi. Here I provided a small piece of code to add FENE bond (following [Kremer, Grest J. Chem. Phys. 92, 5057 (1990)](https://doi.org/10.1063/1.458541)) and the cosine angle potential to it. These are very common interactions for modeling chromatin or many other polymeric systems. I tried to follow the code-style used in Molly and also provided relevant equations in the docs. Please let me know if I missed something. Hope this small piece of code is useful for others.

New files:
* `src/interactions/fene_bond.jl`
* `src/interactions/cosine_angle.jl`


src/interactions/fene_bond.jl
```diff
@@ -0,0 +1,61 @@
...
+```
+"""
+struct FENEBond{D, K} <: SpecificInteraction
+    k::K
+    r0::D
+    ϵ::K
+    σ::D
+end
+
+FENEBond(; k=1, r0=1.5, ϵ=1.0, σ=1.0) = FENEBond{typeof(r0), typeof(k)}(k, r0, ϵ, σ)
```
jgreener64, line 27: Defaults should have relevant units in general.

src/interactions/fene_bond.jl
```diff
@@ -0,0 +1,61 @@
...
+
+@inline @inbounds function force(b::FENEBond, coord_i, coord_j, boundary)
+    ab = vector(coord_i, coord_j, boundary)
+    r = norm(ab)
+    r2 = r^2
+    r2inv = 1.0 / r2
+    r6inv = r2inv^3
+    σ6 = b.σ^6
+    fwca = 0.0
```
jgreener64, line 36: Is a different initialisation required when units are used?

jgreener64 (commented):
Thanks for doing this. Looks good, some comments about units.

Would it also be possible to add a unit test for each in `test/basic`?

src/interactions/fene_bond.jl
ehsanirani: Just removed the default values.

src/interactions/fene_bond.jl
ehsanirani: Thanks for noticing it. I changed the code to handle units properly.

ehsanirani:
I will add some unit tests and an example code, hopefully in the next few days.

ehsanirani:
I revised the code so it respects units now. Also added tests.

jgreener64:
Thanks for this!

PR #77, opened 2022-07-13
adding overload so GPU kernels execute without unsupported dynamic function call

leios:
Several GPU kernels are not functional because of an `unsupported dynamic function call to force(...)`. This adds a simple fallback for calling `force(..., weight_14)` as just `force(...)`. It seems to work on my GPU, at least.


jgreener64:
Thanks, I'll remove the redundant branch in force.jl after merging this.

PR #82, opened 2022-08-06
SHAKE constraint for bonds

chemicalfiend:
Added the code from my sandbox repository (https://github.com/chemicalfiend/md-constraints) for SHAKE bonds. A toy test for this : https://github.com/chemicalfiend/md-constraints/blob/main/test/molly-shake-test.jl  can be tried to verify working.

Some concerns : All the tests from before the summer seem to work fine with the code I've added here. The tests concerning ReplicaSystems and TemperatureREMD seem to fail ; I've commented out the test-sets concerning those in this PR but I'll be sure to add it back in once they run (I'll have to discuss with @JaydevSR  on how to smooth those issues out) and the test with Zygote failed in one case (I'll start a comment thread for that). 

I'll be adding the multi-atom and angle cases in a separate PR soon, along with the RATTLE PR for velocity updates. I'm also contemplating bringing over the Bond Length and Bond Angle loggers from the sandbox to keep track of variations (if any) and to verify that the constraints are working, though I'm not sure if they'd have any other use cases.

So, in brief, the immediate todo's for this PR look like : 

- [x]  ~~Get the Replica and Zygote tests to run~~
- [ ]  Need to clean up the toy test and add it to docs as an example
- [ ]  Add bond-length and bond-angle loggers (??)

### DIFF
--- a/src/loggers.jl
+++ b/src/loggers.jl
@@ -484,7 +484,7 @@ A logger that records exchanges in a replica exchange simulation.
 The logged quantities include the number of exchange attempts (`n_attempts`),
 number of successful exchanges (`n_exchanges`), exchanged replica indices (`indices`),
 exchange steps (`steps`) and the value of Δ i.e. the argument of metropolis rate for
-the exchanges (`delta`).
+the exchanges (`deltas`).
 """
 mutable struct ReplicaExchangeLogger{T}
     n_replicas::Int
@@ -492,7 +492,7 @@ mutable struct ReplicaExchangeLogger{T}
     n_exchanges::Int
     indices::Vector{Tuple{Int, Int}}
     steps::Vector{Int}
-    delta::Vector{T}
+    deltas::Vector{T}
     end_step::Int
 end
 
@@ -511,7 +511,7 @@ function log_property!(rexl::ReplicaExchangeLogger,
                        n_threads::Integer=Threads.nthreads())
     push!(rexl.indices, indices)
     push!(rexl.steps, step_n + rexl.end_step)
-    push!(rexl.delta, delta)
+    push!(rexl.deltas, delta)
     rexl.n_exchanges += 1
 end
 


JaydevSR:
Hi, so what seems to be the problem with the replica exchange test, are they interfering with the code you wrote somehow?

chemicalfiend:
The Zygote test seemed to **pass** the second time around, I didn't make any changes though. I'm attaching the error I got the first time for reference. Must have been some stochastic effect?? 

Sat Aug  6 10:54:49 AM IST 2022

chemicalfiend:
Replica System error @JaydevSR. A similar one pops up for TemperatureREMD too (assuming I comment this one out)
Sat Aug  6 12:19:37 PM IST 2022

From my understanding my code shouldn't conflict with the ReplicaSystem stuff. I have added a parameter called "constraints" to the System struct, but not adding any constraints shouldn't have any effect. I also made some changes to VelocityVerlet's simulate! function. If you have something going on with VelocityVerlet in this test, it might have conflicted. I'll try getting to the bottom of it.

JaydevSR:
As you have changed the arguments for the `System` type, you should also add the same argument for the `ReplicaSystems` which will then pass it to construct replicas (which are of the `System` type). Just add an argument to `ReplicaSystem` constructor and pass it to the line where `System` is constructed with the type information.

chemicalfiend:
Okay, looks like all the tests pass now. Thanks

src/types.jl
```diff
@@ -442,12 +457,13 @@ interface described there.
 - `gpu_diff_safe::Bool`: whether to use the code path suitable for the
     GPU and taking gradients. Defaults to `isa(coords, CuArray)`.
 """
-mutable struct ReplicaSystem{D, G, T, CU, A, AD, PI, SI, GI, RS, B, EL, F, E, K} <: AbstractSystem{D}
+mutable struct ReplicaSystem{D, G, T, CU, A, AD, PI, SI, GI, CN, RS, B, EL, F, E, K} <: AbstractSystem{D}
```
JaydevSR, line 460:
```suggestion
mutable struct ReplicaSystem{D, G, T, CU, A, AD, PI, SI, GI, RS, B, EL, F, E, K} <: AbstractSystem{D}
```

src/types.jl
```diff
@@ -442,12 +457,13 @@ interface described there.
...
     atoms::A
     atoms_data::AD
     pairwise_inters::PI
     specific_inter_lists::SI
     general_inters::GI
+    constraints::CN
```
JaydevSR, line 466:
```suggestion

```

src/types.jl
```diff
@@ -548,16 +566,16 @@ function ReplicaSystem(;
...
+    replicas = Tuple(System{D, G, T, CU, A, AD, PI, SI, GI,CN, C, V, B, NF,
                             typeof(replica_loggers[i]), F, E, K}(
             atoms, atoms_data, pairwise_inters, specific_inter_lists,
-            general_inters, copy(coords), replica_velocities[i], boundary, deepcopy(neighbor_finder),
+            general_inters, constraints, copy(coords), replica_velocities[i], boundary, deepcopy(neighbor_finder),
             replica_loggers[i], force_units, energy_units, k_converted) for i in 1:n_replicas)
     RS = typeof(replicas)
 
-    return ReplicaSystem{D, G, T, CU, A, AD, PI, SI, GI, RS, B, EL, F, E, K}(
+    return ReplicaSystem{D, G, T, CU, A, AD, PI, SI, GI, CN, RS, B, EL, F, E, K}(
```
JaydevSR, line 576:
```suggestion
    return ReplicaSystem{D, G, T, CU, A, AD, PI, SI, GI, RS, B, EL, F, E, K}(
```

src/types.jl
```diff
@@ -548,16 +566,16 @@ function ReplicaSystem(;
...
             atoms, atoms_data, pairwise_inters, specific_inter_lists,
-            general_inters, n_replicas, replicas, boundary, 
+            general_inters, constraints, n_replicas, replicas, boundary,
```
JaydevSR, line 578:
```suggestion
            general_inters, n_replicas, replicas, boundary, 
```

JaydevSR (commented):
Saving constraint information in the `ReplicaSystem` structure is not required so you can removed that.

src/types.jl
```diff
@@ -41,6 +42,11 @@ Custom neighbor finders should sub-type this type.
 """
 abstract type AbstractNeighborFinder end
 
+"""
+A constraint for bonds and angles.
+"""
+abstract type Constraint end
```
jgreener64, line 48: Unless it's necessary I think it's a better idea not to introduce a new abstract type. I've been meaning to remove the remaining unnecessary abstract types and will do at some point.

src/simulators.jl
```diff
@@ -137,13 +137,17 @@ function simulate!(sys,
     sim.remove_CM_motion && remove_CM_motion!(sys)
 
     for step_n in 1:n_steps
-        sys.coords += sys.velocities .* sim.dt .+ (remove_molar.(accels_t) .* sim.dt ^ 2) ./ 2
-        sys.coords = wrap_coords.(sys.coords, (sys.boundary,))
+        sys_copy = sys
```
jgreener64, line 140: This doesn't actually copy the system? The two systems in the `run_constraints!(sys, sys_copy.coords, sim.dt)` line refer to the same thing. I would `copy(sys.coords)` if required.

src/constraints.jl
```diff
@@ -0,0 +1,84 @@
+export
+    run_constraints!,
+    SHAKE,
+    apply_constraint!
+
+
+
+function get_masses(sys)
```
jgreener64, line 8: The `masses` function exists for this.

src/constraints.jl
```diff
@@ -0,0 +1,84 @@
...
+    masses = []
+    for atom in sys.atoms
+        push!(masses, atom.mass)
+    end
+    return masses
+end
+
+function run_constraints!(sys, new_coords, dt)
```
jgreener64, line 16: `apply_constraints!` might be better.

src/constraints.jl
```diff
@@ -0,0 +1,84 @@
...
+
+struct SHAKE{D, B} <: Constraint
+    d::D
+    bond_list::B
+end
+
+
+function apply_constraint!(sys, new_coords, dt, constr::SHAKE)
+    
+    bond_list = constr.bond_list
```
jgreener64, line 36: These setup steps that stay the same throughout the simulation can be moved to the `SHAKE` constructor and stored in the struct.

src/constraints.jl
```diff
@@ -0,0 +1,84 @@
+export
+    run_constraints!,
```
jgreener64, line 2: Exported functions should have docstrings.

jgreener64 (commented):
Looks like a good start. The Zygote error is unrelated. I wouldn't worry too much about the loggers at this stage unless they are useful for testing.

Currently the constraints are applied to the `VelocityVerlet` integrator but it would be good to apply them to `Verlet`, `StormerVerlet` and the Langevin variants with this PR too.

There should be a couple of unit tests added: a single call to `apply_constraint!` on bonds showing they can be constrained, and a simulation showing that the bonds have the correct distance throughout.

Finally it would be nice to see the timings of a simulation with and without the constraints to check it can run quickly.

PR #84, opened 2022-08-10
SHAKE constraint implementation for bonds 

chemicalfiend:
I accidentally closed the previous PR (by merging it back into my old repository).

I've made most of the changes requested in the comments of the previous PR by @jgreener64 .

I kept the name of the master function that applies all the constraints as `run_constraints!` since I felt `apply_constraints` would be confusing when I've used `apply_constraint!` for a single constraint. If there's no better nomenclature I don't mind using `apply_constraints` too. 

Copying over the list of remaining tasks: 

- [x]  Timing simulations with and without constraints
- [x]  Add a unit test with SHAKE
- [ ]   Updating docs.


jgreener64:
Changes look good so far.

I would move the `wrap_coords` line below the `run_constraints!` line in each simulator since we want to guarantee each step ends with the coordinates in the bounding box. For `LangevinSplitting` you could add another `wrap_coords` line after `run_constraints!`.

chemicalfiend:
Completed the requested changes, including adding the test for constraints. Here is a benchmark for a simulation with constraints and  without (script :  https://github.com/chemicalfiend/md-constraints/blob/main/benchmarks/shake-bonds.jl) : 
Fri Aug 12 09:18:45 AM IST 2022

I think it does okay on time, but is a bit memory intensive. After completing some work with the other constraints, I can come back around and work on making these a bit more efficient.

jgreener64 (commented):
Looks good, I have a few more comments as it is starting to take shape. The benchmark results seem fine for now.

One thing that did occur is what happens if an atom appears in multiple constraints? Currently you cycle through the atom pairs once so an atom could be moved from a fulfilled constraint. How do other packages deal with this? Does it make sense to cycle repeatedly until all constraints are satisfied simultaneously?

chemicalfiend:
Made the requested changes. New benchmarks (same blob : https://github.com/chemicalfiend/md-constraints/blob/main/benchmarks/shake-bonds.jl) :

Sat Aug 13 10:26:03 AM IST 2022

Looks like you were right about `masses` being the cause of the memory issue, so it's all good.

In response to your question about atoms being in multiple constraints : 

* From my understanding, as per the LAMMPS implementation that I'm following, which constrains clusters, the atoms involved in every constraint are marked and flagged and the iterations proceed till all constraints are simultaneously satisfied. In particular, I felt the LAMMPS implementation of SHAKE was more accurate since it accounted for the quadratic term in the update equation.

* The GROMACS implementation of SHAKE (see here : https://github.com/gromacs/gromacs/blob/bd791c7cba317d8d59c86c2edcc923d6263f8d0a/src/gromacs/mdlib/shake.cpp/#L259-L346) while ignoring the quadratic term in the same equation (from the same paper), does use an iterative procedure that involves convergence across all constraints.

* My understanding is that if we think of the coordinate updates on a per-atom basis instead of a per-constraint basis, we would essentially have to apply this equation in effect : 

Sat Aug 13 10:36:52 AM IST 2022

Where the sum is over all constraints. Since the updates happen at every iteration, I suppose the coordinates of the atom are bumped each time a constraint is met, so this equation is effectively carried out in the procedure followed here as well as in the GROMACS implementation.

jgreener64:
In this implementation though you could have atoms ending up not satisfying constraints. For example:
```julia
using Molly

constraint_list = (SHAKE([0.1u"nm", 0.1u"nm"], [1, 2], [2, 3]),)

coords = [
    SVector(0.88, 1.0, 1.0)u"nm",
    SVector(1.00, 1.0, 1.0)u"nm",
    SVector(1.12, 1.0, 1.0)u"nm",
]
boundary = CubicBoundary(2.0u"nm", 2.0u"nm", 2.0u"nm")

sys = System(
    atoms=fill(Atom(), 3),
    constraints=constraint_list,
    coords=coords,
    boundary=boundary,
)

apply_constraints!(sys, sys.coords, 0.002u"ps")

ds = distances(sys.coords, boundary)

println(ds[1, 2]) # 0.1150000000000001 nm
println(ds[2, 3]) # 0.09999999999999987 nm
```
I think the approach should iterate until all are satisfied.

chemicalfiend:
I added a line to debug the thing : 

Wed Aug 17 03:38:45 PM IST 2022

And it seems to have all the constraints taken care of. Though when I ran the same line of code alongside the system in your comment : 


```julia
using Molly
using LinearAlgebra
constraint_list = (SHAKE([0.1u"nm", 0.1u"nm"], [1, 2], [2, 3]),)

coords = [
    SVector(0.88, 1.0, 1.0)u"nm",
    SVector(1.00, 1.0, 1.0)u"nm",
    SVector(1.12, 1.0, 1.0)u"nm",
]
boundary = CubicBoundary(2.0u"nm", 2.0u"nm", 2.0u"nm")

sys = System(
    atoms=fill(Atom(), 3),
    constraints=constraint_list,
    coords=coords,
    boundary=boundary,
)

apply_constraints!(sys, sys.coords, 0.002u"ps")

ds = distances(sys.coords, boundary)

lengths = []

push!(lengths, norm(vector(sys.coords[1], sys.coords[2], sys.boundary)))

push!(lengths, norm(vector(sys.coords[1], sys.coords[2], sys.boundary)))

println(lengths[1])
println(lengths[2])

```

The output comes out as follows :

Wed Aug 17 04:01:47 PM IST 2022

I'm getting the feeling the order of steps performed in simulate!( ) may be more relevant here than iterating till all constraints are satisfied (I tried implementing it such that it iterates and the output still comes out the same). 

I'm not able to pinpoint exactly where it's going wrong in the pipeline though....

chemicalfiend:
Yep, looks like you were right. It needed iterations. It was scrambled up in the implementation I followed....I sorted it out and implemented it. Thank you so much for pointing this out. I have it ready. I'll clean up the implementation and submit it by tonight.

PS: @jgreener64 I posted a comment here earlier, if you got an email notification about it or something, just ignore it.

chemicalfiend:
Made the change. If you have any other concerns or additions, let me know.

src/constraints.jl
```diff
@@ -0,0 +1,86 @@
...
+    end
+end
+
+"""
+    SHAKE(d, bond_list)
+
+Constrains a set of bonds to a particular distance d.
+"""
+struct SHAKE{D, B}
+    d::D
```
jgreener64, line 24: This could be called `dists`.

It would also be good to have a constructor from keyword arguments like https://github.com/JuliaMolSim/Molly.jl/blob/master/src/interactions/harmonic_bond.jl#L17.

The docstring also needs updating.

src/constraints.jl
```diff
@@ -0,0 +1,86 @@
...
+    apply_constraints!,
+    SHAKE
+
+
+"""
+    apply_constraints!(system, old_coords, dt)
+
+Applies all the bond and angle constraints associated with the system.
+"""
+
```
jgreener64, line 11: Remove this line or the docstring won't be associated with the function.

src/constraints.jl
```diff
@@ -0,0 +1,86 @@
...
+                sys.coords[i0] += δri0
+                sys.coords[i1] += δri1
+            end
+    
+        end
+
+        lengths = [norm(vector(sys.coords[constraint.is[r]], sys.coords[constraint.js[r]], sys.boundary)) - constraint.d[r] for r in 1:length(constraint.is)]
+        
+        if ustrip(maximum(abs.(lengths))) < 1e-10
```
jgreener64, line 82: It would be good for this to be a field `tolerance` of `SHAKE`, with a default of `1e-10u"nm"` and no `ustrip`.

src/types.jl
```diff
@@ -541,10 +546,12 @@ function ReplicaSystem(;
     k_converted = convert_k_units(T, k, energy_units)
     K = typeof(k_converted)
 
-    replicas = Tuple(System{D, G, T, CU, A, AD, PI, SI, GI, C, V, B, NF,
+    constraints = ()
```
jgreener64, line 549: This could be a keyword argument to the `ReplicaSystem` constructor, and add it to the docstring.

src/constraints.jl
```diff
@@ -0,0 +1,86 @@
...
+            r01 = vector(old_coords[i1], old_coords[i0], sys.boundary)
+
+            # Distance vector after unconstrained update
+            s01 = vector(sys.coords[i1], sys.coords[i0], sys.boundary)
+        
+            if abs(ustrip(norm(s01) - constraint.d[r])) > 1e-10
+                m0 = mass(sys.atoms[i0])
+                m1 = mass(sys.atoms[i1])
+                a = (1/m0 + 1/m1)^2 * norm(r01)^2
+                b = 2.0 * (1/m0 + 1/m1) * dot(r01, s01)
```
jgreener64, line 54: `2.0` -> `2` to avoid promotion of `Float32`s.

src/constraints.jl
```diff
@@ -0,0 +1,86 @@
...
+                c = norm(s01)^2 - ((constraint.d[r])^2)
+                D = (b^2 - 4*a*c)
+                
+                if (ustrip(D) < 0.0)
+                    @warn "SHAKE determinant negative. Setting to 0.0"
+                    D = 0.0u"nm^4/u^2"
```
jgreener64, line 60: `D = zero(D)` would deal better with the no units case.

src/constraints.jl
```diff
@@ -0,0 +1,86 @@
...
+                δri0 = r01.*(g/m0)
+                δri1 = r01.*(-g/m1)
+
+                sys.coords[i0] += δri0
+                sys.coords[i1] += δri1
+            end
+    
+        end
+
+        lengths = [norm(vector(sys.coords[constraint.is[r]], sys.coords[constraint.js[r]], sys.boundary)) - constraint.d[r] for r in 1:length(constraint.is)]
```
jgreener64, line 80: Need `abs` here.

test/simulation.jl
```diff
@@ -416,6 +416,81 @@ end
...
+                             )
+                )
+
+       
+    for i in 1:length(sys.coords)
+        sys.coords[i] += [rand()*0.01, rand()*0.01, rand()*0.01]u"nm"        
+    end
+
+    old_coords = sys.coords
+    apply_constraints!(sys, sh, old_coords, 0.002u"ps")
```
jgreener64, line 469: We should probably test a case where iteration is required, i.e. where each atom can be in multiple constraints. Perhaps the test of a single `apply_constraints!` call could be done on a different set of atoms with a water-like bonding arrangement, i.e. constraints between 1-2, 2-3, 4-5, 5-6, etc. You could generate the coordinates with `place_atoms`, record the constraint distances from these coordinates, add random noise, call `apply_constraints!`, then check the constraint distances are matched.

jgreener64 (commented):
Getting there - once these comments are addressed it can be merged.

chemicalfiend:
Made the requested changes.

jgreener64 (commented):
Looks ready, you will need to resolve conflicts with master before I can merge.

chemicalfiend:
Resolved conflicts. You can check and merge it accordingly

jgreener64:
Nice one :tada:

Would be good to get the velocity update done next? That is an important part of the implementation.

chemicalfiend:
That's the plan. Will get the velocity update along with SETTLE for water in the next PR. I have the codes ready, just tiny bit of debugging and testing needed.

PR #86, opened 2022-08-14
REMD generalization and H-REMD

JaydevSR:
This PR adds a generalization to the REMD simulator defined before and based on that implements a H-REMD simulator.


JaydevSR:
Any suggestions on test simulation for H-REMD?

JaydevSR:
I noticed some bugs here and there so I am making it a draft or for now.

jgreener64:
Looks good so far. Regarding testing, I guess it depends whether we have the interactions available to run a typical H-REMD simulation. In that case we could test the behaviour matches what is expected. If a suitable example can't be found or takes too long to run, we could just check for an expected number of exchanges like for T-REMD. However this kind of algorithm does lend itself to subtle bugs so ideally we would have a relevant test.

About adding softness to the LJ/Coulomb potentials, this could be done but I think they should be new interactions to keep the current implementations simple and fast.

src/simulators.jl
```diff
@@ -508,26 +508,160 @@ function simulate!(sys::ReplicaSystem{D, G, T},
...
+        for i in eachindex(sys.replicas)
+            random_velocities!(sys.replicas[i], sim.temperatures[i]; rng=rng)
+        end
+    end
+
+    simulate_remd!(sys, sim, n_steps, tremd_exchange!; rng=rng, n_threads=n_threads)
+        end
+
+function tremd_exchange!(sys::ReplicaSystem{D,G,T},
+                        sim::TemperatureREMD,
```
JaydevSR, line 521:
```suggestion
                        sim::TemperatureREMD{T},
```

Should this be done (at other places in the file as well) in order to ensure that float type of system and simulator is same?

jgreener64: At the moment we leave dealing with float types to the user and allow them to use any combination they want, so I think leave it out for now.

JaydevSR:
I am writing some simulation test right now using the softcore potentials. To add some unit test akin to `TemperatureREMD`, I am checking the mean of the temperatures to fall within a certain range. Can I write similar tests on Total/Potential energy (i.e. are those expected to follow some trend based on the reaction coordinate say the $\lambda$ value of softcore potentials?

jgreener64:
As I understand it the soft core potentials are supposed to reduce the height of energy barriers. I'm not sure how the global values of the energy are supposed to differ on average; maybe look at plots of energy v distance (https://juliamolsim.github.io/Molly.jl/stable/examples/#Variations-of-the-Morse-potential) and potential energy of the system over time.

You would expect to see closer minimum pair distances and less hard "bouncing" with a softer core, if you can think of a way to test for that.

JaydevSR:
I was not able to think of any other tests for H-REMD that are not too complicated to implement, so for now only testing for exchange efficiency and temperatures. I have added a TODO comment just in case we can add something in the future.

jgreener64:
Looks good to me :tada:

JaydevSR:
@jgreener64 Now, I was wondering what can I do next. I can try adding more unit test and add some improvements to the docs. And also I can add some REMD simulator if that is not too complicated (considering the time frame of the GSoC project). Something like hybrid temperature and Hamiltonian scheme (which will only require changing a few lines in the T-REMD exchange function) if this sort of simulator is required. As the current interface is very general it should be straightforward to add new REMD simulator to the library in the future and for users to define their own REMD schemes.

PR #87, opened 2022-08-15
correct a typo

ehsanirani:
found a typo in `cosine_angle.jl`, being corrected here.


jgreener64:
Surprised that one slipped through, thanks.

PR #89, opened 2022-08-18
Add `LennardJonesSoftCore` and `CoulombSoftCore`

JaydevSR:
This PR adds soft core version of the Lennard jones and Coulomb pairwise interactions following the expressions in [Beutler, T. C.; Mark, A. E.; van Schaik, R. C.; Gerber, P. R.; van Gunsteren, W. F. Chem. Phys. Lett. 1994, 222, 529−539](https://doi.org/10.1016/0009-2614(94)00397-1).

### DIFF
--- a/src/interactions/coulomb.jl
+++ b/src/interactions/coulomb.jl
@@ -1,4 +1,4 @@
-export Coulomb
+export Coulomb, CoulombSoftCore
 
 @doc raw"""
     Coulomb(; cutoff, nl_only, weight_14, coulomb_const, force_units, energy_units)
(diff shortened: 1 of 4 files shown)


src/interactions/coulomb.jl
```diff
@@ -115,3 +115,137 @@ end
...
+    CoulombSoftCore(; cutoff, sc_softness, sc_lambda, sc_power, nl_only, lorentz_mixing, weight_14,
+                    coulomb_const, force_units, energy_units)
+
+The Coulomb electrostatic interaction between two atoms with a soft core.
+The potential energy is defined as
+```math
+V(r_{ij}) = \frac{q_i q_j}{4 \pi \varepsilon_0 (r_{ij}^6 + \alpha * sigma_{ij}^6 * \lambda^p)^{\frac{1}{6}}}
+```
+
+Here, ``\\alpha``, ``\\lambda``, and ``\\p`` are `sc_softness`, `sc_lambda`, and `sc_power` respectively which are used
```
jgreener64, line 129: Given that we generally use Greek characters elsewhere for keyword arguments I think these can be `α`, `λ` and `p`.

JaydevSR:
In the added tests, the values which are being tested against were explicitly calculated for the particular case of $\alpha=1$, $\lambda=0.5$ and $p=2$ using the formula by me. So I think this address all the comments.

jgreener64:
:tada:

PR #91, opened 2022-08-19
Fix inline math for docstring

JaydevSR:
Escaping not required as using `@doc raw`.


JaydevSR:
I seem to be missing a lot of such small things 😮‍💨

jgreener64:
We all miss such things.

PR #95, opened 2022-08-25
Speed up for softcore potentials and fixes to Mie

### LINKED ISSUE #94: `Mie` interaction gives `InexactError` due to the `convert` in constructor
If we use some value for `m` and `n` such that `(n / (n - m)) * (n / m) ^ (m / (n - m))` gives a floating point number then, https://github.com/JuliaMolSim/Molly.jl/blob/7d95bcc27217343cf0283e10dcc13da3ccf378fa/src/interactions/mie.jl#L37 leads to inexact error (if `typeof(m)` is `Int`). Maybe it is better to do something like:

```julia
m, n, mn_fac = promote(m, n, (n / (n - m)) * (n / m) ^ (m / (n - m)))
```

JaydevSR:
Initial implementation of `LennardJonesSoftCore` and `CoulombSoftCore` was quite slow (several microseconds for single force calculation). This PR add a speed up to make it useable (now the times are about 150ns per force calculation, 3x slower than non-softcore interactions but still way better than before).

Also fixes #94 .

### DIFF
--- a/src/interactions/mie.jl
+++ b/src/interactions/mie.jl
@@ -34,7 +34,7 @@ function Mie(;
                 force_units=u"kJ * mol^-1 * nm^-1",
                 energy_units=u"kJ * mol^-1",
                 skip_shortcut=false)
-    mn_fac = convert(typeof(m), (n / (n - m)) * (n / m) ^ (m / (n - m)))
+    m, n, mn_fac = promote(m, n, (n / (n - m)) * (n / m) ^ (m / (n - m)))
     return Mie{skip_shortcut, typeof(cutoff), typeof(m), typeof(force_units), typeof(energy_units)}(
         m, n, cutoff, nl_only, lorentz_mixing, force_units, energy_units, mn_fac)
 end
--- a/src/interactions/mie.jl
+++ b/src/interactions/mie.jl
@@ -34,9 +34,9 @@ function Mie(;
                 force_units=u"kJ * mol^-1 * nm^-1",
                 energy_units=u"kJ * mol^-1",
                 skip_shortcut=false)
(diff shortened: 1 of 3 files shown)


src/interactions/mie.jl
```diff
@@ -34,7 +34,7 @@ function Mie(;
                 force_units=u"kJ * mol^-1 * nm^-1",
                 energy_units=u"kJ * mol^-1",
                 skip_shortcut=false)
-    mn_fac = convert(typeof(m), (n / (n - m)) * (n / m) ^ (m / (n - m)))
+    m, n, mn_fac = promote(m, n, (n / (n - m)) * (n / m) ^ (m / (n - m)))
```
jgreener64, line 37: Maybe give the `m` and `n` local variables different names here since they can change type.

jgreener64 (commented):
Looks good, thanks.

src/interactions/mie.jl
JaydevSR: Do you mean?

```julia
function Mie(;
                m,
                n,
                cutoff=NoCutoff(),
                nl_only=false,
                lorentz_mixing=true,
                force_units=u"kJ * mol^-1 * nm^-1",
                energy_units=u"kJ * mol^-1",
                skip_shortcut=false)
    m_p, n_p, mn_fac = promote(m, n, (n / (n - m)) * (n / m) ^ (m / (n - m)))
    return Mie{skip_shortcut, typeof(cutoff), typeof(m), typeof(force_units), typeof(energy_units)}(
        m_p, n_p, cutoff, nl_only, lorentz_mixing, force_units, energy_units, mn_fac)
end
```

PR #97, opened 2022-09-04
RATTLE constraint for velocities

chemicalfiend:
I implemented the RATTLE constraint which constrains velocities along constraints to be 0. While the test works for one usage of `apply_constraints` works, it is not converging for multiple steps (refer the test here : https://github.com/chemicalfiend/md-constraints/blob/main/test/rattle-test.jl ; the first 2 tests pass, but the `simulate!` command keeps running forever)

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,15 @@
 # Molly.jl release notes
 
+## v0.14.1 - Jan 2023
+
+### New features
+- The Buckingham potential is added as `Buckingham`.
+- Polymer melt and density functional theory examples are added to the documentation.
+
+### Bug fixes
+- A bug introduced by a new version of AtomsBase.jl is fixed. In addition, atoms in systems without `atoms_data` defined now return `:unknown` from `AtomsBase.atomic_symbol`.
+- A bug in the export of `OpenMMResidueType` is fixed.
+
 ## v0.14.0 - Dec 2022
 
 ### Breaking changes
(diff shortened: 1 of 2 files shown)


jgreener64:
Sounds like `maximum(vpset)` never gets low enough in the case where it fails. If you print out `maximum(vpset)` each iteration what does it give you? 10^-15 may be too strict a requirement.

Also, do you have a link to the implementation you are basing this on?

chemicalfiend:
1e-15 was too strict a requirement, indeed. I think because the units of v . r is nm^2 / ps we can only expect it to converge to about an order of O(10^-3) . I tried printing out maximum(vpset) and no matter what I did it seemed to only hit roughly 1e-2 nm^2/ps for each of the the test cases I was working on, so I relaxed the tolerance bit by bit until it worked. I'm concerned it might be a little too relaxed now...

As for references, [this](https://github.com/lammps/lammps/blob/b4e42f68b14c348203ba8e58f1ef27dc58f21a24/src/RIGID/fix_rattle.cpp/#L312-L356) has the general implementation for pairs; I used the iteration scheme prescribed [here](https://github.com/gromacs/gromacs/blob/bd791c7cba317d8d59c86c2edcc923d6263f8d0a/src/gromacs/mdlib/shake.cpp/#L348-L422) ; I also referred the paper [here](https://doi.org/10.1016/0021-9991(83)90014-1) directly .

jgreener64:
Thanks for the links. Yes I think that is too relaxed a requirement.

Is the line `r01 = vector(old_coords[i1], old_coords[i0], sys.boundary)` and the corresponding line after correct? The `vector` function arguments would be `i0` then `i1` if you wanted the vector from `i0` to `i1`. Here they are the other way around. I notice it is the same order as for SHAKE but there you are mainly taking `norm`s or doing other operations that would cancel out an error.

PR #100, opened 2022-09-14
Update docs.md

mi-ayala:
min_dist is not a keyword argument anymore.


JaydevSR:
Hi @mi-ayala, can you open an issue regarding implementing such a potential. I am sure we can discuss on this and see what works there (instead in a PR thread). Also, please share the link to the paper you are referring to in the issue as it seems that email attachments are not visible on GitHub. Thanks.

PR #104, opened 2022-11-04
Monte-Carlo for molecules

JaydevSR:
This PR adds some simple Monte-Carlo functionality i.e., Metropolis Monte-Carlo for molecules.  I was initially thinking of writing the simulator in `simulators.jl` but there may be some monte-carlo specific code (like the random translation trial moves here) that doesn't fit there so I added a new file.

Some things that need to be done before merge:

- [x] Use of neighbor finder for calculating potential energy at each step?
- [ ] Each step uses `copy()` which is not very efficient, any suggestions for improvement?
- [x] Example simulations (one example given below)
- [x] Docs and Tests

I have run a test simulation given below:
```julia
n_atoms = 100
atom_mass = 10.0u"u"
atoms = [Atom(mass=atom_mass, charge=10, σ=0.3u"nm", ϵ=0.2u"kJ * mol^-1") for i in 1:n_atoms]
boundary = CubicBoundary(2.0u"nm", 2.0u"nm", 2.0u"nm")
coords = place_atoms(n_atoms, boundary; min_dist=0.3u"nm")
pairwise_inters = (Coulomb(),)

temp_mult = [1598.0u"K", 1198.0u"K", 798.0u"K", 398.0u"K", 198.0u"K", 98.0u"K", 8.0u"K"]
sys = System(
	  atoms=atoms,
	  pairwise_inters=pairwise_inters,
	  coords=coords,
	  boundary=boundary,
	  loggers=(coords=CoordinateLogger(n_atoms),),
)

trial_args = Dict(:shift_scaling => 0.1, :length_units => u"nm")
for t in temp_mult
    sim = MetropolisMonteCarlo(; 
		    temperature=t,
		    trial_moves=random_normal_translation!,
		    trial_args=trial_args
	    )
    
    simulate!(sys, sim, 20000)
end
```

This is what the simulation looks like:

https://user-images.githubusercontent.com/59474411/199914858-3d8d077e-9701-4ad5-abcd-48c06018acc4.mp4

As expected the atoms are arranging into a lattice (sort of). Another simulation that was suggested by @jgreener64 was to sample the torsional angle of ethane. I am not quite sure how to set that up so some help will be appreciated.


jgreener64:
Looks like a great start. If the lattice formation is clearer in 2D you could do that. I think the `copy()` is okay for the moment, profiling would help point to the slow steps though.

> Use of neighbor finder for calculating potential energy at each step?

I think call `find_neighbors` every step and use the neighbours, as the other simulators do. In many cases, for example with large trial moves, not using a neighbour list would be more efficient since neighbour calculation would need to be run every step. It can be left to the user though since the default is `NoNeighborFinder`.

> sample the torsional angle of ethane

I can look at this myself after the PR is merged if it's easier.

A couple of other points:
- The `potential_energy` only needs to be calculated once per step, with `E_old` being updated at the end of each step.
- It would also be nice to have a logger similar to the `ReplicaExchangeLogger` that logs energies and acceptance/failure.
- For the random translation functions I would make the norm of the displacement match the stated uniform/Gaussian distribution, rather than doing it separately by dimension. This can be done by generating x/y/z displacements with `randn`, dividing by the norm to get a unit vector, and multiplying by the distance drawn from the appropriate distribution.

JaydevSR:
> It would also be nice to have a logger similar to the ReplicaExchangeLogger that logs energies and acceptance/failure.

So, for the logger I am thinking that instead of logging the energies for which we already have a logger, we should log the value of $\displaystyle\frac{E}{k_B T}$ i.e. the argument of the boltzmann rates. This makes sense to log for a Monte-Carlo simulation. Also should I just log the number of acceptances and selections or also a bit array of when the acceptances occur?

JaydevSR:
A 2D simulation which shows lattice formation with 100 atoms in 4nm square boundary:

https://user-images.githubusercontent.com/59474411/200545210-7b8d61c5-e23a-445a-8eef-eda561721dd0.mp4

Similarly for 3D with 4nm boundary and 100 atoms:

https://user-images.githubusercontent.com/59474411/200545464-c5f4fffe-c61d-4933-a8db-a0df5a1383b9.mp4

jgreener64:
Yes I think logging that value sounds good. I would log the bit array of when the acceptances occur too.

JaydevSR:
Now the question remains is where the logger would go. If we put it in loggers named tuple in the System type, then we cannot pass in info about whether the acceptance occurred or not. And also, it would recalculate any potential energy that we have already calculated in the metropolis part. In the replica exchange, I added a field to replica system. So, should there be a similar optional thing for the System type, like a `montecarlo_logger` field?

jgreener64:
This seems to be a general problem when writing simulator-specific loggers. I think we should change the arguments of the `run_loggers!` and `log_property!` function definitions to accept additional `kwargs...`. In this case we could then pass the acceptance info and potential energy as keyword arguments to `run_loggers!` and use them in the relevant `log_property!` function. I think this should mostly be a case of adding `kwargs...` to the keyword arguments of functions in `src/loggers.jl`.

JaydevSR:
I have added most of the things that are needed for the PR ✌️. Please tell me if some more tweaks are needed or if I missed something.

src/montecarlo.jl
```diff
@@ -0,0 +1,103 @@
...
+"""
+    MetropolisMonteCarlo(; <keyword arguments>)
+
+A Monte Carlo simulator for [`System`](@ref) that uses the Metropolis algorithm to sample the configuration space.
+The method of `simulate!` for this simulator accepts an optional keyword argument `log_states::Bool=true` which 
+tells whether to run the loggers or not (for example, during equilibration).
+
+# Arguments
+- `temperature::T`: The temperature of the system.
+- `log_interval::Int`: The interval at which to log the system state.
```
jgreener64, line 14: I think these are out of date.

src/montecarlo.jl
```diff
@@ -0,0 +1,103 @@
...
+        end
+    end
+end
+
+"""
+	random_uniform_translation!(sys::System; scaling=1.0, length_units=unit(sys.coords[1][1]))
+
+Performs a random translation of the coordinates of a randomly selected atom in `sys`. 
+The translation is generated using a uniformly selected direction and uniformly selected length 
+in range [-0.5, 0.5] scaled by `scaling` and with units `length_units`.
```
jgreener64, line 67: I think this should read `in range [0.0, 1.0]`.

test/simulation.jl
```diff
@@ -720,6 +720,48 @@ end
...
+    n_steps = 10_000
+    atom_mass = 10.0u"u"
+    atoms = [Atom(mass=atom_mass, charge=1.0, σ=0.3u"nm", ϵ=0.2u"kJ * mol^-1") for i in 1:n_atoms]
+    boundary = CubicBoundary(4.0u"nm", 4.0u"nm", 4.0u"nm")
+    coords = place_atoms(n_atoms, boundary; min_dist=0.3u"nm")
+
+    temp = 198.0u"K"
+
+    nb_matrix = trues(n_atoms, n_atoms)
+    for i in 1:(n_atoms ÷ 2)
```
jgreener64, line 734: This block isn't required since there are no bonds.

test/simulation.jl
```diff
@@ -720,6 +720,48 @@ end
...
+
+    simulator = MetropolisMonteCarlo(
+        temperature=temp,
+        trial_moves=random_normal_translation!,
+        trial_args=Dict(:scaling => 0.1, :length_units => u"nm")
+    )
+
+    @time simulate!(sys, simulator, n_steps; log_states=false)
+    @time simulate!(sys, simulator, n_steps; log_states=true)
+    @info "Acceptance Rate: $(sys.loggers.mcl.n_trials / sys.loggers.mcl.n_accept)"
```
jgreener64, line 762: Is there anything that can be tested here? Such as the acceptance rate being in a certain range, the ending energy being lower than the starting energy, or the mean distance to nearest neighbour having a certain value?

src/montecarlo.jl
```diff
@@ -0,0 +1,103 @@
...
+        else
+            sys.coords = coords_old
+            log_states && run_loggers!(sys, neighbors, i; n_threads=n_threads, success=false,
+                        energy_rate=E_old / (k_b * sim.temperature))
+        end
+    end
+end
+
+"""
+	random_uniform_translation!(sys::System; scaling=1.0, length_units=unit(sys.coords[1][1]))
```
jgreener64, line 63: I think `scaling` and `length_units` can be one argument, something like `shift_size`.

jgreener64 (commented):
Looks good, just some small things before merging.

JaydevSR:
I have added tests for acceptance rate (test for bad cases), avg potential energy (should decrease from a random config at start to end for a repulsive system), and the mean nearest neighbor distance (for a repulsive system it should not be less than the Wigner-Seitz radius and for the given boundary it is not possible for it to be 2 times the Wigner-Seitz radius).

jgreener64:
Great, thanks for this.

PR #105, opened 2022-12-31
Started Muller-Brown potential

ejmeitz:
Hello,

Just started working on the Muller-Brown (MB) potential as a first project so this is definitely a draft. I had some questions about how this potential should be incorporated into Molly. MB is not like any of the other potentials in Molly at the moment, it is only 2D and is not a pairwise interaction (just a surface). Assuming I can figure out how to integrate MB into Molly I will add test coverage in a future commit.

Questions:
1. Other interactions are typed as SpecificInteraction or PairwiseInteraction, MB potential does not fit into either of these. Would I need to create something new (e.g. PotentialEnergySurface)
2. Because it is only 2D how do I make sure this potential is operable within the framework of Molly that is mostly built around pairwise interactions or bonded interactions (e.g. 2 atoms get passed to the potential() function). I have also not implemented a force() function yet. I do not really see how to make the interface used by other potentials amenable to MB.
3. I'm pretty new to Julia so I have a lot of uncertainty around the typing and in general how you go about a large code base without OOP. My brain is stuck thinking in abstract classes and interfaces so any pointers are appreciated.

Thanks,
Ethan


JaydevSR:
Hi @ejmeitz, I think you are looking to implement something referred to as a General Interaction in Molly's framework. One example is:
https://github.com/JuliaMolSim/Molly.jl/blob/9ced8f7261807467f928b4ff207f0fac96f4c525/src/interactions/implicit_solvent.jl#L338

I think a potential surface can be implemented similarly. As for the typing, the basic idea is to use variable type parameters in the struct definitions and then in the respective constructor we figure out if the types of the inputs to the constructor make sense and if they do we initialize the struct with those parameters and inputs. If this is done right everything works together nicely. You can see this in all the types (structs) defined in Molly. Take a look at the file src/types.jl for numerous examples.

I think @jgreener64 will be able to clarify further if I am wrong.

jgreener64:
Thanks for giving this a go @ejmeitz.

@JaydevSR is correct, this can be implemented as a general interaction (https://juliamolsim.github.io/Molly.jl/dev/docs/#General-interactions). In the first instance you can write a function that takes in a 2D coordinate and gives the force, and also one for the energy. These functions can then be broadcasted over the system coordinates inside the `forces` and `potential_energy` functions. Broadcasting works here since the force/energy only depends on the particle being considered.

ejmeitz:
@JaydevSR @jgreener64 thank you for the tips! I think I have the code conforming to the framework correctly now. I'm working on tests, but still had a few questions about the code I wrote.

1. Should I enforce that an input has certain units? e.g. the `A` vector should have units of energy, `x0` units of length etc. Or is that on the user to keep consistent?
2. Should I enforce that the user only inputs 2D data into this system? The broadcasting in the forces and potential_energy is agnostic of the dimension of each sub-array currently, but should only work for 2D.
3. Formats for output of forces() and potential_energy()
- Should PE be a single number for whole system or a single number for each atom
- Force should have info for every particle but does the type of the output matter as long as its some kind of nested array?

I can probably answer 3 myself, but I am still figuring out how to get my dev version of Molly into a Julia environment for local testing.

jgreener64:
> Should I enforce that an input has certain units?

Leave it to the user to ensure consistency, this is usually fine since arithmetic with inconsistent units will throw an error.

> Should I enforce that the user only inputs 2D data into this system?

If you replace `coord` with `coord::SVector{2}` in the arguments of `potential` then an error will be thrown in non-2D cases.

> Should PE be a single number for whole system or a single number for each atom

For the whole system.

> Force should have info for every particle but does the type of the output matter as long as its some kind of nested array?

Should be `SVector{2, T}` for each atom (`T` is determined by the force unit).

ejmeitz:
@jgreener64 pretty sure the code is all correct now, but I can't figure out how to get the codecov tests to pass. I added some tests and they cause the ubuntu check to fail. For now they're just commented out at the bottom of interactions.jl in my last commit. Once I figure out this last piece I plan on starting the Gay-Berne and NEQUIP potentials or a NoseHoover thermostat.

I also tried plotting the trajectories as a sanity check for myself to see if they match [this ](https://www.wolframcloud.com/objects/demonstrations/TrajectoriesOnTheMullerBrownPotentialEnergySurface-source.nb)Wolfram notebook. In my simulation the particle just goes to the bottom of the energy potential (see below) instead of oscillating around the minimum which was a little un-expected. The energy contours and forces looked alright so I'm not sure why I did not get the correct dynamics (used VV integrator for a few million steps).

image

jgreener64:
You can run the tests locally or check the logs (https://github.com/JuliaMolSim/Molly.jl/actions/runs/3833098088/jobs/6524174032) to see the issue. In this case testing that `force(inter, local_min2)` is approximately equal to `0.0u"kJ * mol^-1 * nm^-1"` fails since the first is an `SVector{2}` and the second is a single value.

> In my simulation the particle just goes to the bottom of the energy potential (see below) instead of oscillating around the minimum which was a little un-expected.

This will depend on the simulator you use, I don't know which one the Wolfram notebook used. You would see more oscillation with a Berendsen thermostat for instance.

The code looks broadly okay, I'll add some review comments once tests pass.

ejmeitz:
Couldn't get tests for "potential" to work as that function is not exposed to the test set (or maybe I'm being dumb). Can also write an example for docs if necessary.

src/interactions/muller_brown.jl
```diff
@@ -0,0 +1,85 @@
...
+where A, a, b, c, $x_0$, $y_0$ are 4 element arrays
+"""
+#To-Do
+# A should have energy_units, in general are units working
+# Test coverage
+# Example for documentation
+# Should potential return just a number?
+
+# General interaction:
+# https://juliamolsim.github.io/Molly.jl/dev/docs/#General-interactions
```
jgreener64, line 19: Remove these lines.

src/interactions/muller_brown.jl
```diff
@@ -0,0 +1,85 @@
...
+struct MullerBrown{D,T,L,F,E}
+    A::D # Array w/ units of energy
+    a::T # Array w/ units of 1/L^2
+    b::T # Array w/ units of 1/L^2
+    c::T # Array w/ units of 1/L^2
+    x₀::L # Array w/ units of L
```
jgreener64, line 25: I would call these `x0` and `y0` for simplicity.

src/interactions/muller_brown.jl
```diff
@@ -0,0 +1,85 @@
...
+    y₀::L # Array w/ units of L
+    force_units::F
+    energy_units::E
+end
+
+
+function MullerBrown(A, a, b, c, x₀, y₀;
```
jgreener64, line 32: The standard values of these parameters can be given here as default arguments, i.e. `; A=SVector(-200.0, -100.0, -170.0, 15.0)u"kJ * mol^-1"`. Having them as `SVector`s will improve performance.

src/interactions/muller_brown.jl
```diff
@@ -0,0 +1,85 @@
...
+    if length(A) != 4 || length(a) != 4 || length(b) != 4 || length(c) != 4 || length(x₀) != 4 || length(y₀) != 4
+        throw(ArgumentError("The length of each array passed to the Muller-Brown potential should be 4."))
+    end
+
+    return MullerBrown{typeof(A), typeof(a), typeof(x₀), typeof(force_units), typeof(energy_units)}(
+        A, a, b, c, x₀, y₀, force_units, energy_units)
+end
+
+#Total potential energy of system
+@inline @inbounds function potential_energy(inter::MullerBrown{D,T,L}, sys, neighbors=nothing) where {D,T,L}
```
jgreener64, line 45: Whilst it isn't doing much harm, you don't need the `{D,T,L}` here since it isn't used in the function. Also applies to other functions.

src/interactions/muller_brown.jl
```diff
@@ -0,0 +1,85 @@
...
+    return sum(potential.(Ref(inter),sys.coords))
+end
+
+@inline @inbounds @fastmath function potential(inter::MullerBrown{D,T,L}, coord::SVector{2}) where {D,T,L}
+    
+    x,y = coord
+
+    res = 0.0 * inter.energy_units #give result units of energy
```
jgreener64, line 53: It's a bit annoying to have to do this but making it `ustrip(zero(coord[1])) * inter.energy_units` means simulations using `Float32` won't get promoted to `Float64` by `0.0`. Also applies below to force.

src/interactions/muller_brown.jl
```diff
@@ -0,0 +1,85 @@
...
+    res_x = 0.0 * inter.force_units #give result units of force
+    res_y = 0.0 * inter.force_units
+    for i in 1:4
+        dx = x - inter.x₀[i]
+        dy = y - inter.y₀[i]
+        exp_part = inter.A[i]*exp(inter.a[i]*(dx^2) + inter.b[i]*dx*dy + inter.c[i]*(dy^2))
+        res_x += exp_part*(2*inter.a[i]*dx + inter.b[i]*dy)
+        res_y += exp_part*(inter.b[i]*dx + 2*inter.c[i]*dy)
+    end
+    return SVector{2}([-res_x, -res_y])
```
jgreener64, line 84: `SVector(-res_x, -res_y)` avoids an array allocation.

src/interactions/muller_brown.jl
```diff
@@ -0,0 +1,85 @@
...
+    return MullerBrown{typeof(A), typeof(a), typeof(x₀), typeof(force_units), typeof(energy_units)}(
+        A, a, b, c, x₀, y₀, force_units, energy_units)
+end
+
+#Total potential energy of system
+@inline @inbounds function potential_energy(inter::MullerBrown{D,T,L}, sys, neighbors=nothing) where {D,T,L}
+    return sum(potential.(Ref(inter),sys.coords))
+end
+
+@inline @inbounds @fastmath function potential(inter::MullerBrown{D,T,L}, coord::SVector{2}) where {D,T,L}
```
jgreener64, line 49: Since `potential` is specific to Muller-Brown it might be better named as `potential_muller_brown`. Also applies to `force`. You can access it in tests with `Molly.potential_muller_brown` - this is required since it isn't exported.

jgreener64 (changes requested):
Looks good, I made a few suggestions.

The other thing that would be nice in addition to the potential energy test is a test in the simulations file that runs a short simulation and checks that it finds the minimum. Pretty much what you plotted above.

ejmeitz:
@jgreener64  Renaming the force function to force_muller_brown breaks something in the simulation loop. Molly is trying to call force() but finds nothing. If I leave it as force things seem to run fine. It does not seem to care what I call the potential energy function. All tests pass if I do not rename but I want to make sure I understand how everything is interacting here. Like you said I thought forces() and potential_energy() would be called not force() and potential().

Thanks, Ethan

jgreener64:
If you rename `force` to `force_muller_brown` then simulation should not be calling `force` at all, so look at the line number in the error stacktrace and see where that is. Maybe you forgot to change where the function is broadcasted in `forces`? Potential energy is not called during simulation unless you use an energy logger.

ejmeitz:
Any idea why the minimization test fails on my laptop but passed in the GitHub checks? I get an error on line 62: `@test all(x -> isapprox(x, 0.4u"nm"; atol=1e-3u"nm"), dists_flat)` which is kinda weird cause I never touch that code and it passes on github. I think the tests I added should pass but am unable to get past the minimization test on my laptop for now....will try again on desktop later.

ejmeitz:
Yeah I ran is like 4 times last night and printed them they are all within the expected tolerance. Here's what I get re-running this morning:  `[0.4000429540445738 nm, 0.4004080941145199 nm, 0.40004454772593906 nm]` and it still fails even though the expected val is 0.4 and the atol is 0.001

Using intel i7-10510U and integrated graphics.

jgreener64:
That's strange, does this return `true` or `false`? What about without the unit annotations?
```julia
using CUDA, Unitful
dists_flat = CuArray([0.4000429540445738, 0.4004080941145199, 0.40004454772593906]u"nm")
all(x -> isapprox(x, 0.4u"nm"; atol=1e-3u"nm"), dists_flat)
```

ejmeitz:
It returns true with and without units. Super weird, tests passed on GitHub though.

jgreener64:
Let's just call it weirdness for now.

Thanks for doing this PR!

PR #110, opened 2023-02-01
Crystal Structure Generation

ejmeitz:
Hello,

I've been working on some code to generate crystal structures and get back a list or coordinates, atoms and the boundary needed to run the simulation. The code is more or less untested and I just wanted to discuss the implementation of the interface. There are many types in this file, but the only ones I intend to export are the functions at the bottom of the file which are not fully implemented. When complete these functions should cover all possible Bravais lattices in 3D and provide sufficient coverage of crystal structure for as a starting place. As it stands, the functions at the bottom of the file (e.g. FCC, Diamond) are just functions that return a crystal type for you. Then you'd have to call replicate_unit_cell to get back the atoms, coords and boundary. I'd prefer this just be one call to FCC() that gives you atoms, coords and boundary,

Some background:

A crystal is defined by a lattice (repeating grid of points) and a basis (atomic pattern at each lattice point). As such the crystal type takes a BravaisLattice, and Vector of BasisAtoms in its constructor. The way in which the Bravais lattice is initialized specifies whether it will be FCC or BCC etc. and the basis atoms allow for more complicated structure like Diamond or Wurtzite to be constructed.


ejmeitz:
Yeah that is a good idea. I'll probably call it Bravais.jl.

I guess leave this pull request open and eventually I'll replace this code with something that calls that library.

jgreener64:
Sounds good. Regarding naming there is also the Atomic or Atom prefix, e.g. AtomicCrystals.jl.

ejmeitz:
Code is mostly done, just making the ReadMe pretty and making sure everything is actually correct: https://github.com/ejmeitz/SimpleCrystals.jl

Thinking about how to interface with Molly now.

ejmeitz:
Are you ok with me adding a second constructor for the System() struct that takes a Crystal struct from the library I made? It would just end up calling the original constructor after it parses the Crystal struct for useful infos.

jgreener64:
Yes should be fine.

ejmeitz:
Cool, almost working now...

One last thing before I fully test it out, what is the difference between fields 1,2, & 5 in AtomData? and what are the res_number and res_name? I have data about the element like its name, symbol etc just confused about what should go where. Thanks!

```
atom_type::String
atom_name::String
res_number::Int
res_name::String
element::String
```

jgreener64:
These bits of data are quite specific to biomolecular force fields. In that context you have residues forming chains to make proteins. You can just leave them with the default values apart from the element.

`atom_type` - the force field contains parameters (e.g. `σ`/`ϵ`) for each atom type.
`atom_name` - the name of the atom in the residue template, e.g. `CD` is the delta carbon.
`res_number` - the residue number in the chain.
`res_name` - the name of the residue the atom belongs to, e.g. `ALA`.
`element` - the element of the atom as a string, e.g. `C`.

ejmeitz:
Stuck at the moment...adding my package to Molly breaks compilation for some reason. Specifically it cannot find the `dualize` function from ForwardDiff imported into Molly in zygote.jl. I'm a bit confused because my package has no dependencies related to auto-diff packages.

The whole package dev thing still confuses me a little but I'm a bit stumped here. I did have to change Molly's [compat] for AtomsBase to 0.3 but I dont see why that would break dualize.

ejmeitz:
Might delete this branch and restart, idk what happened when I pulled all the CUDA branch. Seems broken though lol.

PR #118, opened 2023-03-09
Hotfix for Julia 1.9

axsk:
Using the new extension system, it seems the `dualize` function defined in `FordwardDiff`'s `ForwardDiffStaticArraysExt` extension module is not importable (at least I could not manage to do it).

However, since the signature defined here (in Molly) is only used from within Zygote, I supposed it's safe to simply define it independent of `ForwardDiff`'s `dualize`. 
Maybe this `dualize` could/should be called `offset_dualize` to clearly distinguish it?

Using this little trick allowed me to load Molly on 1.9.0rc1.
However, I couldn't run `Pkg.test` because it fails due to some GL stuff not working (running it on a remote node without even an display server) and I was not using the differentiable capabilities so far, so I cannot say anything about correctness.

### DIFF
--- a/src/zygote.jl
+++ b/src/zygote.jl
@@ -3,7 +3,6 @@
 # Here be dragons
 
 using ForwardDiff: Chunk, Dual, partials, value
-using ForwardDiff.ForwardDiffStaticArraysExt: dualize
 using Zygote: unbroadcast
 
 iszero_value(x::Dual) = iszero(value(x))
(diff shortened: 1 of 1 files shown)


jgreener64:
Thanks for fixing this. I asked on Slack and it seems that this function is indeed not importable. Looks good to me.

On machines where there is no display you can run the tests with `VISTESTS=0 julia test/runtests.jl`.

PR #120, opened 2023-03-21
Add Side Length checker

exAClior:
I added checks for side length to be greater than zero for `CubicBoundary` and `RectangularBoundary`.

I did two unit tests locally. One of them gave an error shown below. Looks like zygote might create a `CubicBoundary` with negative side length. However, I was not able to reproduce this again. 
```
Gradients: Error During Test at /Users/yushengzhao/projects/Molly.jl/test/zygote.jl:1
  Got exception outside of a @test
  AssertionError: Side lengths need to be larger than 0
  Stacktrace:
    [1] CubicBoundary
      @ ~/projects/Molly.jl/src/spatial.jl:38 [inlined]
    [2] accum(x::NamedTuple{(:side_lengths,), Tuple{SizedVector{3, Float64, Vector{Float64}}}}, y::SVector{3, Float64})
      @ Molly ~/projects/Molly.jl/src/zygote.jl:137
    [3] (::Zygote.var"#back#291"{:contents, Zygote.Context{false}, Core.Box, CubicBoundary{Float64}})(Δ::SVector{3, Float64})
      @ Zygote ~/.julia/packages/Zygote/xGkZ5/src/lib/lib.jl:237
    [4] (::Zygote.var"#2149#back#292"{Zygote.var"#back#291"{:contents, Zygote.Context{false}, Core.Box, CubicBoundary{Float64}}})(Δ::SVector{3, Float64})
      @ Zygote ~/.julia/packages/ZygoteRules/OgCVT/src/adjoint.jl:71
    [5] Pullback
      @ ~/projects/Molly.jl/test/zygote.jl:136 [inlined]
    [6] (::typeof(∂(λ)))(Δ::Float64)
      @ Zygote ~/.julia/packages/Zygote/xGkZ5/src/compiler/interface2.jl:0
    [7] (::Zygote.var"#68#69"{typeof(∂(λ))})(Δ::Float64)
      @ Zygote ~/.julia/packages/Zygote/xGkZ5/src/compiler/interface.jl:45
    [8] gradient(::Function, ::Float64, ::Vararg{Float64})
      @ Zygote ~/.julia/packages/Zygote/xGkZ5/src/compiler/interface.jl:97
    [9] macro expansion
      @ ~/projects/Molly.jl/test/zygote.jl:196 [inlined]
   [10] macro expansion
      @ /Applications/Julia-1.8.app/Contents/Resources/julia/share/julia/stdlib/v1.8/Test/src/Test.jl:1363 [inlined]
   [11] top-level scope
      @ ~/projects/Molly.jl/test/zygote.jl:2
   [12] include(fname::String)
      @ Base.MainInclude ./client.jl:476
   [13] top-level scope
      @ ~/projects/Molly.jl/test/runtests.jl:82
   [14] include(fname::String)
      @ Base.MainInclude ./client.jl:476
   [15] top-level scope
      @ none:6
   [16] eval
      @ ./boot.jl:368 [inlined]
   [17] exec_options(opts::Base.JLOptions)
      @ Base ./client.jl:276
   [18] _start()
      @ Base ./client.jl:522
Test Summary: | Pass  Error  Total     Time
Gradients     |    9      1     10  1m41.3s
ERROR: LoadError: Some tests did not pass: 9 passed, 0 failed, 1 errored, 0 broken.
in expression starting at /Users/yushengzhao/projects/Molly.jl/test/zygote.jl:1
in expression starting at /Users/yushengzhao/projects/Molly.jl/test/runtests.jl:81
ERROR: Package Molly errored during testing
```

### DIFF
--- a/src/spatial.jl
+++ b/src/spatial.jl
@@ -27,17 +27,15 @@ Setting one or more values to `Inf` gives no boundary in that dimension.
 """
 struct CubicBoundary{T}
     side_lengths::SVector{3, T}
+    CubicBoundary(side_lengths::SVector{3, E}) where E = (
+        all(a -> a >= 0*unit(a),  side_lengths) ?
+        new{E}(side_lengths) :
+        throw(DomainError("Side lengths need to be positive, got $side_lengths"))
+    )
 end
 
-function CubicBoundary(x, y, z)
-    @assert all(a -> a > 0*unit(a), [x, y, z]) "Side lengths need to be larger than 0, got [$x,$y,$z]"
-    CubicBoundary(SVector{3}(x, y, z))
-end
-
-function CubicBoundary(arr)
-    @assert all(x -> x > 0*unit(x), arr) "Side lengths need to be larger than 0, got $arr"
-    CubicBoundary(SVector{3}(arr))
-end
+CubicBoundary(x, y, z) = CubicBoundary(SVector{3}(x, y, z))
+CubicBoundary(arr) = CubicBoundary(SVector{3}(arr))
 
 Base.getindex(b::CubicBoundary, i::Integer) = b.side_lengths[i]
 Base.firstindex(b::CubicBoundary) = b.side_lengths[1]
@@ -52,16 +50,15 @@ Setting one or more values to `Inf` gives no boundary in that dimension.
 """
 struct RectangularBoundary{T}
     side_lengths::SVector{2, T}
+    RectangularBoundary(side_lengths::SVector{2, E}) where E = (
+        all(a -> a >= 0 * unit(a), side_lengths) ?
+        new{E}(side_lengths) :
+        throw(DomainError("Side lengths need to be positive, got $side_lengths"))
+    )
 end
 
-function RectangularBoundary(x, y)
-    @assert all(a -> a > 0*unit(a), [x, y]) "Side lengths need to be larger than 0, got [$x, $y]"
-    RectangularBoundary(SVector{2}(x, y))
-end
-function RectangularBoundary(arr)
-    @assert all(x -> x > 0*unit(x), arr) "Side lengths need to be larger than 0, got $arr"
-    RectangularBoundary(SVector{2}(arr))
-end
+RectangularBoundary(x, y) = RectangularBoundary(SVector{2}(x, y))
+RectangularBoundary(arr) = RectangularBoundary(SVector{2}(arr))
 
 Base.getindex(b::RectangularBoundary, i::Integer) = b.side_lengths[i]
 Base.firstindex(b::RectangularBoundary) = b.side_lengths[1]
--- a/src/spatial.jl
+++ b/src/spatial.jl
@@ -30,12 +30,12 @@ struct CubicBoundary{T}
 end
 
 function CubicBoundary(x, y, z)
-    @assert all(a -> a > 0, [x, y, z]) "Side lengths need to be larger than 0"
+    @assert all(a -> a > 0*unit(a), [x, y, z]) "Side lengths need to be larger than 0"
     CubicBoundary(SVector{3}(x, y, z))
 end
 
 function CubicBoundary(arr)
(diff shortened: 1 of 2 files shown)


JaydevSR:
Shouldn't this be a `DomainError`? It seems more appropriate for the cases when only certain set of arguments are allowed.

jgreener64 (changes requested):
Thanks for doing this.

Indeed it could be a `DomainError`. Also, the check can be made once as part of an [inner constructor](https://docs.julialang.org/en/v1/manual/constructors/#man-inner-constructor-methods) in the struct definition.

With regards to the error, I just remembered that the gradient code uses boundaries with zero length sides to accumulate gradients. So maybe just check that side lengths aren't negative to allow this to work.

JaydevSR:
> > Shouldn't this be a `DomainError`? It seems more appropriate for the cases when only certain set of arguments are allowed. Also, array construction (i.e. `[x, y, z]` etc.) can also be easily skipped in the condition by using `&&`s instead, not that it would make a big difference but still 🤔
> 
> You are absolutely correct about both. Thank you so much for pointing them out! I considered other ways instead of doing `[x,y,z]` but nothing gives a clean-looking code since I would need to include the `unit(x)` for all three sides. Or maybe I am misinterpreting your suggestion of using `&&`?

This is not an issue here infact it will make no difference at all as this is just setup code for a simulation. But usually we try to avoid array allocation in simulation codes so I just pointing it out out of a habit 😅. So you can easily ignore this comment.

exAClior:
> Thanks for doing this.
> 
> Indeed it could be a `DomainError`. Also, the check can be made once as part of an [inner constructor](https://docs.julialang.org/en/v1/manual/constructors/#man-inner-constructor-methods) in the struct definition.
> 
> With regards to the error, I just remembered that the gradient code uses boundaries with zero length sides to accumulate gradients. So maybe just check that side lengths aren't negative to allow this to work.

Thank you so much for pointing out the inner constructor, I think it solves some of the concerns about using `[x,y,z]` Hopefully, the code is now a bit neater. 

I also allowed negative values to pass too.

src/spatial.jl
```diff
@@ -27,6 +27,11 @@ Setting one or more values to `Inf` gives no boundary in that dimension.
 """
 struct CubicBoundary{T}
     side_lengths::SVector{3, T}
+    CubicBoundary(side_lengths::SVector{3, E}) where E = (
+        all(a -> a >= 0*unit(a),  side_lengths) ?
+        new{E}(side_lengths) :
+        throw(DomainError("Side lengths need to be positive, got $side_lengths"))
```
jgreener64, line 33: Maybe change to "Side lengths can't be negative" to match the check.

src/spatial.jl
```diff
@@ -27,6 +27,11 @@ Setting one or more values to `Inf` gives no boundary in that dimension.
 """
 struct CubicBoundary{T}
     side_lengths::SVector{3, T}
+    CubicBoundary(side_lengths::SVector{3, E}) where E = (
+        all(a -> a >= 0*unit(a),  side_lengths) ?
```
jgreener64, line 31: You can use `zero(a)` rather than `0*unit(a)`.

src/spatial.jl
```diff
@@ -27,6 +27,11 @@ Setting one or more values to `Inf` gives no boundary in that dimension.
 """
 struct CubicBoundary{T}
     side_lengths::SVector{3, T}
+    CubicBoundary(side_lengths::SVector{3, E}) where E = (
```
jgreener64, line 30: This could be written in the standard function form, i.e. `function`, for readability. Also with an explicit `if`/`else`.

jgreener64 (changes requested):
Looks good, just minor changes.

src/spatial.jl
exAClior: My bad, forgot to change this to include the zero case.

src/spatial.jl
exAClior: Indeed, I have made the change. Thanks for pointing it out.

PR #123, opened 2023-04-27
Fix GitHub Pages Link

ejmeitz:
Fix default link for GitHub pages on the Molly github page.


jgreener64:
Might be better to point to the stable docs, otherwise looks okay.

PR #125, opened 2023-05-07
Crystals v2

ejmeitz:
Don't even know what happened in the last branch. Unbroke whatever merge issues I had from pulling the CUDA branch.


ejmeitz:
Notes:
- Possible type conflict between SimpleCrystals.Atom & Molly.Atom
- Default values for AtomsData? Just copied from class rn.
- Class for replica system too?

ejmeitz:
Maybe something like:

```julia
function Atom(;
                index=1,
                type_id=1)
    return Atom(index, type_id)
end

AtomsDictionary = {"1" : AtomsData(...), "2" : AtomsData(...)}
```

ejmeitz:
@jgreener64 if you get the chance could you look at the constructors I added to types.jl and help me figure out why they're type unstable? Both constructors I added here have unstable types and it kills the run-time speed.

jgreener64:
What do you mean by type instability in this context? Can you give example code that shows it?

ejmeitz:
Yeah for sure, below is an example of constructing a system from a 4x4x4 unit-cell FCC system. The construction of the System struct is type-unstable and when I call simulate!() with this struct the run-times are much slower than I expect. The results of the simulation are what I expect but the output from @code_warntype on the construction of the System indicates that the compiler cannot assign concrete types to everything.

```julia
fcc_crystal = FCC(5.4u"Å", :Ar, SVector(4,4,4))

n_atoms = length(fcc_crystal)
atom_mass = atomic_mass(fcc_crystal,1)
velocities = [random_velocity(atom_mass, temp) for i in 1:n_atoms]

sys = @code_warntype System(
    fcc_crystal,
    velocities=velocities,
    pairwise_inters=(LennardJones(),),
    loggers=(ke=KineticEnergyLogger(10),
            pe=PotentialEnergyLogger(10)),
    energy_units = u"kJ * mol^-1",
    force_units = u"kJ * mol^-1 * Å^-1")
```
If you actually want this code to run with LJ you have to use that extra convenience constructor and modify all the atoms in the system to have the proper sigma and epsilon values.
```julia
σ = 3.4u"Å" 
ϵ = (4.184*0.24037)u"kJ * mol^-1"
updated_atoms = []
for i in range(1,length(sys.atoms))
    push!(updated_atoms, Molly.Atom(index = sys.atoms[i].index, charge = sys.atoms[i].charge,
     mass = sys.atoms[i].mass, σ = σ, ϵ = ϵ, solute = sys.atoms[i].solute))
end

sys = System(sys, atoms = updated_atoms)
```

jgreener64:
The constructor itself can be type unstable since it is only called once at the start. The concern is what is returned, i.e. `typeof(sys)`. In the first case this is concrete so all is well - here this can be checked by seeing that all the type parameters in `typeof(sys)` are concrete.

The problem is that `updated_atoms = []` creates a `Vector{Any}` which is not concrete. A quick fix is `sys = System(sys, atoms = [updated_atoms...])`.

ejmeitz:
Cool yeah 


> [updated_atoms...]

fixed the call to simulate! which is what matters. I guess I'm unhappy now that the user has to avoid that pitfall or will suffer 50% performance loss.

In general, the interface works from the few system I've ran but I still am not satisfied with the extra work/knowledge put onto the user with the current implementation. The code 2 comments above is basically what you will have to do every time. Open to suggestions of how I could change SimpleCrystals or the code I've added to Molly. For now my only idea requires breaking changes to Molly (moving $\sigma$ and $\epsilon$ and similar params to potential definitions.

jgreener64:
σ and ϵ only need to be set in cases where they will be used, i.e. there is a Lennard Jones or similar interaction. I am less familiar with these systems - how often would a Lennard Jones-type interaction be used when simulating? Always, sometimes or never?

Moving the parameters to the interactions is probably the right way to go long term but requires a lot of changes and careful thought about how to not make setup harder in other contexts.

jgreener64:
I'm not going to stop you giving it a go, but I would warn you that it is really a lot of work and probably not worth it just to make the above setup process easier. There are considerations about differentiable simulations and GPU performance, for example. It might be better to just document the setup procedure clearly.

The lack of properties for Buckingham potentials etc. is an issue but it is possible to use a custom atom type without a performance penalty, one is used in the tests for Buckingham:https://github.com/JuliaMolSim/Molly.jl/blob/2c0631401a2b51d5215308fc1d877456b8f60411/test/interactions.jl#L80-L85

ejmeitz:
Sounds good, I guess for now I'll just write documentation and add some tests.

ejmeitz:
Tests should pass, in conclusion this PR adds:

- Integration with[ SimpleCrystals.jl](https://github.com/ejmeitz/SimpleCrystals.jl)
-  Docs + Tests
- Convenience constructor for `System` built from another `System` object

Writing the tests for this made me realize that my NoseHoover implementation doesn't work for solids, made another PR to add that to the docs. Might need help to fix that.

test/simulation.jl
```diff
@@ -1044,3 +1044,79 @@ end
...
+        dt=2u"fs",
+        temperature=temp,
+        friction=1.0u"ps^-1",
+    )
+
+    simulate!(sys, simulator, 100_000, run_loggers = false)
+    simulate!(sys, simulator, 200_000)
+ 
+    #Integrator is stochastic so give a wide berth on the tolerance
+    @test  -1850 < ustrip(mean(values(sys.loggers.tot_eng))) < -1650
```
jgreener64, line 1097: Minor nitpick here - test against the unitful energy rather than the `ustrip`ped value, this tests against accidental unit conversions.

test/simulation.jl
```diff
@@ -1044,3 +1044,79 @@ end
...
+
+
+    #Test Unsupported Crystals
+    hex_crystal = SimpleCrystals.Hexagonal(a, :Ar, SVector(2,2))
+    try
```
jgreener64, line 1102: Use `@test_throws`, currently no error will skip the test. Also below.

src/types.jl
```diff
@@ -566,6 +566,116 @@ function System(;
...
+
+A physical system to be simulated, constructed from a SimpleCrystals.jl `Crystal` struct.
+Properties unused in the simulation or in analysis can be left with their
+default values.
+`atoms`, `atoms_data`, `coords` and `boundary` are automatically calcualted from the cyrstal struct. Extra
+atom paramaters like `σ` will have to be added manually after construction using the convenience constructor:
+System(system; <keyword arguments>)
+
+
+This is a sub-type of `AbstractSystem` from AtomsBase.jl and implements the
```
jgreener64, line 580: No need for this or the below arguments line, this docstring gets appended to the previous.

src/types.jl
```diff
@@ -566,6 +566,116 @@ function System(;
...
+    coords = SimpleCrystals.position(crystal, :)
+
+    # Build bounding box
+    side_lengths = norm.(eachrow(bounding_box(crystal)))
+    if any(typeof(crystal.lattice.crystal_family) .<: [CubicLattice, OrthorhombicLattice, TetragonalLattice])
+        boundary = CubicBoundary(side_lengths...)
+    elseif any(typeof(crystal.lattice.crystal_family) .<: [SquareLattice, RectangularLattice])
+        boundary = RectangularBoundary(side_lengths...)
+    elseif D == 2 #Honeycomb, Hex2D, & Oblique
+        throw(error("$(crystal.lattice.crystal_family) is not supported as it would need a 2D triclinic boundary.
```
jgreener64, line 614: Use `ArgumentError`.

src/types.jl
```diff
@@ -566,6 +566,116 @@ function System(;
...
+             Try defining the crystal with a rectangular or square unit cell."))
+    else #3D non-cubic systems
+        @assert all(crystal.lattice.crystal_family.lattice_angles .< 90u"°") "All crystal lattice angles must be less than 90°"
```
jgreener64, line 617: `ArgumentError` if this is a valid crystal that the constructor does not support, `error` otherwise.

jgreener64 (changes requested):
Looks good, just a few changes.

PR #129, opened 2023-05-22
Add option to disable loggers in all Simulator structs

ejmeitz:
Thoughts on adding `run_loggers` parameter to all Simulator structs? I noticed it exists in the `StepestDescentMinimizer` but not other things like `VelocityVerlet`.

I ran into a use case where I would like to disable logging for the first 300k steps of a VV scheme and then re-enable it after the system has come to the correct temperature. Could also be worth allowing users to change what gets logged in-between calls to `simulate!` within the same code (maybe you can do this already I dunno). Something like `modify_loggers!(sys, ...)`. I also think there's a lot of performance to be gained here, maybe by writing to a file instead of RAM, just a thought.

I have not gone thru and made all the necessary changes just wanted to open this to discussion.

### DIFF
--- a/src/simulators.jl
+++ b/src/simulators.jl
@@ -47,17 +47,18 @@ function SteepestDescentMinimizer(;
 end
 
 """
-    simulate!(system, simulator, n_steps; n_threads=Threads.nthreads(), run_loggers = Bool)
-    simulate!(system, simulator; n_threads=Threads.nthreads(), run_loggers = Bool)
+    simulate!(system, simulator, n_steps; n_threads=Threads.nthreads(), run_loggers = true)
+    simulate!(system, simulator; n_threads=Threads.nthreads(), run_loggers = true)
 
-Run a simulation on a system according to the rules of the given simulator.
+Run a simulation on a system according to the rules of the given simulator. `run_loggers` is true
+by default except for the `SteepestDescentMinimizer`.
 
 Custom simulators should implement this function.
 """
 function simulate!(sys,
                     sim::SteepestDescentMinimizer;
                     n_threads::Integer=Threads.nthreads(),
-                    run_loggers = false)
+                    run_loggers=false)
     sys.coords = wrap_coords.(sys.coords, (sys.boundary,))
     neighbors = find_neighbors(sys, sys.neighbor_finder; n_threads=n_threads)
     run_loggers && run_loggers!(sys, neighbors, 0; n_threads=n_threads)
@@ -126,7 +127,7 @@ function simulate!(sys,
                     sim::VelocityVerlet,
                     n_steps::Integer;
                     n_threads::Integer=Threads.nthreads(),
-                    run_loggers = true)
+                    run_loggers=true)
     sys.coords = wrap_coords.(sys.coords, (sys.boundary,))
     !iszero(sim.remove_CM_motion) && remove_CM_motion!(sys)
     neighbors = find_neighbors(sys, sys.neighbor_finder; n_threads=n_threads)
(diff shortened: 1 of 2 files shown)


jgreener64:
Yes I'm up for adding `run_loggers` to all the simulators. It is on the to-do list but I haven't got round to it yet. It would be a good PR if you have the time.

ejmeitz:
Added `run_loggers` to every simulator besides the REMD ones as those looked a little different and I did not feel confident modifying that code. Also thinking about adding `enable_loggers(sim)`, `disable_loggers(sim)` and `modify_loggers(sys)`. The first two I am not sure are possible because the simulation structs are not mutable.

Default is set to true for everything besides `SteepestDescentMinimizer`

ejmeitz:
Nah the REMD code was pretty easy to read I've just never looked at it before. I made changes to that in a separate commit if you would like to see #2268ccd. Let me know if that looks alright.

I also moved run_loggers to simulate! instead of being a simulator property. This way its not in an immutable struct and can actually be toggled.

src/simulators.jl
```diff
@@ -29,40 +29,38 @@ Not currently compatible with automatic differentiation using Zygote.
...
-                                    run_loggers=false,
                                     log_stream=devnull)
-    return SteepestDescentMinimizer(step_size, max_steps, tol, run_loggers, log_stream)
+    return SteepestDescentMinimizer(step_size, max_steps, tol, log_stream)
 end
 
 """
-    simulate!(system, simulator, n_steps; n_threads=Threads.nthreads())
-    simulate!(system, simulator; n_threads=Threads.nthreads())
+    simulate!(system, simulator, n_steps; n_threads=Threads.nthreads(), run_loggers = Bool)
```
jgreener64, line 50: `run_loggers=true`. Then below could add "`run_loggers` is `true` by default except for `SteepestDescentMinimizer`".

src/simulators.jl
```diff
@@ -29,40 +29,38 @@ Not currently compatible with automatic differentiation using Zygote.
...
 
 Run a simulation on a system according to the rules of the given simulator.
 
 Custom simulators should implement this function.
 """
 function simulate!(sys,
                     sim::SteepestDescentMinimizer;
-                    n_threads::Integer=Threads.nthreads())
+                    n_threads::Integer=Threads.nthreads(),
+                    run_loggers = false)
```
jgreener64, line 60: No spaces round `=` in arguments for consistency.

src/simulators.jl
```diff
@@ -728,7 +734,7 @@ function simulate!(sys::ReplicaSystem,
         end
     end
     
-    return simulate_remd!(sys, sim, n_steps; rng=rng, n_threads=n_threads)
+    return simulate_remd!(sys, sim, n_steps; rng=rng, n_threads=n_threads, run_loggers)
```
jgreener64, line 737: `run_loggers=run_loggers` for consistency (and above).

src/simulators.jl
```diff
@@ -766,15 +772,16 @@ function remd_exchange!(sys::ReplicaSystem{D, G, T},
...
 
 Run a REMD simulation on a [`ReplicaSystem`](@ref) using a REMD simulator.
 """
 function simulate_remd!(sys::ReplicaSystem,
                         remd_sim,
                         n_steps::Integer;
                         rng=Random.GLOBAL_RNG,
-                        n_threads::Integer=Threads.nthreads())
+                        n_threads::Integer=Threads.nthreads(),
+                        run_loggers = true)
```
jgreener64, line 784: This should be passed through to the `Threads.@spawn simulate!` call to control those loggers too. With that I think the REMD strategy is okay, i.e. `run_loggers` turns on/off both types of logger. Maybe we could add a `run_exchange_loggers` later if required.

src/simulators.jl
```diff
@@ -860,7 +867,7 @@ function simulate!(sys::System{D, G, T},
                    sim::MetropolisMonteCarlo,
                    n_steps::Integer;
                    n_threads::Integer=Threads.nthreads(),
-                   log_states::Bool=true) where {D, G, T}
+                   run_loggers = true) where {D, G, T}
```
jgreener64, line 870: Remove the reference to `log_states` from the docstring. Also there is a test that calls `log_states` which will need to change.

jgreener64 (changes requested):
Looks good, some minor comments.

Issue #128, opened 2023-05-20
AtomsBase not properly implemented


mfherbst:
Just a remark that I noted that the latest AtomsBase interface (0.3) is not properly supported by Molly. E.g. `hasatomkey(sys, :testkey)` on a Molly system gives an infinite stacktrace.

We should take a look at this and integrate with [`AtomsBaseTesting`](https://juliamolsim.github.io/AtomsBase.jl/stable/testing/) in the CI of Molly.

jgreener64:
Thanks for pointing this out Michael - we'll add AtomsBaseTesting and fix any issues.

jgreener64:
Should be fixed by #131.

PR #131, opened 2023-05-29
AtomsBase testing

ejmeitz:
This PR will add better support of AtomsBase types and implement AtomsBaseTesting into the CI.

From #128


ejmeitz:
I think this is pretty much done from code that needs to be added to Molly. I made some PRs in AtomsBase to add features that Molly requires (e.g. Infinite boundaries) and to make the testing framework more flexible.

ejmeitz:
Asssuming the tests pass this is probably good enough to cover what @mfherbst was asking for.

Based on how this [PR ](https://github.com/JuliaMolSim/AtomsBase.jl/issues/74) ion AtomsBase turns out I might have to make modifications later. For now I think it is good enough but welcome suggestions/edits.

src/types.jl
```diff
@@ -834,16 +839,41 @@ The masses of the atoms in a [`System`](@ref) or [`ReplicaSystem`](@ref).
...
 Base.getindex(s::Union{System, ReplicaSystem}, i::Integer) = AtomView(s, i)
+
 Base.length(s::Union{System, ReplicaSystem}) = length(s.atoms)
 Base.eachindex(s::Union{System, ReplicaSystem}) = Base.OneTo(length(s))
 
 AtomsBase.species_type(s::Union{System, ReplicaSystem}) = typeof(s[1])
-AtomsBase.atomkeys(::Union{System, ReplicaSystem}) = ()
+AtomsBase.atomkeys(s::Union{System, ReplicaSystem}) = (:position, :velocity, :atomic_mass, :atomic_number)
+AtomsBase.hasatomkey(s::Union{System, ReplicaSystem}, x::Symbol) = x ∈ atomkeys(s)
+AtomsBase.keys(sys::System) = (:atoms,:coords,:boundary,:velocities,:atoms_data,:topology,
```
mfherbst, line 862: Can you not just get them from the `fieldnames` ?

src/types.jl
```diff
@@ -834,16 +839,41 @@ The masses of the atoms in a [`System`](@ref) or [`ReplicaSystem`](@ref).
...
+        :loggers, :k, :force_units, :energy_units)
+AtomsBase.keys(sys::ReplicaSystem) = (:atoms,:replica_coords,:boundary,
+        :n_replicas,:replica_velocities,:atoms_data,:topology,:replica_topology,
+        :pairwise_inters,:replica_pairwise_inters,:specific_inter_lists,
+        :replica_specific_inter_lists,:general_inters,:replica_general_inters,
+        :constraints,:replica_constraints,:neighbor_finder,:replica_loggers,
+        :exchange_logger,:k,:force_units,:energy_units)
+AtomsBase.haskey(sys::Union{System, ReplicaSystem}, x::Symbol) = hasfield(typeof(sys), x)
+Base.getindex(sys::Union{System, ReplicaSystem}, x::Symbol) = 
+    hasfield(typeof(sys), x) ? getfield(sys, x) : ArgumentError("No field `$x`. Allowed keys are $(keys(sys)).")
```
mfherbst, line 873: KeyError ?

src/types.jl
```diff
@@ -834,16 +839,41 @@ The masses of the atoms in a [`System`](@ref) or [`ReplicaSystem`](@ref).
...
+Base.pairs(sys::Union{System, ReplicaSystem}) = (k => sys[k] for k in keys(sys))
+Base.get(sys::Union{System, ReplicaSystem}, x::Symbol, default) = 
+    hasfield(typeof(sys), x) ? getfield(sys, x) : return default
```
mfherbst, line 876: no return needed.

mfherbst: Also I would use `haskey`

src/types.jl
```diff
@@ -877,6 +915,22 @@ function AtomsBase.atomic_symbol(s::Union{System, ReplicaSystem}, i::Integer)
     end
 end
 
+function AtomsBase.atomic_number(s::Union{System, ReplicaSystem})
+    if length(s.atoms_data) > 0
+        return map(ad -> ad.atomic_number, s.atoms_data)
+    else
+        return fill(:unknown, length(s))
```
mfherbst, line 922: I would also return `-1` for consistency.

src/types.jl
```diff
@@ -858,8 +888,16 @@ AtomsBase.velocity(s::ReplicaSystem, i::Integer) = s.replicas[1].velocities[i]
 AtomsBase.atomic_mass(s::Union{System, ReplicaSystem}) = masses(s)
 AtomsBase.atomic_mass(s::Union{System, ReplicaSystem}, i::Integer) = mass(s.atoms[i])
 
-AtomsBase.atomic_number(s::Union{System, ReplicaSystem}) = fill(missing, length(s))
-AtomsBase.atomic_number(s::Union{System, ReplicaSystem}, i::Integer) = missing
+# Base.getindex(sys::Union{System, ReplicaSystem}, i::Integer, x::Symbol) =
+#     getfield(getindex(sys, i), x)
+function Base.getindex(system::Union{System, ReplicaSystem}, i, x::Symbol)
+    if hasatomkey(system, x)
```
mfherbst, line 894: In contrast to what I said earlier, it might be better to explicitly list the mandatory atom keys here (the ones you currently put in `atomkeys`) and for `:charge` and similar do something else (e.g. explicitly insert the code that retrieves them).

mfherbst (commented):
Some comments from my end.

src/types.jl
ejmeitz: So basically manually add branches to the if statement for keys not in AtomsBase?

mfherbst: You could do that or alternatively you could use the trick I proposed only for a whitelisted set of keys (the mandatory ones).

ejmeitz: Ok cool that worked. Pushed changes.

I'm not the biggest fan of having to manually maintain that list of `custom_keys` (line 886) inside the `getindex` function. Maybe I'm missing a cleaner way to do that but it could also be useful add a `AtomsBase.extra_keys()` or something like that to specify extensions to the interface. This would also let you define the getindex function inside of the AtomsBase library that makes use of this function to define a generic `getidnex `for `AtomView`.

mfherbst: Hmm. Not fully sure if that might be over the top, but I'm open to discussing it further.

Certainly we should have programmatic ways to access the mandatory keys and the optional, standardised keys (like charge).

Would youmind opening an issue at atomsbase? I feel this is also an excellent topic for a JuliaCon hackaton. Would you agree?

ejmeitz: Yea I do, I'll open a PR.

src/types.jl
```diff
@@ -283,15 +286,17 @@ struct AtomData
     res_number::Int
     res_name::String
     element::String
+    atomic_number::Int
```
jgreener64, line 289: Isn't this redundant with `element`? If we add it here it has to be set up properly elsewhere, which would require a lookup using the element. It might be easier to remove this and have that lookup where this is used. https://github.com/JuliaPhysics/PeriodicTable.jl could be useful here.

src/types.jl
```diff
@@ -899,3 +950,79 @@ end
...
+Converts AtomsBase `AbstractSystem` to Molly `System``. To add kwargs not present in the
+AtomsBase interface (e.g. pair potentials) call the convenience constructor:
+
+`System(sys::System)`
+"""
+function System(sys::AbstractSystem{D}) where D
+
+    #Convert BC to Molly types
+    bb = bounding_box(sys)
+    boundary_conditions = AtomsBase.boundary_conditions(sys)
```
jgreener64, line 967: Use a different variable name here to avoid a clash with the function.

src/types.jl
```diff
@@ -899,3 +950,79 @@ end
...
+        if any(ustrip.(box_lengths) .== Inf)
+            throw(ArgumentError("Triclinic domain does not support infinite boundaries"))
+        end
+        molly_boundary = TriclinicBoundary(bb...)
+    else
+        throw(ArgumentError("Molly does not support 2D triclinic domains"))
+    end
+
+    idx = 0
+    atoms = map(sys) do atom
```
jgreener64, line 1004: `for` loop seems clearer if you are incrementing an index.

src/types.jl
```diff
@@ -899,3 +950,79 @@ end
...
+        idx += 1
+        Molly.Atom(; index = idx, charge = get(atom, :charge, 0.0), mass = atomic_mass(atom))
+    end
+
+    atoms_data = map(sys) do atom
+        AtomData(; element = String(atomic_symbol(atom)), atomic_number = atomic_number(atom))
+    end
+
+    coords = map(sys) do atom
```
jgreener64, line 1013: I think you can just use `position(sys)`, and similarly for velocity?

src/types.jl
```diff
@@ -899,3 +950,79 @@ end
...
+"""
+function System(sys::AbstractSystem{D}) where D
+
+    #Convert BC to Molly types
+    bb = bounding_box(sys)
+    boundary_conditions = AtomsBase.boundary_conditions(sys)
+    
+    #Check if box is cubic
+    angles = []
+    for i in range(1,D)
```
jgreener64, line 971: `1:D`

src/types.jl
```diff
@@ -899,3 +950,79 @@ end
...
+    end
+    isCubic = all(angles .== 0.0)
+
+    if !isinfinite(sys)
+        box_lengths = norm.(bb)
+    end
+    box_lengths = convert(Vector, box_lengths) #was SVector need to be mutable
+
+    if any(typeof.(boundary_conditions) .== DirichletZero)
+        raise(ArgumentError("Molly does not support DirichletZero boundary conditions."))
```
jgreener64, line 984: `throw`

jgreener64 (commented):
Thanks for improving compatibility here. Some of the stuff with keys and fields seems a little strange, that may be the price of supporting AtomsBase though. The ability to convert from an AtomsBase system is useful.

src/types.jl
ejmeitz: Have to define an empty array the other way and then push to it. This is less lines of code. Happy to change it though.

src/types.jl
ejmeitz: I guess it is redundant. `atomic_number `is just what the value is called in AtomsBase and its a required key for the system. I'll implement the `atomic_number` function to use PeriodicTable.jl.

ejmeitz: Actually, I just read the docs for AtomsBase both `element `and `atomic_number `are required keys:

The only required properties to be specified of the system is the species and implementations of standard functions accessing the properties of the species, currently

Geometric information: [position](https://juliamolsim.github.io/AtomsBase.jl/stable/apireference/#Base.position), [velocity](https://juliamolsim.github.io/AtomsBase.jl/stable/apireference/#AtomsBase.velocity), [n_dimensions](https://juliamolsim.github.io/AtomsBase.jl/stable/apireference/#AtomsBase.n_dimensions)
Atomic information: [atomic_symbol](https://juliamolsim.github.io/AtomsBase.jl/stable/apireference/#AtomsBase.atomic_symbol), [atomic_mass](https://juliamolsim.github.io/AtomsBase.jl/stable/apireference/#AtomsBase.atomic_mass), [atomic_number](https://juliamolsim.github.io/AtomsBase.jl/stable/apireference/#AtomsBase.atomic_number), [element](https://juliamolsim.github.io/AtomsBase.jl/stable/apireference/#AtomsBase.element)
Atomic and system property accessors: getindex, haskey, get, keys, pairs

src/types.jl
jgreener64: It may not be safe in future Julia versions, which could parallelise `map` or otherwise change the execution order.

src/types.jl
jgreener64: They are required for AtomsBase, but I think they are redundant in this struct. The atomic number could be retrieved from the element at the point where it is required in the AtomsBase function.

ejmeitz:
@jgreener64 all good!

test/basic.jl
```diff
@@ -419,3 +419,18 @@ end
...
+@testset "AbstractSystem -> Molly System" begin
+    system = make_test_system().system;
+    #Update values to be something that works with Molly
+    system = AbstractSystem(system; 
+                boundary_conditions = [Periodic(), Periodic(), Periodic()],
+                bounding_box = [[1.54732, 0.0, 0.0],
+                                [0.0, 1.4654985, 0.0],
+                                [0.0, 0.0, 1.7928950]]u"Å");
+    molly_sys = System(system)
+    test_approx_eq(system, molly_sys; common_only = true)
```
mfherbst, line 433: Sorry, only got round to look at this now. This is not the best way to do this. You can directly overwrite things (or drop keys) from `make_test_system` and then one should be able to use `test_approx_eq` directly without using the `common_only=true` flag.

PR #132, opened 2023-06-05
RATTLE Implementation

ejmeitz:
Implemented iterative scheme for RATTLE....completely untested but I have a many clarifying questions/thoughts.

- Its unclear to me if these implementations work with all integrators. The derivations all use VV. I don't think other MD codes limit you to this but I don't really see why these methods are generally applicable (they definitely shouldn't work for NVT & NPT out of the box)
    -  Related to this I'm not sure if RATTLE should be called after half-step updates or before or on both. I more or less get the math but not how these constraints should be integrated into a real MD code.
- I think the whole `apply_constraints!` interface will need to change as RATTLE and SHAKE have to be applied at different points in the integration process. That said not sure how to fix that at the moment in a pretty way.
- What does your SHAKE implementation follow? I used [this](https://www.sciencedirect.com/science/article/pii/0021999183900141?ref=pdf_download&fr=RR-2&rr=7d2484153c7239b2) for RATTLE which has the SHAKE steps in it also. I've never seen that quadratic formula to find g in any paper. Granted every paper I read does something totally different.
- LAMMPS docs have me a little confused. They say they solve RATTLE analytically (and the source matches that statement) but they also said they follow the Andersen paper which solves it iteratively. I think analytically would be better but not really clear how that is done.


jgreener64:
> I more or less get the math but not how these constraints should be integrated into a real MD code.

I agree this is not clear and much of implementing this correctly will be calling the constraint functions at the correct point. See for instance OpenMM, which has examples of where to call the constraints for [Verlet](https://github.com/openmm/openmm/blob/master/platforms/reference/src/SimTKReference/ReferenceVerletDynamics.cpp) and [Langevin](https://github.com/openmm/openmm/blob/master/platforms/reference/src/SimTKReference/ReferenceLangevinMiddleDynamics.cpp) dynamics. There is also a Julia RATTLE implementation [here](https://github.com/sunoru/MolecularDynamicsIntegrators.jl) but I don't know how good it is.

> I think the whole `apply_constraints!` interface will need to change as RATTLE and SHAKE have to be applied at different points in the integration process.

Yes I think we might want to switch to `constrain_coords!` and `constrain_velocities!` and call them at the appropriate points. Whether this transfers to other constraint algorithms remains to be seen, but it should be enough for SHAKE/RATTLE.

> What does your SHAKE implementation follow?

I think it follows LAMMPS, you might get more from the original PRs (#82 and #84).

> LAMMPS docs have me a little confused.

Look in the source code if possible?

ejmeitz:
Just some notes from LAMMPS SHAKE [source](https://github.com/lammps/lammps/blob/0978308f5e30dde39b3d625336f699cb4bcc3f16/src/RIGID/fix_shake.cpp#L4):
- Unconstrainted position update "assumes NVE update, seems to be accurate enough for NVT,NPT,NPH as well" Line 1685
- Shake bond constraint line [1774 ](https://github.com/lammps/lammps/blob/0978308f5e30dde39b3d625336f699cb4bcc3f16/src/RIGID/fix_shake.cpp#L1774)(our existing code is basically a copy of this)
- 2 & 3 bond clusters lines [1877 ](https://github.com/lammps/lammps/blob/0978308f5e30dde39b3d625336f699cb4bcc3f16/src/RIGID/fix_shake.cpp#L1877)& [2051](https://github.com/lammps/lammps/blob/0978308f5e30dde39b3d625336f699cb4bcc3f16/src/RIGID/fix_shake.cpp#L2051)
- 2 bonds 1 angle [line 2303](https://github.com/lammps/lammps/blob/0978308f5e30dde39b3d625336f699cb4bcc3f16/src/RIGID/fix_shake.cpp#L2303)
-

ejmeitz:
Useful thread from when this was implemented in LAMMPS, they're as confused as us :)

https://matsci.org/t/fix-shake-a-hybrid-method-between-shake-and-rattle/20507/25

ejmeitz:
Note: The replica system constructor will not parse the constraints correctly as it does not actually call the System constructor as a function rather it specifies all of the types explicitly.

ejmeitz:
Anyone have any clue what might be causing these particles to stick together? Its supposed to be hydrogen gas and all the constraints etc. work but this is clearly wrong. It also does not look like the rotational DoF are active. Based on the LAMMPS simulation I ran they should be spinning pretty fast.

https://github.com/JuliaMolSim/Molly.jl/assets/54505069/062ccc57-1254-436b-9c16-94d89c0367e6

ejmeitz:
Anyone have any clue why when I simulate hydrogen gas the rotational DoF don't get activated? The particle just moved around the box constrained interacting as if they cannot rotate.

I'm not really convinced this is an artifact of SHAKE but I don't know what else could cause it.

jgreener64:
Could you give code to reproduce the issue?

It might also be worth comparing to the same simulation without constraints, i.e. with a strong harmonic bond and a small time step. If that simulation doesn't show the issue then it is probably something to do with the constraints.

ejmeitz:
This is the input I am using. I uploaded the xyz file for this with 1 molecule (where you can also seem what I am talking about). In the 1 molecule case the two atom do not interact with each other so this rotation should just from the initial velocity. It might look like the molecule does not translate from the initial velocity but I think they do and its just small. The weirder thing to me is that the particles stop rotating. After looking at all the output I noticed that the velocity vectors eventually end up pointing towards the COM of the molecule effectively canceling eachother out. Why this is the case I have no idea. I tested the exact same system in LAMMPS and this does not happen. I plan to test the original implementation of SHAKE in Molly (I have only changed how it is initialized not the algorithm itself) to see if this phenomena occurs as well.

```julia

r_cut = 8.5u"Å"
temp = 300.0u"K"

#Initialize 10 random atoms (0.0252259036145 molecules / nm^3 typical for H2 gas double check)
n_atoms_half = 200
atom_mass = 1.0u"u"
atoms = [Atom(index = i, mass=atom_mass, σ=2.928u"Å", ϵ=0.074u"kcal* mol^-1") for i in 1:n_atoms_half]
max_coord = 200.0u"Å" # nm
coords = [max_coord .* rand(SVector{3}) for i in 1:n_atoms_half]

#Add bonded atoms
bond_length = 0.74u"Å" #hydrogen bond length
constraints = []
for j in range(1, n_atoms_half)
    push!(atoms, Atom(index = j + n_atoms_half, mass = atom_mass, σ=2.928u"Å", ϵ=0.074u"kcal* mol^-1"))
    push!(coords, coords[j] .+ SVector(bond_length,0.0u"Å",0.0u"Å"))
    push!(constraints, DistanceConstraint(SVector(j, j+n_atoms_half), bond_length))
end

constraint_algorithm = SHAKE(similar(coords))

velocities = [random_velocity(atom_mass, temp) for i in 1:length(atoms)]
# velocities = [[0.0u"nm/ps", 0.0u"nm/ps", 0.0u"nm/ps"] for i in 1:length(atoms)]
boundary = CubicBoundary(max_coord)

neighbor_finder = DistanceNeighborFinder(eligible = trues(length(atoms),length(atoms)), dist_cutoff = 1.5*r_cut)

sys = System(
        atoms = atoms,
        coords = coords,
        boundary = boundary,
        velocities=velocities,
        pairwise_inters=(
            LennardJones(
                cutoff = ShiftedPotentialCutoff(r_cut),
                use_neighbors = true,
                energy_units = u"kcal * mol^-1",
                force_units = u"kcal * mol^-1 * Å^-1"
                ),
            ),
        neighbor_finder = neighbor_finder,
        constraints = constraints,
        constraint_algorithm = constraint_algorithm,
        loggers=( 
            tot_eng = TotalEnergyLogger(1),
            coords_out = CoordinateLogger(1),
            vels_out = VelocityLogger(1),
            xyz = XYZLogger(1, "C:/Users/ejmei/Desktop/pos.xyz", false)
            ),
        energy_units = u"kcal * mol^-1",
        force_units = u"kcal * mol^-1 * Å^-1"
        )


# Equilibirate
simulator = VelocityVerlet(dt = 0.002u"ps")
# simulator = NoseHoover(dt=2.0u"fs", temperature=temp)
simulate!(sys, simulator, 100_000, n_threads = 1)
```
[pos.zip](https://github.com/JuliaMolSim/Molly.jl/files/11900657/pos.zip)

ejmeitz:
I've found the issue, our original implementation of SHAKE was only modifying the positions when it needed to be modifying the forces and it was doing so in the incorrect place anyways.

jgreener64:
I am having a look at this but it is a large and complex change and I am not so familiar with the algorithms. There are a couple of thoughts I have so far:

- The approach of modifying the accelerations rather than the coordinates/velocities may be limiting. I guess this is how LAMMPS does it but it probably means more maths to get it working for other simulators. Does this make it harder to support things like half-step velocities and user-defined simulators? We would want to support `Langevin` with this PR, for example. An approach that applies constraints to coordinates or velocities with two functions would allow those functions to be called at the end of the integration step, or at least in a more understandable way. It might be worth seeing what other software does, for example the OpenMM custom integrator interface which is compatible with constraints.

- The docs describe compatibility with SHAKE and RATTLE per simulator but I'm not sure this is the way it should be presented. As I understand it SHAKE is sufficient when velocities are not directly used, and if velocities are used then both SHAKE and RATTLE are required. So the user doesn't need to know which simulator works with which, but rather whether that simulator supports correct constraints or not. From my understanding all do here all apart from `Langevin` and `LangevinSplitting` (`MetropolisMonteCarlo` and `SteepestDescentMinimizer` wouldn't tend to use constraints).

- It would be nice to see the performance of constraints. A test of the implementation and of performance would be to constrain all bonds to hydrogen in a simulation of a protein similar to that in the docs.

Let me know what you think. There's a lot of good work in here but I want to make sure it's a flexible and accurate approach to constraints. We shouldn't require too much caution on behalf of the user when using constraints, which basically all biomolecular simulations will do.

ejmeitz:
- I believe you have to modify the forces so that the forces used in the velocity update also satisfy the constrained equations of motion. If you only update the positions the forces used to update velocities will be different from the forces used to update the positions (for F(t) and F(t+dt)). I will check the other codes to see what order they do things in, but based on the derivation I did awhile ago the Lagrange multiplier you derive modifies the force which indirectly modifies the positions. I'm not sure why this would limit other algorithms from having constraints. As long as the scheme integrates position SHAKE will work and as long as the scheme integrates velocitie RATTLE will work. So from what I understand Langevin should support SHAKE (pretty sure I coded that) but not RATTLE as the velocities in Langevin are not actually integrated rather interpolated/extrapolated.
-This is related to the last few sentences of the last answer. RATTLE only works on schemes that have a velocity-verlet like velocity integration step. At least that is my understanding. This was a question I never really found an exact answer to though. Like LAMMPS lets you use RATTLE/SHAKE on integrators the original papers would suggest should not work. If you know anyone with expertise here that could be helpful.
-Completely agree, I plan on doing a bunch of tests and finishing this after my qualifying exam next week :)

What kind of interface are you looking for with constraints? From the user POV they just choose SHAKE/RATTLE and define the distance constraints and the dispatch handles the rest.

Also I have derivations of SHAKE and RATTLE fully in LaTeX that do the steps skipped in most papers if you would like to see those.

jgreener64:
Okay I get it a bit more now. I think I am getting confused about why the choice of SHAKE or RATTLE is given to the user. Each integrator should either be correct with SHAKE (if the velocities are derived) or RATTLE (if they are integrated). Is that correct? So perhaps we could call the constraint method `SHAKERATTLE` or similar, and just apply it appropriately as you are in the integrators.

By interface I guess I mean docs about how to add constraints to new simulators. From your implementation I see the following two rules, does this sound correct?
- Apply constraints to the accelerations after they are computed each step.
- If the velocity update uses the stored accelerations from the last step, you also need to apply the velocity constraints after the velocity update.

That makes me think of one question, which is related to my mentioning of `Langevin` before: do the `Langevin` velocities need to be corrected to satisfy constraints after the noise is added? At the minute the coordinate update with noised velocities will lead to the coordinates violating constraints. This might be worth comparing to other software.

It would be useful to upload the derivations here if you have them, e.g. as PDF, it will be a useful reference.

> I plan on doing a bunch of tests and finishing this after my qualifying exam next week

Great, good luck.

ejmeitz:
I think it is totally reasonable to combine SHAKE & RATTLE into one user-facing class, but we have to be sure to say that if the integrator does not support rattle then velocity constraints will not be done.

Yep this all sounds correct. 
> * Apply constraints to the accelerations after they are computed each step.
> *If the velocity update uses the stored accelerations from the last step, you also need to apply the velocity constraints after the velocity update.

There is one extra caveat that I don't know how to handle yet. When you apply constraints they have to be the last thing that modifies the accelerations in that step. So as it stands constraints will not play nice with the `recalculate_forces` function. I was not super sure how to fix this since the `recalculate_forces` is currently at the end of the update.

I'll check out the Langevin thing, you should be able to apply the random perturbation and then calculate the constraint forces to bring the atoms back to the correct positions.

jgreener64:
> So as it stands constraints will not play nice with the `recalculate_forces` function.

I think if you add a `apply_position_constraints!` line as follows
```julia
        if recompute_forces
            accels_t = accelerations(sys, neighbors; n_threads=n_threads)
            # Add apply_position_constraints!
        else
            accels_t = accels_t_dt
        end
```
then this is okay. The forces already calculated this step have been constrained, used and discarded, and these recomputed forces are constrained ready for the next step.

> These velocities affect the position/force update of the next iteration and then the constraint forces are calculated from that.

But the noised velocities are used immediately after in the same step to update the coordinates. So those coordinates will not necessarily have constraints satisfied. I don't know how much of a problem this is.

Thanks for the derivations.

ejmeitz:
> So those coordinates will not necessarily have constraints satisfied. I don't know how much of a problem this is.

Hm yeah I'll have to look into the consequences of this. To be honest I'm not sure why the constraints are still satisfied the way its coded at the moment.

In LAMMPS Langevin is implemented as only a modification to the forces:
```
It only modifies forces to effect thermostatting. Thus you must use a separate time integration fix, like [fix nve](https://docs.lammps.org/fix_nve.html) to actually update the velocities and positions of atoms using the modified forces.
```
which would work fine with the constraint framework as you just calculate the constraint forces after the Langevin forces. I might re-write Langevin so that this is the case.

ejmeitz:
Todo list:

- [ ] Get constraints to play nice with `recompute_forces` as outlined above
- [ ] Fix Langevin constraints so they match OpenMM docs
- [ ] Thorough external testing, comparisons to LAMMPS and OpenMM
- [ ] Thorough internal test coverage
- [ ] Docs/Doc strings
- [ ] Change interface to combine `SHAKE` and `RATTLE` into one (still not sold on this)

jgreener64:
A few general comments:
- In `VelocityVerlet`/`Langevin`/`NoseHoover` it might be possible to keep the simulators clean by combining the three steps `reset_vel_correction!`, `apply_position_constraints!` and `addto_vel_correction!`. Since these always occur together they could be one function, e.g. a version of `apply_position_constraints!` with a flag to indicate that it should do the relevant storage.
- I think we should just keep the `constraints` tuple in `System`, rather than having a separate `constraint_algorithm` field. For instance `constraints` could be `(SHAKE_RATTLE(...),)` where the atom information is given to `SHAKE_RATTLE`. This means it is possible to use different constraint algorithms in the same simulation if desired. The presence of constraints can be checked by checking for the empty tuple or `nothing`.
- What is the status of the tests passing and the commented out tests?

ejmeitz:
- I'll look into implementing the first two points, shouldn't be too hard.

- I forgot to uncomment the tests, I manually ran all the simulators locally for hydrogen and they work well. The SHAKE code is identical to before so I'd be really surprised if that breaks. I think the only extra test that should be written is an addition to the poly-atomic test to also check `RATTLE` constraints. 
- And of course `NoseHoover` is broken now that we constrain positions instead of forces. I've tried a couple permutations of how the constraints are applied. The OpenMM integrator is hard to learn much from since it is the more advanced form and also integrates at half steps whereas I originally coded `NoseHoover` to operate like `VelocityVerlet`.

- There is a protein test but I just have it constraint hydrogen bonds right now, I do not think it is super relevant to testing the code and will probably remove it until we have more complicated constraints.

ejmeitz:
- For the first point I think I'd have to dispatch on which simulator is being used, so ultimately will be more code generated but the `simulate!` function would stay cleaner
- Working on this, its mostly done locally just need to test that I didn't break things.
- Thoughts on not supporting `NoseHoover` for this first pass? That integrator should realistically be updated to `NoseHooverChain` at some point anyway.

jgreener64:
> Is there a reason to not use a class to represent no constraints vs. `nothing`

I guess it could be a type, but I see it a bit like the interactions where the default is `()` and whatever tuple is given is iterated over. So maybe `()` is a better default than `nothing`. If it is a type then we would have to allow a type or tuple of types for multiple sets of constraints, which might get confusing.

> Thoughts on not supporting `NoseHoover` for this first pass?

Fine by me.

ejmeitz:
I have no clue how to fix these merge conflicts. It won't show me what's conflicting.

ejmeitz:
The tests break somewhere in ReplicaSystem. I probably messed something up fixing the merge conflicts (there were so many). I did not immediately see what was wrong, I also don't know what this data stuff is really supposed to do so could've messed something up there.

ejmeitz:
Tests passed locally, just to summarize this PR.

New Features:
- Rattle implemented for VelocityVerlet
- Velocity corrections added to all integrators that use SHAKE
- Unified RATTLE/SHAKE interface into one class
- Setup constraint framework so that it will be easy to add more complicated algorithms
- A function to calculate the # of DoF lost by using SHAKE, this correction is applied when calculating tempearture


Things to double check before merging:
- Energy conservation with constraints, maybe do a parity check with OpenMM for Langevin
- Check DoF function, I am not sure it is generic

What I was unable to achieve:
- Constraints for NoseHoover, Overdamped Langevin, or LangevinSplitting. I think each of these should be possible to implement they just are not now.
- Angle constraints (should be a relatively easy new PR)
- Faster version of SHAKE that can be parallelized (also left to another PR)

jgreener64:
Nice. I'll try and review this properly over the next few days.

From an initial pass, could you remove all commented-out code and anything that is just setting up for later.

Also the field in `System` can remain as `constraints`, and I don't think the eligible matrix should be updated automatically for constraints (we leave it to the user to setup elsewhere for clarity).

jgreener64:
For the standard molecular case pairwise interactions are excluded for constrained atoms, but there are more general simulation cases where people may want other behaviour.

ejmeitz:
Can I add a flag for this? I think by default we should do this, it is what I would expect the code to do and also a far more common use case imo. (then again Ive never simulated macromolecules)

src/types.jl
```diff
@@ -539,6 +541,18 @@ function System(;
         throw(ArgumentError("there are $(length(atoms)) atoms but $(length(atoms_data)) atom data entries"))
     end
 
+    df = n_dof(D, length(atoms), boundary)
+    if length(constraints) > 0
+        if neighbor_finder == NoNeighborFinder()
+            throw(ArgumentError("Constraints algorithms require neighbor lists."))
```
jgreener64, line 547: Allow this, since the user may not be providing pairwise interactions at all.

src/types.jl
```diff
@@ -734,10 +750,8 @@ construction where `n` is the number of threads to be used per replica.
     value is passed to the argument `replica_general_inters`.
 - `replica_general_inters=[() for _ in 1:n_replicas]`: the general interactions for 
     each replica.
-- `constraints::CN=()`: the constraints for bonds and angles in the system (to be used if the same 
-    for all replicas). Typically a `Tuple`.
-- `replica_constraints=[() for _ in 1:n_replicas]`: the constraints for bonds and angles in each
```
jgreener64, line 739: I think you can allow `replica_constraints` by calculating `df` for each system. There may be situations where people want different constraints for each replica.

src/units.jl
```diff
@@ -143,7 +144,8 @@ function validate_coords(coords)
     end
 
     valid_length_dimensions = [u"𝐋", NoDims]
-    coord_dimension = dimension(zero(eltype(coords))[1])
+    println(eltype(coords[1]))
```
jgreener64, line 147: Remove.

test/simulation.jl
```diff
@@ -460,130 +460,135 @@ end
...
+    
 
-    lengths = map(eachindex(sh.is)) do r
-        return norm(vector(sys.coords[sh.is[r]], sys.coords[sh.js[r]], sys.boundary))
-    end
+# end
 
-    @test maximum(abs.(lengths .- 0.1u"nm")) < 1e-10u"nm"
-end
+# @testset "SHAKE triatomic" begin
```
jgreener64, line 529: Does this test work?

src/simulators.jl
```diff
@@ -202,14 +217,27 @@ function simulate!(sys,
     neighbors = find_neighbors(sys, sys.neighbor_finder; n_threads=n_threads)
     run_loggers!(sys, neighbors, 0, run_loggers; n_threads=n_threads)
 
+    using_constraints = (length(sys.constraints) > 0)
```
jgreener64, line 220: This can be used in all simulators for clarity rather than always checking `length(sys.constraints) > 0`.

src/simulators.jl
```diff
@@ -202,14 +217,27 @@ function simulate!(sys,
...
+        
+        if length(sys.constraints) > 0
+            ca_coord_storage .= sys.coords
+        end
 
-        old_coords = copy(sys.coords)
         sys.coords += sys.velocities .* sim.dt
-        apply_constraints!(sys, old_coords, sim.dt)
+
+        sys = apply_position_constraints!(sys, ca_coord_storage, n_threads=n_threads)
```
jgreener64, line 235: In general you don't need to set `sys = ` here, even though `apply_position_constraints!` returns `sys`.

jgreener64 (commented):
This is starting to take shape, well done. I left lots of comments. You will also need to merge/rebase recent changes.

I think for now we should not change the neighbour list automatically. For biomolecular simulation it should be done during system setup from a file, as it is for bonds currently. That can be a future change. The user would specify which type of constraints they want, e.g. all bonds to H, then these would be added and the bonds removed.

src/constraints/constraints.jl
```diff
@@ -0,0 +1,247 @@
...
+    
+    for ca in sys.constraints
+        velocity_constraints!(sys, ca, n_threads = n_threads)
+    end
+
+    return sys
+end
+
+
+function check_position_constraints(sys::System, ca::ConstraintAlgorithm)
```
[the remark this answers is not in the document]
ejmeitz: max_err is just the error I calculate, the actual tolerance is already in the constraint algorithm (`ca.tolerance`). Since the constraints are iterative you always should be able to reach that tol.

src/simulators.jl
```diff
@@ -331,19 +366,29 @@ function simulate!(sys,
...
 
+    ca_coord_storage = similar(sys.coords)
+    ca_vel_storage = similar(sys.velocities)
+
     for step_n in 1:n_steps
-        accels_t = accelerations(sys, neighbors; n_threads=n_threads)
+        accels_t = accelerations(sys, neighbors; n_threads=n_threads)    
 
         sys.velocities += accels_t .* sim.dt
+        sys = apply_velocity_constraints!(sys, n_threads=n_threads)
```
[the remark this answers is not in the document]
ejmeitz: I was just copying OpenMM there: https://github.com/openmm/openmm/blob/4e7425295b635f946218f7aa5aaa30a29efcef62/platforms/reference/src/SimTKReference/ReferenceLangevinMiddleDynamics.cpp#L92

src/simulators.jl
```diff
@@ -202,14 +217,27 @@ function simulate!(sys,
...
+            ca_coord_storage .= sys.coords
+        end
 
-        old_coords = copy(sys.coords)
         sys.coords += sys.velocities .* sim.dt
-        apply_constraints!(sys, old_coords, sim.dt)
+
+        sys = apply_position_constraints!(sys, ca_coord_storage, n_threads=n_threads)
+
+        if using_constraints
```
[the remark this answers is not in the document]
ejmeitz: What are you referring to as this here? The placement of the constraint or its existence? I also double checked this one with OpenMM https://github.com/openmm/openmm/blob/71f4b3fcb0b07d0fe41302df03f5373c8424cefa/platforms/reference/src/SimTKReference/ReferenceVerletDynamics.cpp

test/simulation.jl
ejmeitz: Forgot to run it tbh. Will uncomment.

src/types.jl
ejmeitz: You're referring to the Neighbor finder check?

src/constraints/constraints.jl
```diff
@@ -0,0 +1,247 @@
...
+                    push!(connected_constraints,
+                        DistanceConstraint(SVector(ai, neigh_idx), idx_dist_pairs[ai,neigh_idx]))
+                end
+            end
+            connected_constraints = convert(SVector{length(connected_constraints)}, connected_constraints)
+            push!(clusters, ConstraintCluster(connected_constraints))
+        end
+    end
+
+    return clusters
```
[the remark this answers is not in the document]
jgreener64: https://docs.julialang.org/en/v1/manual/performance-tips/#man-performance-abstract-container

src/constraints/constraints.jl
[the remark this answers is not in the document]
jgreener64: :+1:

src/simulators.jl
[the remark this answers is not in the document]
jgreener64: Okay makes sense.

src/simulators.jl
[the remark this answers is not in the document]
jgreener64: Okay makes sense.

src/simulators.jl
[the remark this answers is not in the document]
jgreener64: I think the
```julia
        if using_constraints
            ca_coord_storage .= sys.coords
        end
```
line might need to go above the first coordinate update though?

ejmeitz:
- Note to self, add test for Overdamped Langevin

src/constraints/constraints.jl
```diff
@@ -0,0 +1,232 @@
+export
+    DistanceConstraint,
+    check_position_constraints,
+    check_velocity_constraints,
+    disable_intra_constraint_interactions!,
+    apply_position_constraints!,
+    apply_velocity_constraints!    
+
+abstract type ConstraintAlgorithm end;
```
jgreener64, line 9: Remove this abstract type, unless it is specifically required somewhere.

src/constraints/constraints.jl
```diff
@@ -0,0 +1,232 @@
+export
+    DistanceConstraint,
+    check_position_constraints,
```
jgreener64, line 3: Add docstrings to the exported functions.

src/constraints/shake.jl
```diff
@@ -0,0 +1,141 @@
...
+# Arguments
+- constraints::AbstractVector{<:Constraint}: A vector of constraints to be imposed on the system.
+- n_atoms::Integer: The number of atoms in the system.
+- dist_tolerance: Tolerance used to end iterative procedure when calculating constraints.
+    Should have same units as coords.
+- vel_tolerance: Tolerance used to end iterative procedure when calculating velocity constraints.
+    Should have same units as velocities*coords.
+"""
+struct SHAKE_RATTLE{CC, D, V} <: ConstraintAlgorithm
+    clusters::AbstractVector{CC}
```
jgreener64, line 21: `::CC` for even more concreteness.

jgreener64 (commented):
Okay this looks good, just some minor changes before merge.

src/constraints/constraints.jl
ejmeitz: It's used by the `check_position_constraint` (and velo) functions. It's not explicitly needed right now since we only have one algorithm but I don't think its hurting anything?

src/constraints/constraints.jl
ejmeitz: It's used by the `check_position_constraint` (and velo) functions. It's not explicitly needed right now since we only have one algorithm but I don't think its hurting anything?

jgreener64: If it's used for type annotation of function arguments then it forces anything calling that function to be a sub-type. If those type annotations are removed (as I am suggesting) then it doesn't seem to be used for anything.

jgreener64:
Nice one, thanks :tada:

PR #136, opened 2023-06-27
Add anisotropic MC barostat

bondrewd:
This PR adds an implementation for the anisotropic Monte Carlo barostat following OpenMM's implementation.

The new interface for using this barostat in 3D and 2D systems is:

```julia
MonteCarloAnisotropicBarostat(
    1.0u"bar", # Pressure in the X-axis
    1.0u"bar", # Pressure in the Y-axis
    1.0u"bar", # Pressure in the Z-axis
    temperature,
    boundary
)
```
and
```julia
MonteCarloAnisotropicBarostat(
    1.0u"bar*nm", # Pressure in the X-axis
    1.0u"bar*nm", # Pressure in the Y-axis
    temperature,
    boundary
)

```
respectively.

If you set any pressure value as 'nothing', the barostat is uncoupled in that axis. For example, for coupling only the Y-axis, you could do the following:

```julia
MonteCarloAnisotropicBarostat(
    nothing, # Keep the box in the X-axis fixed
    1.0u"bar", # Scale box in the Y-axis
    nothing, # Keep the box in the Z-axis fixed
    temperature,
    boundary
)
```

After discussing and merging this PR, I will start working on the constant surface tension MC barostat for membrane simulations.

### DIFF
--- a/src/coupling.jl
+++ b/src/coupling.jl
@@ -242,7 +242,7 @@ function apply_coupling!(sys::System{D, G, T}, barostat::MonteCarloBarostat, sim
 end
 
 @doc raw"""
-    MonteCarloBarostat(pressure_X, pressure_Y, pressure_Z, temperature, boundary;
+    MonteCarloAnisotropicBarostat(pressure_X, pressure_Y, pressure_Z, temperature, boundary;
                        n_steps=30, n_iterations=1, scale_factor=0.01, scale_increment=1.1,
                        max_volume_frac=0.3, trial_find_neighbors=false)
 
@@ -304,7 +304,7 @@ function MonteCarloAnisotropicBarostat(
     max_volume_frac=0.3,
     trial_find_neighbors=false,
 )
-    pressure = [pressure_X, pressure_Y, pressure_Z]
+    pressure = SVector(pressure_X, pressure_Y, pressure_Z)
     volume_scale = box_volume(boundary) * float_type(boundary)(scale_factor)
     volume_scale = fill(volume_scale, 3)
 
--- a/src/coupling.jl
+++ b/src/coupling.jl
@@ -242,8 +242,8 @@ function apply_coupling!(sys::System{D, G, T}, barostat::MonteCarloBarostat, sim
 end
 
 @doc raw"""
-    MonteCarloAnisotropicBarostat(pressure_X, pressure_Y, pressure_Z, temperature, boundary;
-                       n_steps=30, n_iterations=1, scale_factor=0.01, scale_increment=1.1,
+    MonteCarloAnisotropicBarostat(pressure, temperature, boundary; n_steps=30,
+                       n_iterations=1, scale_factor=0.01, scale_increment=1.1,
                        max_volume_frac=0.3, trial_find_neighbors=false)
 
 The Monte Carlo anisotropic barostat for controlling pressure.
@@ -273,13 +273,15 @@ It should be used alongside a temperature coupling method such as the [`Langevin
 simulator or [`AndersenThermostat`](@ref) coupling.
 The neighbor list is not updated when making trial moves or after accepted moves.
 Note that the barostat can change the bounding box of the system.
+`pressure` is a SVector of lenght 3 with components pressX, pressY, and pressZ
+representing the target pressure in each axis.
 To keep an axis fixed, set the corresponding pressure to `nothing`.
-For rectangular boundaries `pressure_Z` is not required.
+For rectangular boundaries `pressZ` is not required.
 
 Not currently compatible with automatic differentiation using Zygote.
 """
-mutable struct MonteCarloAnisotropicBarostat{T, P, K, V}
-    pressure::P
+mutable struct MonteCarloAnisotropicBarostat{D, T, P, K, V}
+  pressure::SVector{D,P}
     temperature::K
     n_steps::Int
     n_iterations::Int
@@ -292,9 +294,7 @@ mutable struct MonteCarloAnisotropicBarostat{T, P, K, V}
 end
 
 function MonteCarloAnisotropicBarostat(
-    pressure_X,
-    pressure_Y,
-    pressure_Z,
+    pressure::SVector{D},
     temperature,
     boundary;
     n_steps=30,
@@ -303,10 +303,10 @@ function MonteCarloAnisotropicBarostat(
     scale_increment=1.1,
     max_volume_frac=0.3,
     trial_find_neighbors=false,
-)
-    pressure = SVector(pressure_X, pressure_Y, pressure_Z)
+) where {D}
     volume_scale = box_volume(boundary) * float_type(boundary)(scale_factor)
-    volume_scale = fill(volume_scale, 3)
+    volume_scale = fill(volume_scale, D)
+    length(boundary.side_lengths) != D && error("pressure is not compatible with boundary")
 
     return MonteCarloAnisotropicBarostat(
         pressure,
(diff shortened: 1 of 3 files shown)


jgreener64:
Nice one! Something to consider is constructing it with a `SVector` instead, e.g.
```julia
MonteCarloAnisotropicBarostat(
    SVector(1.0u"bar", 1.0u"bar", 1.0u"bar"),
    temperature,
    boundary
)
```
This gives a little more consistency between the 3D/2D calls and may reduce code reuse internally by allowing one constructor to be defined.

bondrewd:
Thank you for the comment! I have refactored the constructor, using an `SVector` for the pressure. It helps to simplify the interface since the dimension of the pressure vector becomes available.

I also added some tests following the current examples. They are designed to check for regressions. In the future, implementing more rigorous tests might be of interest.

jgreener64:
The implementation seems good from a quick look.

As you say there is the question of how to test this since we don't currently allow logging pressure in one axis. It might be nice to test with a pressure other than `SVector(1.0u"bar", 1.0u"bar", 1.0u"bar")` though. Maybe you could test with `SVector(1.5u"bar", 0.5u"bar", 1.0u"bar")` or similar, and update the ranges for testing as appropriate. Perhaps the box size in each dimension is available to test and shows a difference in this scheme too?

bondrewd:
Thank you for the suggestion! I added more cases for testing the pressure combination you suggested and different couplings: only X-axis, only Y-axis, only Z-axis, and uncoupled.

test/simulation.jl
```diff
@@ -942,6 +942,83 @@ end
...
+    VolumeLogger(n_steps) = GeneralObservableLogger(box_volume_wrapper, typeof(1.0u"nm^3"), n_steps)
+    
+    baro_f(pressure) = MonteCarloAnisotropicBarostat(pressure, temp, boundary)
+    lang_f(barostat) = Langevin(dt=dt, temperature=temp, friction=friction, coupling=barostat)
+    vand_f(barostat) = VelocityVerlet(dt=dt, coupling=(AndersenThermostat(temp, 1.0u"ps"), barostat))
+    
+    pressure_test_set = (
+        SVector(1.0u"bar", 1.0u"bar", 1.0u"bar"), # XYZ-axes coupled with the same pressure value
+        SVector(1.5u"bar", 0.5u"bar", 1.0u"bar"), # XYZ-axes coupled with different pressure values
+        SVector(1.0u"bar", nothing, nothing), # Only X-axis coupled
```
jgreener64, line 969: Just test one of the x/y/z single couplings, for speed.

src/spatial.jl
```diff
@@ -759,7 +759,9 @@ Not currently compatible with automatic differentiation using Zygote.
 function scale_coords!(sys, scale_factor; ignore_molecules=false)
     if ignore_molecules || isnothing(sys.topology)
         sys.boundary = scale_boundary(sys.boundary, scale_factor)
-        sys.coords = sys.coords .* scale_factor
+        for i in eachindex(sys.coords)
```
jgreener64, line 762: This is slow on the GPU, unfortunately. You might have to write a function that does `coord .* scale_factor` and broadcast that function (like `ustrip_vec`, a double broadcast).

Also, does the branch with molecules below work when `scale_factor` is a `SVector`?

src/coupling.jl
```diff
@@ -239,3 +240,157 @@ function apply_coupling!(sys::System{D, G, T}, barostat::MonteCarloBarostat, sim
...
+If the topology of the system is set then molecules are moved as a unit so properties
+such as bond lengths do not change.
+
+The barostat assumes that the simulation is being run at a constant temperature but
+does not actively control the temperature.
+It should be used alongside a temperature coupling method such as the [`Langevin`](@ref)
+simulator or [`AndersenThermostat`](@ref) coupling.
+The neighbor list is not updated when making trial moves or after accepted moves.
+Note that the barostat can change the bounding box of the system.
+For 3D systems, `pressure` is a SVector of lenght 3 with components pressX, pressY,
```
jgreener64, line 276: Typo of length x 2.

src/coupling.jl
```diff
@@ -239,3 +240,157 @@ function apply_coupling!(sys::System{D, G, T}, barostat::MonteCarloBarostat, sim
...
+    n_steps=30,
+    n_iterations=1,
+    scale_factor=0.01,
+    scale_increment=1.1,
+    max_volume_frac=0.3,
+    trial_find_neighbors=false,
+) where {D}
+    volume_scale = box_volume(boundary) * float_type(boundary)(scale_factor)
+    volume_scale = fill(volume_scale, D)
+    length(boundary.side_lengths) != D && error("pressure is not compatible with boundary")
```
jgreener64, line 309: Use `ArgumentError`, and print in the error message what the dimensions of the pressure and the boundary are.

src/coupling.jl
```diff
@@ -239,3 +240,157 @@ function apply_coupling!(sys::System{D, G, T}, barostat::MonteCarloBarostat, sim
...
+) where {D, G, T}
+
+    !iszero(step_n % barostat.n_steps) && return false
+    all(isnothing, barostat.pressure) && return false
+
+    kT = sys.k * barostat.temperature
+    n_molecules = isnothing(sys.topology) ? length(sys) : length(sys.topology.molecule_atom_counts)
+    recompute_forces = false
+
+    axis = undef
```
jgreener64, line 341: `axis = 0` is the same type.

src/coupling.jl
```diff
@@ -239,3 +240,157 @@ function apply_coupling!(sys::System{D, G, T}, barostat::MonteCarloBarostat, sim
...
+    temperature,
+    boundary;
+    n_steps=30,
+    n_iterations=1,
+    scale_factor=0.01,
+    scale_increment=1.1,
+    max_volume_frac=0.3,
+    trial_find_neighbors=false,
+) where {D}
+    volume_scale = box_volume(boundary) * float_type(boundary)(scale_factor)
```
jgreener64, line 307: Rename this to avoid the change of type on the next line.

src/coupling.jl
```diff
@@ -239,3 +240,157 @@ function apply_coupling!(sys::System{D, G, T}, barostat::MonteCarloBarostat, sim
...
+    while true
+        axis = rand(1:D)
+        !isnothing(barostat.pressure[axis]) && break
+    end
+    mask1 = falses(D)
+    mask2 = trues(D)
+    mask1[axis] = true
+    mask2[axis] = false
+
+    for attempt_n in 1:barostat.n_iterations
```
jgreener64, line 351: The `n_iterations` parameter is not in the OpenMM implementation but I added it in case people want to run multiple trials per step. Given that these trials should be independent, I think the above lines picking the dimension should go inside the loop.

jgreener64 (commented):
Left a few comments before merge. I think the only one that requires real thought is `scale_coords!` with the `SVector`.

test/simulation.jl
bondrewd: I removed `only x-axis` and `only z-axis`.

src/spatial.jl
bondrewd: I added a new `scale_vec` function:

```julia
"""
    scale_vec(v, c)
Broadcasted form of `.* c`, allowing scaling all the vectors in v by c, where c is
either a constant or a vector.
"""
scale(v, c) = v .* c
scale_vec(v, c) = scale.(v, Ref(c))
```

and used it in `scale_coords!`.

I also refactored the branch used when a topology is present:

```julia
# Move all atoms in a molecule by the same amount according to the molecule center
# Then move the atoms back to the molecule center and wrap in the scaled boundary
shift_vecs = scale_vec(mol_centers, scale_factor .- 1)
```

src/coupling.jl
bondrewd: Fixed.

src/coupling.jl
bondrewd: Changed.

```julia
length(boundary.side_lengths) != D && throw(ArgumentError("pressure vector length ($(D)) " *
    "must match boundary dimensionality ($(length(boundary.side_lengths)))"))
```

src/coupling.jl
bondrewd: Changed to `axis = 0`.

src/coupling.jl
bondrewd: I moved the axis selection part inside the trial loop:

```julia
for attempt_n in 1:barostat.n_iterations
        axis = 0
        while true
            axis = rand(1:D)
            !isnothing(barostat.pressure[axis]) && break
        end
        mask1 = falses(D)
        mask2 = trues(D)
        mask1[axis] = true
        mask2[axis] = false
```

src/coupling.jl
bondrewd: I renamed the first `volume scale`:

```julia
volume_scale_factor = box_volume(boundary) * float_type(boundary)(scale_factor)
volume_scale = fill(volume_scale_factor, D)
```

jgreener64:
Great, thanks.

PR #138, opened 2023-07-04
fix bug in MC anisotropic barostat

bondrewd:
The length scaling factor in the anisotropic barostat is currently:

```julia
l_scale = SVector{D}(mask1 * (D == 2 ? sqrt(v_scale) : cbrt(v_scale)) + mask2)
```

However, since in the anisotropic case, only one axis is scaled at a time, it should be replaced to:

```julia
l_scale = SVector{D}(mask1 * v_scale + mask2)
```


jgreener64:
Looks good, just rebase/merge recent changes I made to shorten the tests and I will merge this.

PR #139, opened 2023-07-04
implement MC membrane barostat

bondrewd:
This PR adds an implementation for the Monte Carlo membrane barostat following OpenMM's implementation.

This barostat is only available for 3D systems, and the API is as follows:

```julia
MonteCarloMembraneBarostat(
    1.0u"bar",       # Pressure in the system
    1.0u"bar*nm" # Surface tension
    temperature,
    boundary;
    xy_isotropy=false,              # If true, scale the XY-axes isotropically
    constant_volume=false,    # If true, keep the volume constant by scaling the Z-axis accordingly
    z_axis_fixed=false             # If true, keep the Z-axis fixed
)
and
```

In this barostat, `constant_volume` and `z_axis_fixed` cannot be true simultaneously.

### DIFF
--- a/src/coupling.jl
+++ b/src/coupling.jl
@@ -7,7 +7,8 @@ export
     RescaleThermostat,
     BerendsenThermostat,
     MonteCarloBarostat,
-    MonteCarloAnisotropicBarostat
+    MonteCarloAnisotropicBarostat,
+    MonteCarloMembraneBarostat
 
 """
     apply_coupling!(system, coupling, simulator, neighbors=nothing,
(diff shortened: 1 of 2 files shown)


src/coupling.jl
```diff
@@ -393,3 +394,189 @@ function apply_coupling!(sys::System{D, G, T},
     end
     return recompute_forces
 end
+
+@doc raw"""
+    MonteCarloMembraneBarostat(pressure, temperature, boundary; n_steps=30,
```
jgreener64, line 399: Update the arguments here, also briefly describe the arguments and the constraint on the two arguments in the docstring.

jgreener64 (commented):
This looks really nice, just one comment before merge.

By the way if you have any cool membrane examples they would be a nice addition to the examples section of the docs.

src/coupling.jl
bondrewd: Done.

jgreener64:
There seem to be occasional test failures with this barostat for the `constant_volume=true` case. See for example https://github.com/JuliaMolSim/Molly.jl/actions/runs/5466737179/jobs/9952101066.

Any chance you could take a look?

bondrewd:
Sure! Today I tested the MC membrane barostat and found that for some seeds, the 25-atoms-system adopts conformations in which one axis is favored.

I tried increasing the number of atoms, e.g., 100, and found the same behavior:
movie

Overall, it seems that an LJ gas might not be the most suitable system for testing the MC membrane barostat. The reason is that the surface tension term `\gamma * dA` favors the scaling in the XY plane (for a positive surface tension). For membrane systems, there exists a symmetry in the X- and Y-axes that balance the scaling. For a gas that is fully symmetric (point masses with radial LJ potential), there is no counterbalance for the surface tension.

One possible solution is to momentarily comment out the test until the MARTINI PR is merged. Then we could, for example, use a small and equilibrated membrane system composed of, let's say, 128 lipids, and `@test` some properties such as the `pressure`, the `volume`, the `area per lipid`, and the `membrane thickness`. Since coarse-grained is relatively cheap to compute, it might not have a huge impact on the CI execution time. @jgreener64, what do think about this idea?

The other issue that needs to be addressed is the implementation of correctness tests since what I have described above only covers regressions. One possible way is to sample with the barostat and test the fluctuations in volume and pressure. This might be a bit off-topic, but testing MD code is not so trivial and we might want to study how other more mature software are doing it.

jgreener64:
Thanks for looking into it! Yes in the short term I will just remove the tests in `if barostat.constant_volume` and long term we can use MARTINI to test pressure more thoroughly.

I agree that correctness testing is difficult, especially when stochasticity is involved. I have consistently run into this when implementing algorithms for the package. In general it would be useful to study other software and see how they test correctness, or at least try to match them exactly for more algorithms (currently we match OpenMM exactly for a short NVE protein simulation). Consistent random numbers across languages is one issue there.

PR #140, opened 2023-07-04
Add Unit Checks and Restrictions

ejmeitz:
Currently, the `random_velocity` function assumes the system masses are passed as `u` but if you use `g/mol` then this function will break whatever simulation you try to run. I have not actually fixed the code here it's just something I would like to fix at some point.

I'm not really sure the best way to handle this. On one hand it would be nice if all things units related were just automated but that's really hard. 

A quick fix would be to remove the default value if it is not generically valid, but also maybe we should not let people mix molar and non-molar quantities in the simulation. Its a recipe for things to break in Unitful.


ejmeitz:
Ok I made some headway on how to handle this issue by dispatching on the units passed to the function. I was able to fix this for almost all of the functions. The only one's remaining are the 4 which take a `System `as the parameter and use sys.k as the Boltzmann constant.

In the functions I added I used multiple dispatch to prevent the user from mixing units they shouldn't. For example, `random_velocity` now has 3 dispatch versions. One where the mass is `Unitful.Mass` another where the mass is `MolarMass ` and (new dimension I defined) and a third where the mass is just a float. The two functions with units do not let the user set the Boltzmann constant and the third does not provide a default value for `k`. I think this avoids most units issues at compile time, but I could not see a similar way to do this with the `random_velocities` functions that take the `System` object. 

In the current implementation if the user calls:
`random_velocities(sys, 10.0)` with the system masses having units
there is no dispatch for this since the user constructed the `System` with units but passed a unitless temperature. Furthermore, if the user constructed the system with molar masses the Sys.k value would not be per-mol and would also break. I don't see a super clean way to fix this. I would ideally like to use dispatch like I did in the last commit I made.

jgreener64:
So as mentioned in https://github.com/JuliaMolSim/Molly.jl/pull/144 it might just make sense to switch from `u` to `g/mol` throughout the entire package and docs. Will that let us remove all the `accel_remove_mol` and related functions, and also make this case with velocities easier to deal with? Verbose unit checking and conversion code is probably an indication that the design can be improved.

ejmeitz:
Yeah if we use `g/mol` with `kJ/mol` we shouldn't need the `accel_remove_mol` functions. In general as long as the energy, force and mass are uniformly molar or non-molar we should be fine. I think the velocities issue is as fixed as it can be. I removed all unit assumptions in the `random_velocity` functions so now the user has to pass the Boltzmann constant if they are building a velocity vector before the System gets constructed. When `random_velocities!` gets called it will use the System Boltzmann constant which I now changed to be per mol if the other units were passed as per mol.

I think most of the coding is done I just have to search through the codebase now to find places I broke or where units were used.

ejmeitz:
@jgreener64 you can probably review this now. The features I want to add are added...ended up being a lot bigger than I intended. Basically, units are enforced to be as consistent as possible. Mixing molar and non-molar mass, force, or energy is completely banned and length units are also forced to be consistent. Time and other units are pretty free to be whatever and should be handled by Unitful alright.

I was able to remove all of the `accel_remove_mol` and `energy_add_mol` but I still have a few `energy_add_mol` in the Barostats because pressure should be intensive.

Some things I'm not the biggest fan of still are the default values for Atom assume units and I'd rather just remove those default values. For sigma and epsilon that would cause all sorts of issues but for mass I don't really see why that would break things. Also the `random_velocity` functions gives velocity back in sqrt(energy/mass) which is velocity but it looks kinda ugly when printed. I don't think it breaks anything just something I couldn't fix.

jgreener64:
This is a big change but I'll try and review it in the next week. Do the commented out tests pass?

src/spatial.jl
```diff
@@ -194,6 +194,9 @@ Unitful.ustrip(u::Unitful.Units, b::CubicBoundary) = CubicBoundary(ustrip.(u, b.
 Unitful.ustrip(b::RectangularBoundary) = RectangularBoundary(ustrip.(b.side_lengths))
 Unitful.ustrip(u::Unitful.Units, b::RectangularBoundary) = RectangularBoundary(ustrip.(u, b.side_lengths))
 
+Unitful.unit(b::Union{CubicBoundary,RectangularBoundary}) = unit.(b.side_lengths)
```
jgreener64, line 197: Are these required for the PR? I'm not sure `unit` should return multiple values, it errors on vectors for example.

src/units.jl
```diff
@@ -0,0 +1,261 @@
...
+end
+
+function check_other_units(atoms, boundary, constraints, sys_units::NamedTuple)
+    box_units = unit(boundary)
+
+    if !all(sys_units[:length] .== box_units)
+        throw(ArgumentError("Simulation box constructed with $(box_units) but length unit on coords was $(sys_units[:length])"))
+    end
+
+    σ_units = unit.(getproperty.(atoms, :σ))
```
jgreener64, line 96: These are not required properties for atoms, so this could be inside a `hasproperty` check.

jgreener64 (commented):
I think overall this looks good, the important concept being to define the mass as g/mol throughout and hence avoid the need for all the adding/removing of molar dimensions.

I am less sure about the changes to the Boltzmann constant. In particular many of the tests now have to define k where they didn't before. Is there a way to avoid this? Maybe a default value for no units. Do you still think there is something wrong with the changes to k?

Maybe if @noeblassel has a moment he could look over the changes to k - in particular `convert_k_units` in src/units.jl - and see if it makes sense.

There is quite a lot of unit checking in here which I was hoping to leave to unitful, i.e. the errors would be informative enough to use directly. Maybe it is more useful to do all these checks at the start though.

Otherwise I think you can proceed to fix the units test and add a small number of error tests for incorrect units. I can fix the AD tests after merge.

src/spatial.jl
ejmeitz: Not really I just added it for convenience.

ejmeitz:
The Boltzmann constant checks are also what I am least happy with. Its a bit convoluted and like you said was rather painful to incorporate into existing tests. The issue with kB is that it can be molar or non-molar so the issues we were having with mass can also appear because of kB. I do not really have a better idea of how to handle this, will have to think a bit more.

As for all the extra checks I added, I found that sometimes the Unitful errors are not that useful for figuring out what in your input script broke things. For example, my boundary had the wrong units and that bug manifested in the velocity function.

src/units.jl
ejmeitz: hasproperty will always be true because there are default values though. I do agree if they are unused they should not be checked, but the real fix for that is just not having them in the Atom constructor. I'll just check if they match the defaults for now to determine whether or not I check the units.

ejmeitz:
Ok I think I managed to fix almost everything:
- `convert_k_units` ignores the value for `k` provided in System constructor unless `energy_units` is `NoUnits`
- Fixed old Units test to get it to pass

Things that are still a little weird
- `random_velocity(10.0u"g/mol", 10u"K")` gives velocities in `J^1/2 g^-1/2` which Unitful handles totally fine but any logger or property that is calculated from velocities might have funky units. I could not figure out how to get this to auto-reduce
- Tests that don't pass: Zygote, the OpenMM tests in protien.jl that compare the no-units version to the units version (test 32)

I will push a test that tests invalid unit combos soon.

src/units.jl
jgreener64: Simulations don't have to be run with the `Atom` type, they can be run with any type. Other types may not have those properties.

jgreener64:
Could you fix the issue with the OpenMM test? Then this more or less looks okay.

ejmeitz:
OpenMM tests pass.

Had to make some assumptions about the Boltzmann constant in the System constructors in setup.jl. My current implementation assumes that with NoUnits the boltzman constant has molar units (opposite of default in other System construtors). Not sure what units these files can have but the current implementation looks like it expects kJ * mol.

test/zygote.jl
```diff
@@ -67,6 +68,7 @@ end
             coulomb_const=f32 ? Float32(ustrip(Molly.coulombconst)) : ustrip(Molly.coulombconst),
             force_units=NoUnits,
             energy_units=NoUnits,
+            k = k
```
jgreener64, line 71: Remove this line, I think this is what's causing the gradient test to fail.

src/types.jl
```diff
@@ -462,15 +445,15 @@ interface described there.
...
-    simulations.
+- `k::K=Unitful.k or Unitful.k*Unitful.Na`: the Boltzmann constant, which may be modified in some
+    simulations. k is chosen based on `energy_units` passed.
 - `force_units::F=u"kJ * mol^-1 * nm^-1"`: the units of force of the system.
     Should be set to `NoUnits` if units are not being used.
 - `energy_units::E=u"kJ * mol^-1"`: the units of energy of the system. Should
     be set to `NoUnits` if units are not being used.
 """
 mutable struct System{D, G, T, A, C, B, V, AD, TO, PI, SI, GI, CN, NF,
-                      L, K, F, E, M} <: AbstractSystem{D}
```
jgreener64, line 473: Leave this order as is to avoid changing the API where possible (also below).

src/units.jl
```diff
@@ -0,0 +1,266 @@
...
+function energy_remove_mol(x)
+    if dimension(x) == u"𝐋^2 * 𝐌 * 𝐍^-1 * 𝐓^-2"
+        T = typeof(ustrip(x))
+        return x / T(Unitful.Na)
+    else
+        return x
+    end
+end
+
+# function energy_add_mol(x, energy_units)
```
jgreener64, line 248: Remove commented lines.

src/spatial.jl
```diff
@@ -421,51 +422,110 @@ function wrap_coords(v, boundary::TriclinicBoundary)
...
+function maxwell_boltzmann(mass::Unitful.Mass, temp::Unitful.Temperature, k::BoltzmannConstUnits; rng=Random.GLOBAL_RNG)
     T = typeof(convert(AbstractFloat, ustrip(temp)))
     σ = sqrt(k * temp / mass)
     return rand(rng, Normal(zero(T), T(ustrip(σ)))) * unit(σ)
 end
 
-function maxwell_boltzmann(mass, temp; rng=Random.GLOBAL_RNG)
-    k = unit(temp) == NoUnits ? ustrip(mb_conversion_factor) : mb_conversion_factor
-    return maxwell_boltzmann(mass, temp, k; rng=rng)
+function maxwell_boltzmann(mass::MolarMass, temp::Unitful.Temperature; rng=Random.GLOBAL_RNG)
```
jgreener64, line 514: This could be rolled into the below method with `k_molar=Unitful.k * Unitful.Na` (also above)?

jgreener64 (commented):
Nearly there!

src/types.jl
ejmeitz: Does that also include the order of kwargs or just the order of types in the constructors?

ejmeitz: Like I can revert it back to L, K, F, E, M but I think `k` has to be the last kwarg (or at least after `energy_units`)

src/spatial.jl
```diff
@@ -421,51 +422,110 @@ function wrap_coords(v, boundary::TriclinicBoundary)
...
+function random_velocity(mass::Union{Unitful.Mass, MolarMass}, temp::Unitful.Temperature; dims::Integer=3, rng=Random.GLOBAL_RNG)
+    return SVector([maxwell_boltzmann(mass, temp; rng=rng) for i in 1:dims]...)
+end
+
+function random_velocity(mass::Union{Unitful.Mass,MolarMass}, temp::Unitful.Temperature,
+     k::Union{BoltzmannConstUnits, MolarBoltzmannConstUnits}; dims::Integer=3, rng=Random.GLOBAL_RNG)
+    return SVector([maxwell_boltzmann(mass, temp, k; rng=rng) for i in 1:dims]...)
+end
+
+function random_velocity(mass::AbstractFloat, temp::AbstractFloat, k::AbstractFloat; dims::Integer=3, rng=Random.GLOBAL_RNG)
```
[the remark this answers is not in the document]
ejmeitz: on all 3 or just the `AbstractFloat` one?

test/zygote.jl
ejmeitz: I'm not that familar with Zygote but I do not think this error is from my changes:

```
Compiling Tuple{typeof(Molly.convert_k_units), Type{Float64}, Float64, Unitful.FreeUnits{(), NoDims, nothing}}: try/catch is not supported.
  Refer to the Zygote documentation for fixes.
  https://fluxml.ai/Zygote.jl/latest/limitations
```

jgreener64: I can look into that after merge.

src/types.jl
jgreener64: Just the type order.

src/spatial.jl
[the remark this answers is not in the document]
jgreener64: The `AbstractFloat` one since that lacks a default.

ejmeitz:
Also I noticed the barostat tests fail every once in awhile. I'm pretty sure they're stochastic and the ranges aren't quite wide enough. They usually fail pretty close to the maximal values allowed.

test/zygote.jl
ejmeitz: I think it might be the `@assert` macros but commenting them out did not seem to fix the issue.

PR #144, opened 2023-07-09
[WIP] Add support for .top force field to MolecularForceField constructor

bondrewd:
This PR adds support for parsing Gromacs `.top` and `.itp` files and refactors the MolecularForceField constructor to enable initializing systems using force fields and topologies in Gromacs format.

### DIFF
--- a/README.md
+++ b/README.md
@@ -134,4 +134,4 @@ A paper involving more contributors with further details on the software will be
 
 ## Interested?
 
-An [advert for a PhD position](https://www2.mrc-lmb.cam.ac.uk/students/international-phd-programme/projects/joe-greener) in Cambridge, UK that will involve significant coding in Molly is open until 5th December 2023. An advert for a postdoc will open in October 2023. Contact Joe Greener for more information.
+An [advert for a postdoc position](https://www.findapostdoc.com/search/job-details.aspx?jobcode=11512) (closes 30/10/23) and an [advert for a PhD position](https://www2.mrc-lmb.cam.ac.uk/students/international-phd-programme/projects/joe-greener) (closes 05/12/23) in Cambridge, UK that will involve significant coding in Molly are open. Contact Joe Greener with informal enquiries.
--- a/docs/src/documentation.md
+++ b/docs/src/documentation.md
@@ -21,7 +21,7 @@ n_atoms = 100
 atom_mass = 10.0u"g/mol"
 atoms = [Atom(mass=atom_mass, σ=0.3u"nm", ϵ=0.2u"kJ * mol^-1") for i in 1:n_atoms]
 ```
-See the [Unitful.jl](https://github.com/PainterQubits/Unitful.jl) docs for more information on the unit annotations.
+See the [Unitful.jl](https://github.com/PainterQubits/Unitful.jl) documentation for more information on the unit annotations.
 Molly re-exports Unitful.jl, [StaticArrays.jl](https://github.com/JuliaArrays/StaticArrays.jl) and [AtomsBase.jl](https://github.com/JuliaMolSim/AtomsBase.jl) since they are often required to run simulations.
 You can use your own atom types in Molly, provided that the [`mass`](@ref) function is defined and any fields required by the interactions are present.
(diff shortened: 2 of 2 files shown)


bondrewd:
@jgreener64 After implementing the `.top` parser, I found refactoring the current MolecularForceField structure is necessary. Currently, it is specific to OpenMM force fields, fixing the types of some interactions (PeriodicTorsion for dihedrals). I suggest making the MolecularForceField more general, allowing the representation of all the interactions currently supported in Molly. For example, right now RBTorsionTypes cannot be represented.

Regarding the Gromacs `.top` parser, it is done. I added it to a new file names `parsers.jl` to avoid cluttering the `setup.jl` file. It supports the preprocessor directives `#include`, `#define`, `#ifdef`, `#ifndef`, `#else`, and `#endif`. The only change that I will add that came to my head while writing this is support for uppercase and mix-case directives since, for example, "[ atoms ]" and "[ ATOMS ]" are both valid. The parser is still far from optimal, but optimizing it can be pushed into the future.

My next step for completing this PR will be refactoring the MolecularForceField structure to increase its representativity of the current interactions implemented in Molly.

ejmeitz:
Hey,

I've been trying to re-write some code so that Molly is more strict on what units you're allowed to use. One of the latest issues I ran into was actually with the MolecularForceField constructor. Specifically, I am trying to limit the mixture of molar and non-molar quantities. At the moment the default units use `u` for mass and `kJ/mol` for energy which is of course mixing molar and non-molar. Are these units standard for whatever file is being parsed here? Mixing units like this causes major issues down the line with Unitful and the branch I have open currently won't let you construct a `System` with this type of units. How should I change this part of the code to ameliorate this?

Specifically this code:
```julia
if units
        M = typeof(T(1u"u"))
        D = typeof(T(1u"nm"))
        E = typeof(T(1u"kJ * mol^-1"))
        K = typeof(T(1u"kJ * mol^-1 * nm^-2"))
    else
        M, D, E, K = T, T, T, T
    end
```

bondrewd:
@ejmeitz Hi!

I think there are more places where masses are parsed as numbers with the `u"u"` unit:

https://github.com/JuliaMolSim/Molly.jl/blob/e353dfbd444e2ac4ea05c4aae0b58b3658815445/src/setup.jl#L211

One solution would be to add an argument in the constructor for the mass type and set `u"g/mol"` as the default value `constructor(... ; ..., M=u"g/mol", ...)`.

bondrewd:
@ejmeitz As for the standard mass units, it depends on the force field. For example, in Gromacs, the mass unit in the `.top` files [is defined as `u"u"`](https://manual.gromacs.org/documentation/2019/reference-manual/definitions.html).

ejmeitz:
Cool thanks for the details. I guess I could also auto-convert `u` into `g/mol` after the parsing since they're basically identical. I wonder if Gromacs assumes `1u = 1g/mol` internally or if they deal with all the stupid conversions.

That said, looking at your changes I probably can't add this feature to my branch without having a conflict with the changes you're adding.

jgreener64:
Looks good so far.

> I suggest making the MolecularForceField more general, allowing the representation of all the interactions currently supported in Molly

Yes I agree.

I think units should probably change from `u` to `g/mol` throughout the whole package and docs, but that discussion (#140) can take place separate from this PR which can be rebased later if need be.

bondrewd:
This weekend I made the `.top` parser type stable. I am refactoring the `MolecularForceField` structure, constructor, and the `System` constructor using a `MolecularForceField` the following week.

@jgreener64 I would like to include a `.top` and a `.itp` as input for some integration tests (since the code coverage is failing). I can think of three options:
- Add the content as raw strings in the `/test/parser.jl` file
- Add the `.top` and `.itp` files for testing in the `/data/` directory
- Add the `.top` and `.itp` files for testing in a new directory: `/test/data/`
Is any of these ok? Or would you happen to have another idea?

jgreener64:
Sounds like good progress.

I would add them to `data` for now. Strings of only a few lines can go in the test file directly.

PR #145, opened 2023-07-21
Fix fieldnames bug in AtomsBase interface

ejmeitz:
Realized fieldnames was called on the system instance not the typeof the system which is not right.


ejmeitz:
How do I change the `SimpleCrystals` compat to be just 0.1.5?? I changed stuff in 0.1.6 that broke a test here.

jgreener64:
`= 0.1.5` but better to fix the test and up the compat to `0.1.6`?

If you are following semver with that package you should also bump to 0.2.0 if there are breaking changes.

ejmeitz:
Yeah I'll fix them, you mind if I just add that in this PR? 

I didn't actually change the interface just the way I coded it in Molly just didn't use the interface (oops). I accessed a property directly (`atom.sym`) but I changed it to atomic_symbol in 0.1.6. I should just be calling the `atomic_symbol` function.

jgreener64:
Yes add it to this PR.

ejmeitz:
Ok changes made, no clue why the agent tests fail in one of the test sets and pass in the others.

PR #150, opened 2023-09-28
[WIP] OverdampedLangevin simulator and PotentialFunction interaction

axsk:
I would really like to have these functions (And eventually a `GirsanovOverdampedLangevin` simulator).

This is still work in progress.


src/interactions/potential.jl
```diff
@@ -0,0 +1,22 @@
...
+
+    return V.f(sys.coords)
+end
+
+function Molly.forces(V::PotentialFunction,
+    sys,
+    neighbors=nothing;
+    n_threads=Threads.nthreads())
+
+    return ForwardDiff.gradient(V.f, sys.coords)
```
axsk, line 19: This is currently not working because of the units (and maybe the Vector{SVector} as well).

ustrip_vec.() could help getting rid of the units, reinterpret with getting rid of the ArrayOfStructs (AoS).
But we would have to reconstruct this afterwards.
I have to say the units create quite a headache for me whenever I have to deal with them.
Maybe we need a convenient conversion interface between the combinations of AoS/Matrix/Vector with and without units?
In the above case something like
```
with_dropped_units(sys.coords, as=Vector, reconstruct=true) do x
  ForwardDiff.gradient(V.f, x)
 end
 ```
 would be great that would transform the coords to a single unitfree vector, call gradient and transform back to a unitful AoS.

Comment (axsk) [reply src/interactions/potential.jl]: Switching to Zygote (of course...) resolved that problem.. Kind of. Zygote seems to struggle with reinterpret, that might be fixed by a gradient rule though..

I still would love to have a simple way to switch between units/no units and AoS/vec

File: src/simulators.jl [hunk 2]
```diff
@@ -930,3 +930,54 @@ function random_unit_vector(float_type, dims)
     vec = randn(float_type, dims)
     return vec / norm(vec)
 end
+
+
+"""
```
jgreener64, line 935: Move this up under the other Langevin simulators, and export it.

src/simulators.jl
```diff
@@ -930,3 +930,54 @@ function random_unit_vector(float_type, dims)
...
+    OverdampedLangevin(; <keyword arguments>)
+
+Simulates the overdamped Langevin equation using the Euler-Maruyama method.
```
jgreener64, line 938: Is there a canonical paper to reference here?

jgreener64 (commented):
Looks good, I'm not sure the potential function stuff is worth putting in the main Molly codebase though.

Would be good to add `OverdampedLangevin` as a simulator in the `"Langevin splitting"` test set, or maybe there is a better way to test some property resulting from an overdamped Langevin simulation.

There was a discussion about how different atomic masses affect the time step, is this implementation consistent with the conclusions of that discussion?

axsk:
> Looks good, I'm not sure the potential function stuff is worth putting in the main Molly codebase though.

I removed that part
> Would be good to add `OverdampedLangevin` as a simulator in the `"Langevin splitting"` test set, or maybe there is a better way to test some property resulting from an overdamped Langevin simulation.

It is actually simulating another dynamics then (underdamped) Langevin. The two would conincide only in the limit `friction` $\rightarrow \infty$
I added a line to the Lennard-Jones test case to at least see that it runs without error.

> There was a discussion about how different atomic masses affect the time step, is this implementation consistent with the conclusions of that discussion?

This is now consistent with these considerations. The units are accounted for by taking the (finite) friction into account.

src/simulators.jl
axsk: The Euler Maruyama scheme is pretty standard, even Wikipedia forgoes the reference..
Considering the limit underdamped Langevin $\rightarrow$ overdamped Langevin and the resulting unit considerations I would point to the "Free energy computations" book by Lelievre, Rousset and Stoltz, but its merely hinted at there..

jgreener64:
Great, thanks.

PR #155, opened 2023-10-25
Fix SimpleCrystals Bugs

ejmeitz:
Tests pass locally. Forgot that Molly stores coords as `Vector{SVector{N,T}}` so just had to update that in the Crystal constructor to fix things as SimpleCrystals no longer stores its positions as `Vector{SVector{N,T}}`.


jgreener64:
Thanks.

PR #160, opened 2023-11-29
relax enzyme version requirement

axsk:
The old requirement  "=0.11.2" inhibited installation of the curernt DiffEqBase.

I don't know why you pinned the Enzyme version, so you very well might not wan't to just merge this.


jgreener64:
Just waiting on https://github.com/EnzymeAD/Enzyme.jl/issues/1166 before relaxing the version requirement. There have been a few issues on recent Enzyme versions.

GPU gradients are the only thing that doesn't work with Enzyme v0.11.10 though, so if you don't need that you can relax it locally.

Issue #101, opened 2022-09-17
Molly needs a logo!


JaydevSR:
As the title says, I think Molly needs a cool logo. Most Julia packages have made their logos using [Luxor.jl](https://github.com/JuliaGraphics/Luxor.jl) maybe one can be made for Molly as well (although I have no experience with Luxor).

jgreener64:
Yes this sounds like a good idea. Something similar to the JuliaMolSim logo (https://github.com/JuliaMolSim) but referencing bonds/molecules/macromolecules would be cool.

JaydevSR:
I experimented a bit with http://qutemol.sourceforge.net/, using this we have an illustration of 1cqh

<image width=400 src="https://user-images.githubusercontent.com/59474411/191063182-fc073a3b-61af-484f-b3c2-e933d7fbc552.png"></image>

@jgreener64  What do you think about using such figures for constructing the logo (in place of the classic julia dots).

jgreener64:
That looks good, a related idea is to find a protein trimer and colour each chain so that it looks like the Julia logo.

[...]

PR #165, opened 2024-02-04
add Molly logo

LegionAtol:
The .png and .svg Molly logo as discussed in issue #101


jgreener64:
The PNG looks great. I can't seem to open the SVG as a vector graphic in something like Inkscape though, am I missing something?

LegionAtol:
I generated the .svg from the .png with an online tool to have it as similar as possible, then I tried to import it into Canva(online tool) and it displayed it correctly, so I thought that if it were to be used within websites or online it would be displayed correctly.
I hadn't tried opening it with vector graphics programs but if needed I can recreate it from scratch with one of these programs

jgreener64:
How did you generate it in the first place, is it possible to save as SVG from there? Ideally we would have a vectorised form to allow us to resize without pixelation.

LegionAtol:
I created the png with Procreate (iPad) and it doesn't allow you to export the work to svg.
So I think there are two options: either I convert the png to svg with an online tool (as I did) or I create it from scratch with a vector graphics program trying to make it identical.
let me know if you have any other ideas

jgreener64:
I think we do want a vector version. Sorry as it is more work.

jgreener64:
Great, thanks for doing this. I will add it to the readme.

Issue #162, opened 2024-01-13
sys.coords does not work as Vector{Vector}


ejmeitz:
Just noticed that if you initialize the coordinates of your system with a `Vector{Vector}` instead of a `Vector{SVector}` you get an issue in the forces function specifically line 163 of forces.jl in the `zero` function. It probably pops up elsewhere also. The heart of the problem is for some reason:
`zero([[1,2,3],[2,3,4])` is not valid but
`zero([SVector(1,2,3), SVector(3,2,3)]` is valid

Either we force the input to be a Vector of SVector or find a work around for the forces. I think that line is just trying to make a matrix the same shape as coords. `similar` might do the trick.

jgreener64:
I think it's fair to assume that the coordinates are some array of `SVector`s or a similar bits type. There is a big performance hit otherwise.

ejmeitz:
Makes sense. I'd like to add an error message about this though as the current error/crash is super unintuitive and I think its a pretty easy mistake to make. Will make a PR at some point.

PR #166, opened 2024-02-10
Warn if eltype positions/velocities are not isbits

ejmeitz:
From #162


jgreener64:
Great, maybe just change `Vector{SVector{}}` to `a vector of SVectors` since we also allow `CuArray`s.

ejmeitz:
Done

PR #170, opened 2024-03-29
Show logo normally on GH page

Ruibin-Liu:


jgreener64:
Great, thanks.

Ruibin-Liu:
It doesn't seem to work? I just checked the dev doc. The logo is not shown.

jgreener64:
Yes I noticed that too. Maybe Documenter.jl doesn't like the `` tags.

There might be a way to add styling to the `[alt_text](url)` image format used further down in the same file. As I recall I didn't use that format here because the image was the wrong size.

PR #178, opened 2024-05-20
Particle Mesh Ewald

ejmeitz:
A work in progress implementation of Smooth Particle Mesh Ewald on a GPU:

SPME follows the original paper:
[A_Smooth_Particle_Mesh_Ewald_Method (2).pdf](https://github.com/JuliaMolSim/Molly.jl/files/15371897/A_Smooth_Particle_Mesh_Ewald_Method.2.pdf)

Currently the missing major components are:
- An analytical derivative to calculate dQ/dR which is needed to calculate the force (a lapse in my understanding of the paper)
- Integration with Molly infrastructure, notably need to implement the particle-particle component as an interaction in Molly (should be easy)
    - If an improved GPU neighbor list is implemented in another PR that would be very useful here
- Implementation of the stress tensor calculation. This theory is currently beyond me and I have not spent time trying to figure it out.
- Thorough testing

Minor things:
- Certain small parts need to be written into GPU kernels still (e.g. calculating scaled coordinates u)
- Parallelize CPU version of charge interpolation
- Optimal choice of FFT plan for the reciprocal space
- I don't think the SPME struct as is can go on the GPU, so either need to use Adapt or write some kind of wrapper to handle moving things to the GPU at the start of the force loop
- Lots of things are just allocated inside functions that could be pre-allocated at the start


jgreener64:
An implementation of this is now on master.

PR #182, opened 2024-07-26
Switch from Zygote to Enzyme

jgreener64:
Switching from using Zygote+Enzyme to just using Enzyme for AD allows us to rewrite much of the main computation code to mutate memory. This improves performance on CPU and GPU.

Opening this PR to test with CI.


wsmoses:
@jgreener64 if any of these issues are still outstanding, I can try to quickly fix if you have MWE's

jgreener64:
Merging this to prevent development stalling elsewhere. There are a few things that need fixing before release related to Enzyme, GPU support and docs updates.

PR #200, opened 2025-06-13
GPU Constraints

ejmeitz:
This PR (WIP) implements SHAKE/RATTLE constraints on GPU using the M-SHAKE algorithm. It also has angle constraints!

The traditional SHAKE implementation is not possible to parallelize easily. M-SHAKE is parallelizable but comes at the cost of the types of constraints you can have. The "fast" implementation of M-SHAKE uses an analytical matrix inverse and is therefore limited to clusters of 3 constraints or less. I do not really see this as a problem. Typically you only constrain H-bonds or the bonds of the solvent molecules. Furthermore, bonds along the back bond are not constrainable due to the assumption of a "central" atom in my current implementation. This can be removed in the future, but I think implementing rigid body dynamics is the correct way to constraint a backbone and not SHAKE.

The types of clusters supported:
- Single bond
- 1 central atom bonded to 2 other atoms
- 1 central atom bonded to 3 other atoms
- 1 central atom bonded to 2 other atoms with the angle constrained

This leaves out things like CH4.

The current implementation is complete for all RATTLE variants and for SHAKE with a single bond as these are all analytical solutions and do not require iteration. The 3 remaining SHAKE kernels require iteration and I am not sure how best to handle that as load balancing is important for GPU performance. Open to suggestions as to how that should be done.

Some extra work will be required to actually support angle constraints. Maybe we leave this to another PR, and I just throw an error if I detect angle constraints.
- [x] Update things like n_dof which assume there are no angle constraints
- [x] Update build_clusters for angle constraints


ejmeitz:
Also I tried to be clever and sort the atoms before hand to remove uncoalesced memory accesses, but the sorting seems to dwarf the benefit from the coalesced reads. This implementation is really not clever at all, but I also do not see room to be clever. Anyway the constraints should be a tiny part of the total cost, so its better they are on GPU to mitigate data moving from CPU to GPU during the simulation.

jgreener64:
Great. I can't run the branch at the minute due to a missing import of StructArrays and a missing type parameter. Could you commit the working version?

ejmeitz:
@jgreener64 The 3-atom constraints test works as well now. This means the 4-atom constraints (1 central + 3 bonds) are un-tested (but should work as the kernels are similar). There is also still some work during the setup to get angle constraints supported, but the shake kernel is done and should work.

ejmeitz:
Ok all the code is written and seems to work. There are tests for diatomic, triatomic, 4 atom, angle, and angle + triatomic at the same time. None of these tests are on GPUs.

Things I don't love about the existing implementation:
- To avoid mutating the input to the `SHAKE_RATTLE` constructor when both `angle_constraints` and `distance_constraints` are passed I make a copy of distance_constraints inside of `build_clusters` : `dist_constraints = copy(dist_constraints)`
- The intermediate storage inside of each kernel has some weird types due to units and I have to do stuff like:
```
@uniform A_type = typeof(zero(L)*zero(L) / zero(M))
```
- I am not sure if the way I allocate A,C, lambda inside the shake/rattle kernels is correct for thread local storage. 
-`constraints_helper.jl` is a freakin mess. It might be slow for systems with many many constraints. It only needs to run once though.

Things that would be worth checking:
- Test a protein with explicit solvent. Constraining all H and all water
- Does it work on GPU (it should if KA.jl doesnt have bugs)
- Performance
    - How does it compare to prior implementation
    - How much faster in GPU vs. CPU
    - How many iterations to converge the various SHAKE kernels. Purportedly this is supposed to be as good as SETTLE for water.
   
Todo:
- Docs, but maybe I do this in a separate PR to avoid making this one even more complex

jgreener64:
Great, I'll try it this week. Do you have access to a GPU to test on?

ejmeitz:
Yeah I have access to an AMD and NVIDIA GPU. Just never used Molly with GPU so was easier to test on CPU. Out of town this week, but can test code on another system if you'd like.

There's a good chance some atray isn't moved to the GPU that should be but hopefully not!

jgreener64:
I wouldn't worry about performance for things that only run once.

I get the following error with CUDA:
```
ERROR: GPU compilation of MethodInstance for Molly.gpu_shake2_kernel!(::KernelAbstractions.CompilerMetadata{…}, ::Vector{…}, ::CuDeviceVector{…}, ::CuDeviceVector{…}, ::CuDeviceVector{…}, ::CubicBoundary{…}) failed
KernelError: passing non-bitstype argument

Argument 3 to your kernel function is of type Vector{Molly.ConstraintCluster}, which is not a bitstype:
  .ref is of type MemoryRef{Molly.ConstraintCluster} which is not isbits.
    .mem is of type Memory{Molly.ConstraintCluster} which is not isbits.


Only bitstypes, which are "plain data" types that are immutable
and contain no references to other values, can be used in GPU kernels.
For more information, see the `Base.isbitstype` function.

Stacktrace:
  [1] check_invocation(job::GPUCompiler.CompilerJob)
    @ GPUCompiler ~/.julia/packages/GPUCompiler/aNr5U/src/validation.jl:108
  [2] macro expansion
    @ ~/.julia/packages/GPUCompiler/aNr5U/src/driver.jl:87 [inlined]
  [3] macro expansion
    @ ~/.julia/packages/Tracy/GcShf/src/tracepoint.jl:158 [inlined]
  [4] compile_unhooked(output::Symbol, job::GPUCompiler.CompilerJob; kwargs::@Kwargs{})
    @ GPUCompiler ~/.julia/packages/GPUCompiler/aNr5U/src/driver.jl:85
  [5] compile_unhooked
    @ ~/.julia/packages/GPUCompiler/aNr5U/src/driver.jl:80 [inlined]
  [6] compile(target::Symbol, job::GPUCompiler.CompilerJob; kwargs::@Kwargs{})
    @ GPUCompiler ~/.julia/packages/GPUCompiler/aNr5U/src/driver.jl:67
  [7] compile
    @ ~/.julia/packages/GPUCompiler/aNr5U/src/driver.jl:55 [inlined]
  [8] #1181
    @ ~/.julia/dev/CUDA/src/compiler/compilation.jl:250 [inlined]
  [9] JuliaContext(f::CUDA.var"#1181#1184"{GPUCompiler.CompilerJob{…}}; kwargs::@Kwargs{})
    @ GPUCompiler ~/.julia/packages/GPUCompiler/aNr5U/src/driver.jl:34
 [10] JuliaContext
    @ ~/.julia/packages/GPUCompiler/aNr5U/src/driver.jl:25
 [11] compile(job::GPUCompiler.CompilerJob)
    @ CUDA ~/.julia/dev/CUDA/src/compiler/compilation.jl:249
 [12] actual_compilation(cache::Dict{…}, src::Core.MethodInstance, world::UInt64, cfg::GPUCompiler.CompilerConfig{…}, compiler::typeof(CUDA.compile), linker::typeof(CUDA.link))
    @ GPUCompiler ~/.julia/packages/GPUCompiler/aNr5U/src/execution.jl:245
 [13] cached_compilation(cache::Dict{…}, src::Core.MethodInstance, cfg::GPUCompiler.CompilerConfig{…}, compiler::Function, linker::Function)
    @ GPUCompiler ~/.julia/packages/GPUCompiler/aNr5U/src/execution.jl:159
 [14] macro expansion
    @ ~/.julia/dev/CUDA/src/compiler/execution.jl:373 [inlined]
 [15] macro expansion
    @ ./lock.jl:273 [inlined]
 [16] cufunction(f::typeof(Molly.gpu_shake2_kernel!), tt::Type{…}; kwargs::@Kwargs{…})
    @ CUDA ~/.julia/dev/CUDA/src/compiler/execution.jl:368
 [17] macro expansion
    @ ~/.julia/dev/CUDA/src/compiler/execution.jl:112 [inlined]
 [18] (::KernelAbstractions.Kernel{…})(::Vector{…}, ::Vararg{…}; ndrange::Int64, workgroupsize::Nothing)
    @ CUDA.CUDAKernels ~/.julia/dev/CUDA/src/CUDAKernels.jl:122
 [19] apply_position_constraints!(sys::System{…}, ca::SHAKE_RATTLE{…}, r_pre_unconstrained_update::CuArray{…}; kwargs::@Kwargs{…})
    @ Molly ~/.julia/dev/Molly/src/constraints/shake_gpu.jl:678
 [20] apply_position_constraints!
    @ ~/.julia/dev/Molly/src/constraints/shake_gpu.jl:658 [inlined]
 [21] #apply_position_constraints!#252
    @ ~/.julia/dev/Molly/src/constraints/constraints_helper.jl:344 [inlined]
 [22] apply_position_constraints!
    @ ~/.julia/dev/Molly/src/constraints/constraints_helper.jl:342 [inlined]
 [23] #simulate!#259
    @ ~/.julia/dev/Molly/src/simulators.jl:91 [inlined]
 [24] simulate!(sys::System{…}, sim::SteepestDescentMinimizer{…})
    @ Molly ~/.julia/dev/Molly/src/simulators.jl:65
 [25] top-level scope
    @ REPL[99]:1
```
If you could try on GPU, and adapt the tests to also run on GPU where available (see `for AT in array_list` elsewhere in the file), that would be great.

I also needed `CUDA.@allowscalar` during constraints setup and system setup, you can move things to the CPU for these steps as they only happen once.

The branchless functions might be replaceable with `cond ? a : b`.

ejmeitz:
I'll work on the first thing. I think I probably need to just use `replace_storage` from StructArrays.jl.

Where did you need the `@allowscalar` by the way? I think the constraints you pass to the System constructor should be on CPU and then I'll move them to whatever backend the coordinates/velocities are on. The "setup" process completely mangles the user input and uses Graphs.jl so it definitely has to be CPU.

Wouldn't the ternary operator introduce warp divergence? I'm not sure what the `?` lowers to inside a GPU kernel.

jgreener64:
> Wouldn't the ternary operator introduce warp divergence?

Ah I meant `ifelse`, my bad.

Fair enough about the constraints being on CPU. We could move them back to CPU for setup if they are on the GPU, so the user isn't confused by the error if they provide them on the GPU.

Project.toml
```diff
@@ -20,13 +20,15 @@ Graphs = "86223c79-3864-5bf0-83f7-82e725a168b6"
 KernelAbstractions = "63c18a36-062a-441e-b654-da1e3ab1ce7c"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 NearestNeighbors = "b8a86587-4115-5ab1-83bc-aa920d37bbce"
+OnlineStats = "a15396b6-48d5-5d58-9928-6d29437db91e"
```
jgreener64: Useful for debugging but I don't think it's worth the dependency.

Project.toml
```diff
@@ -20,13 +20,15 @@ Graphs = "86223c79-3864-5bf0-83f7-82e725a168b6"
...
 PeriodicTable = "7b2266bf-644c-5ea3-82d8-af4bbd25a884"
 Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
 Reexport = "189a3867-3050-52da-a836-e630ba90ab69"
 SimpleCrystals = "64031d72-e220-11ed-1a7e-43a2532b2fa8"
 SparseArrays = "2f01184e-e22b-5df5-ae63-d93ebab69eaf"
 StaticArrays = "90137ffa-7385-5640-81b9-e52037218182"
 Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
+StructArrays = "09ab397b-f2b6-538f-b94a-2f83cf4a842a"
```
jgreener64, line 33: Where is this used? Again would be good to avoid the dependency.

src/constraints/constraints_helper.jl
```diff
@@ -0,0 +1,454 @@
+export
```
jgreener64: Call this file constraints.jl?

src/constraints/constraints_helper.jl
```diff
@@ -0,0 +1,454 @@
...
+
+"""
+    DistanceConstraint(i, j, dist)
+
+
+    DistanceConstraint(i, j, dist)
+
+Constraint between two atoms that maintains a fixed distance between the two atoms.
+"""
+struct DistanceConstraint{D} <: Constraint{D}
```
jgreener64: Relaxing the integer type from `Int` to a parametric type `I` means it can take `Int32`, which makes setup from bonds easier. Allowing `Int32` throughout might help GPU performance a bit.

src/constraints/constraints_helper.jl
```diff
@@ -0,0 +1,454 @@
...
+
+For example, a water molecule can be defined as:
+```julia
+ac = AngleConstraint(1, 2, 3, 104.5u"°", 0.9572u"Å", 0.9572u"Å")
+ac = AngleConstraint(1, 2, 3, 104.5 * π / 180, 0.9572u"Å", 0.9572u"Å")
+```
+where atom 1 is oxygen and atoms 2/3 are hydrogen.
+"""
+struct AngleConstraint{D} <: Constraint{D}
+    i::Int # Central atom
```
jgreener64: Same as above about `Int`.

src/constraints/constraints_helper.jl
```diff
@@ -0,0 +1,454 @@
...
+
+to_distance_constraints(ac::AngleConstraint) = (
+    DistanceConstraint(ac.i, ac.j, ac.dist_ij), # arm 1
+    DistanceConstraint(ac.i, ac.k, ac.dist_ik), # arm 2
+    DistanceConstraint(ac.j, ac.k, ac.dist_jk) # angle
+) 
+
+function to_cluster(ac::AngleConstraint)
+
+    cluster = @SVector [
```
jgreener64: Nothing wrong with `@SVector`, but might be better to be consistent with the rest of the code and use `SVector(...)`.

src/constraints/constraints_helper.jl
```diff
@@ -0,0 +1,454 @@
...
+end
+
+
+struct ConstraintCluster{N, M, D <: DistanceConstraint, I <: Integer}
+    constraints::SVector{N,D}
+    unique_atoms::SVector{M, I}
+end
+
+#* WE ONLY USE THE DISTANCES AFTER EVERYTHING IS SORTED
+# struct KernelConstraintCluster{N, M, D, I <: Integer}
```
jgreener64: Remove?

src/constraints/constraints_helper.jl
```diff
@@ -0,0 +1,454 @@
...
+
+    raw_unique_idxs = unique([raw_is; raw_js])
+    N_unique_atoms = length(raw_unique_idxs)
+    N_constraints = length(raw_is)
+
+    is_single_bond = (N_constraints == 1)
+    is_angle_cluster = (N_constraints == 3 && N_unique_atoms == 3)
+
+    if is_single_bond || is_angle_cluster # no central atom, order agnostic
+        #* WHY DOES @SVECTOR NOT WORK???
```
jgreener64: Remove comments like this.

src/constraints/shake.jl
```diff
@@ -1,141 +1,182 @@
 export
     SHAKE_RATTLE,
     check_position_constraints,
-    check_velocity_constraints
+    check_velocity_constraints,
+    setup_constraints!
+
+const CC = ConstraintCluster #alias so next line is shorter
```
jgreener64: Unused?

src/constraints/shake.jl
```diff
@@ -1,141 +1,182 @@
...
-    dist_tolerance::D
-    vel_tolerance::V
+struct SHAKE_RATTLE{A, B, C, D, E, F, S}
+    clusters12::A
+    clusters23::B
+    clusters34::C
+    angle_clusters::D
+    dist_tolerance::E
+    vel_tolerance::F
+    gpu_block_size::Integer
```
jgreener64: Parametric type `I` for consistency elsewhere.

src/constraints/shake.jl
```diff
@@ -1,141 +1,182 @@
...
-                # Vector between the atoms before unconstrained update (r)
-                r12 = vector(coord_storage[k1], coord_storage[k2], sys.boundary)
+function SHAKE_RATTLE(n_atoms,
+                     dist_tolerance,
+                     vel_tolerance;
+                     dist_constraints = nothing,
+                     angle_constraints = nothing,
+                     gpu_block_size = 64, max_iters = 25)
+
+    @assert ustrip(dist_tolerance) > 0.0 "dist_tolerance must be greater than zero"
```
jgreener64: Prefer `ArgumentError`s here.

src/constraints/shake.jl
```diff
@@ -1,141 +1,182 @@
...
+
 end
 
+
 """
     check_position_constraints(sys, constraints)
 
 Checks if the position constraints are satisfied by the current coordinates of `sys`.
 """
 function check_position_constraints(sys, ca::SHAKE_RATTLE)
```
jgreener64: If these functions aren't specific to `SHAKE_RATTLE` then the argument type can be relaxed and they can be moved to constraints.jl.

src/constraints/shake_gpu.jl
```diff
@@ -0,0 +1,807 @@
+# The RATTLE Equations are modified from LAMMPS:
+# https://github.com/lammps/lammps/blob/develop/src/RIGID/fix_rattle.cpp
+
+
+# Returns a if condition is true, b otherwise (without branching)
+@inline function branchless_select(condition, a, b)
+    return condition * a + (!condition) * b
```
jgreener64: Preference for `ifelse` unless this is faster.

src/constraints/shake_gpu.jl
```diff
@@ -0,0 +1,807 @@
...
+
+@inline function solve2x2exactly(λ, A, C)
+
+    determinant = A[1, 1] * A[2, 2] - A[1, 2] * A[2, 1]
+
+    if iszero(determinant)
+        error("SHAKE determinant is zero, cannot solve")
+    end
+
+    inv_det = 1.0 / determinant
```
jgreener64: `1.0` is `Float64` and will promote the type, use `inv` or `one(T)`. Might speed things up on GPU for `Float32`. Also in other functions.

src/simulators.jl
```diff
@@ -166,10 +166,11 @@ end
 
     for step_n in 1:n_steps
         if using_constraints
-            cons_coord_storage .= sys.coords
+            copy!(cons_coord_storage, sys.coords)
```
jgreener64: I don't think this makes much difference, if it doesn't it is probably best to retain consistency with the other simulators.

src/types.jl
```diff
@@ -507,6 +507,7 @@ mutable struct System{D, AT, T, A, C, B, V, AD, TO, PI, SI, GI, CN, NF,
     energy_units::E
     k::K
     masses::M
+    inv_masses::IM
```
jgreener64, line 514: Whilst we could store this, I wonder if you can use `masses` with `/` rather than `*`.

src/constraints/shake.jl
```diff
@@ -149,13 +190,16 @@ Checks if the velocity constraints are satisfied by the current velocities of `s
 """
 function check_velocity_constraints(sys::System, ca::SHAKE_RATTLE)
```
jgreener64: For a GPU system this required `@allowscalar`, maybe there is a way to use a broadcast or reduction here?

Project.toml
ejmeitz: I have a separate branch that should have better performance that uses this.

src/constraints/shake.jl
ejmeitz: This is generic if the constraint algorithm implements `cluster_keys`

src/types.jl
ejmeitz: I do think this is worth pre-computing. It can be used in a lot of other places (e.g. accelerations). Most MD codes I've seen pre-compute this.

src/constraints/shake.jl
ejmeitz: Could be implemented as a reduction with `max`, this would require an extra temporary allocation but would make GPU possible

src/Molly.jl
```diff
@@ -18,12 +18,14 @@ using EzXML
...
 using Graphs
 using KernelAbstractions
+import KernelAbstractions as KA
 using NearestNeighbors
 using PeriodicTable
 using SimpleCrystals
 using Unitful
 using UnitfulAtomic
 using UnsafeAtomicsLLVM
+using StructArrays
```
jgreener64, line 30: Is this dependency used anywhere?

jgreener64 (commented):
Great, there are a few changes I want to make here but I can do it later.

Just one query about the StructArrays dependency.

Also you will have to merge in the `ewald` branch changes I just pushed to master, sorry about that.

src/Molly.jl
ejmeitz: Its used to store the constraint data. All the `build_clusters` functions return a `StructArray` of the various `ConstraintKernelData` structs in the `constraint_kernel_helper.jl` file.

PR #205, opened 2025-07-28
Track Image Flags

ejmeitz:
This PR adds an array of image flags which are updated by the `wrap_coordinates` function. This enables calculating displacements even when atoms move over the box boundary. This allows us to calculate things like MSD and displacement. 

I also added two loggers `ImageFlagLogger` and `DisplacementLogger`. I might also update the RMSD calculation in analysis.jl to reflect this change.

### DIFF
--- a/src/loggers.jl
+++ b/src/loggers.jl
@@ -23,7 +23,9 @@ export
     AutoCorrelationLogger,
     AverageObservableLogger,
     ReplicaExchangeLogger,
-    MonteCarloLogger
+    MonteCarloLogger,
+    ImageFlagLogger, 
+    DisplacementLogger
 
 """
     apply_loggers!(system, neighbors=nothing, step_n=0, run_loggers=true;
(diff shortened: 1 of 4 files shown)


jgreener64:
Nice idea but I'm not sure this is worth the overhead in general.

The idea of a `DisplacementLogger` is good though. It wouldn't be a `GeneralObservableLogger` because it wouldn't store a value for each time point. It could store the coordinates from the last snapshot and the overall displacements. Each snapshot it would calculate the displacements from the last snapshot with `vector` and add these to the overall displacements.

ejmeitz:
Ok this is updated. I tried to add a test, but not really sure how to know what's "right" except at the first timestep.

Also no clue why there's changes in simulators.jl, my fork is up to date.

ejmeitz:
If the user calls `simualte!` multiple times, I think the initialization logic which is based on `step_n` being 0 will not be correct. The displacements should carry over from the last time `simulate!` was called.

src/loggers.jl
```diff
@@ -956,3 +958,62 @@ function log_property!(mcl::MonteCarloLogger{T},
...
+    DisplacementLogger(T, n_steps; n_update::Integer=1, dims::Integer=3)
+
+Log the displacements of atoms in a system throughout a simulation. Displacements are
+updated every `n_update` steps and saved every `n_steps` steps.
+
+The logger assumes a particle does not cross 2 periodic boxes in `n_update` steps. 
+By default `n_update` is set to one to mitigate this assumption, but it can be 
+set to a higher value to reduce cost. `n_steps` must be a multiple of `n_update`. 
+"""
+mutable struct DisplacementLogger{A, B}
```
jgreener64: `DisplacementsLogger` might be more consistent with the other names.

test/simulation.jl
```diff
@@ -380,14 +384,19 @@ end
...
+            loggers=(
+                coords=CoordinatesLogger(100),
+                disp=DisplacementLogger(100)
+            ),
         )
     end
 
     for simulator in simulators
         @time simulate!(s, simulator, n_steps; n_threads=1)
+        @test sum(first(values(s.loggers.disp))) == 0.0u"nm"
```
jgreener64: Doesn't this return a `SVector`?

jgreener64:
Great, thanks.

PR #208, opened 2025-08-01
Add Type Hierarchy to Interactions

ejmeitz:
Added `NBodyInteraction{N}`, `GeneralInteraction` and `SpecificInteraction` among other things. This PR is not quite enough to let me implement thermodynamic integration, but I'll try to keep the changes bite-sized.

I think we could do a lot in the future by dispatching on these, but to keep changes minimal for now and avoid breaking changes I have opted to only add a new constructor which allows passing all the interactions together, parsing them into their respective types and then calling the original constructor. 

I added the type to each interaction. I've never used the specific interactions before, but I from what I understand I do not need to mark things like `CosineAngle` or `UreyBradley` with `SpecificInteraction` as they are intended to be passed to a `SpecificAtomsInteractionList` type?

### DIFF
--- a/src/types.jl
+++ b/src/types.jl
@@ -1,6 +1,7 @@
 # Types
 
 export
+    PairwiseInteraction,
     InteractionList1Atoms,
     InteractionList2Atoms,
     InteractionList3Atoms,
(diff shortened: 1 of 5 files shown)


ejmeitz:
I will update the docs if you think this change is ok.

src/types.jl
```diff
@@ -25,14 +25,33 @@ export
 
 const DefaultFloat = Float64
 
+"""
+Base type for all interaction types.
+"""
+abstract type Interaction end
```
jgreener64: Export all of these apart from `Interaction`.

src/types.jl
```diff
@@ -25,14 +25,33 @@ export
...
+"""
+abstract type Interaction end 
+
+"""
+Base type for all n-body interactions. `N` denotes the body order
+"""
+abstract type NBodyInteraction{N} <: Interaction end
+
+
+abstract type SpecificInteraction <: Interaction end
```
jgreener64: Is there an argument for `SpecificInteraction{N}` here? Maybe there are cases where that dispatch could be useful.

src/types.jl
```diff
@@ -609,6 +628,55 @@ function System(;
...
+"""
+    System(sys; <keyword arguments>)
+
+Convenience constructor for `System` that takes all interactions in one tuple.
+These are passed through the `interactions` kwarg.
+"""
+function System(;
+                atoms,
+                coords,
+                boundary,
```
jgreener64: You could use `kwargs...` to save lines here.

src/types.jl
```diff
@@ -25,14 +25,33 @@ export
...
+abstract type Interaction end 
+
+"""
+Base type for all n-body interactions. `N` denotes the body order
+"""
+abstract type NBodyInteraction{N} <: Interaction end
+
+
+abstract type SpecificInteraction <: Interaction end
+abstract type GeneralInteraction <: Interaction end
```
jgreener64: I think we should not have `GeneralInteraction` and have it as the "default" for dispatch later. The reason is that AtomsCalculators.jl calculators that don't know about Molly can use this path.

jgreener64 (commented):
Looks good, I think only the docs for pairwise interactions need to change.

ejmeitz:
On vacation rest of the week, but the constructor ends up being a redefinition of the existing constructors. So will need to fix that.

TODO: docs + test for new constructors + fix constructor definition.

jgreener64:
Thanks, merging. We can add methods that use this for dispatch in the future.

PR #213, opened 2025-08-19
LAMMPS Calculator

ejmeitz:
A new package extension for LAMMPS.jl that defines a general interaction which uses LAMMPS to calculate force and energies. It even uses OpenMP to do multi-threaded calculations!

This requires a new tagged version of LAMMPS.jl before we can merge as I made some PRs there to enable in-place calculation of forces. 

In principle this approach could be used to have LAMMPS calculate properties of the system as well.


ejmeitz:
Yeah I decided its just easier to have it here lol. I can live with downloading Molly.

I'll add some tests for this soon. The energies do appear to be correct, but I'm having some issues with multi-threading.

ejmeitz:
@jgreener64 This is ready (save for documentation). I ran the tests manually and they work, but I'm not really sure how to setup tests for a package extension? The `ase_calculator` tests just don't look like they run.

jgreener64:
On balance I think this might be better off as an extension in LAMMPS.jl.

Extensions are available for testing assuming the required packages are imported. I didn't get round to setting up the ASE tests on CI.

ejmeitz:
Yeah I can ask them if they'd be ok adding this there. I do still thing this interoperability should be advertised in Molly somewhere. Cause to be honest if I can't use LAMMPS through Molly, then I would never consider Molly for my research.

jgreener64:
Ah I thought LAMMPS was your project, which was one of the reasons for suggesting it go there.

Maybe it could go in Molly, I'll take a look.

ejmeitz:
I also asked them over at LAMMPS.jl as well. In principle this could be a separate package as well, I just don't want it lost as I think its very useful.

jgreener64:
I agree it is useful and we can document it here either way.

PR #216, opened 2025-09-05
Virial Branch - Quasi complete virial, Improved Pressure & Temperature coupling

allexandre97:
## Virial implementation

Follows the definition in the [LAMMPS docs](https://docs.lammps.org/compute_stress_atom.html). For now the contribution of the KSpace (must research how would one even compute the virial there) and the constraints are not taken into account. We hope their contribution is small enough and can be ignored. 

Note that the 1/N prefactor is ignored in the CPU path. This is because the LAMMPS definition is given *per atom* and assigns the virial contribution equally to each atom involved in the interaction. However, Molly goes over interactions and thus this factor can be ignored.

On the other hand, the GPU path (Giuseppe's kernel in particular) does go over atom pairs. The already present functionality to accumulate forces was duplicated and copied to accumulate the virial. Each row of the virial matrix is treated as a vector and the same kind of accumulation is done in the shared memory and buffers. This implies that we accumulate the virial for each atom (as in LAMMPS) and thus the 1/2 prefactor appears again. The AbstractGPUKernels are equivalent in syntax to the CPU path and do not need the prefactors either.

### Box scaling

Big rewrites in the pressure coupling schemes. In general, the box and coordinate scaling has been changed to accommodate anisotropic and shear stresses. Now, the scaling is done through a tensor $\boldsymbol{\mu}$ instead of through a scalar, allowing independent control of each simulation box axis and the relative angles between them. Generally speaking the scaling is done in 2 steps:

1. Obtain a new box matrix $\mathbf{B}' = \boldsymbol{\mu} \mathbf{B}$
2. Transform from old to new reference frame $\mathbf{r}'= \mathbf{B}' \mathbf{B}^{-1} \mathbf{r}$

The actual process is a bit more involved, requires scaling just the centers of mass of the molecules to avoid distortions on their connectivity, and separating the scaling matrix into a purely rotational and a purely stretch term (done through SVD). After the centers of mass are scaled, the individual molecules are rotated as a rigid body to account for shear deformations.

### Berendsen

With this functionality in place, the Berendsen barostat was extended to handle isotropic (xyz), semiisotropic (xy, z) and anisotropic (x, y, z, xy, xz, yz) pressure coupling. The coupling type is chosen with a symbol passed as a keyword when the barostat is created. Accepts `:isotropic` `:semiisotropic` and `:anisotropic`. The specific implementations is based on the one found in the GROMACS source code.

### Stochastic Cell Rescaling

This should be the preferred barostat choice in general, it is a 1st order weak coupling scheme but specifically designed to correctly reproduce isobaric ensembles. Follows [Bernetti & Bussi (2020)](https://doi.org/10.1063/5.0020514) for the isotropic and semiisotropic pressure control and [Del Tatto, Raiteri, Bernetti & Bussi (2022)](https://doi.org/10.3390/app12031139) for full anisotropic control with shear stresses. Acts in a very similar way to the Berendsen coupling scheme, but adds a stochastic term modeled as a Wiener process which depends on the temperature of the system. The coupling type is also decided with a Symbol passed as a keyword when the barostat is created. Accepts `:isotropic` `:semiisotropic` and `:anisotropic`.

### Monte Carlo Barostat

Merged all three previous struct definitions into a single MC barostat type. The coupling type is decided as for the other barostats, with a symbol keyword passed at the time the barostat is created. Accepts `:isotropic` `:semiisotropic` and `:anisotropic`. The way non-isotropic coupling is handled has also changed. Previously, at each rescaling trial, each dimension of the simulation box was treated separately meaning that the deformation only happened either along a specific box vector selected at random or completely isotropically. Instead of that, this is the new procedure (described for the anisotropic case):

1. Select a random volume increment $dV$ the same way as in the previous implementation.
2. Calculate the volume scaling fraction:
$$\bar V = \frac{V + dV}{V}$$
3. Draw three random number $w_1, w_2, w_3$ and normalise them such that:
$$\hat{w_i} = \frac{w_i}{\sum w_i}$$
4. The scaling for each box vector $i$ with length $l_i$ is determined by:
$$\bar{l_i} = \bar V^{\hat{w_i}}$$

This introduces a coupling between the box vectors and therefore the work term in the energy acceptance criterion is no longer $dW = p dV$. Without going into much detail, the new expression for the work is:

$$dW = (V + dV) \cdot \ln \left( \frac{V + dV}{V} \right) \sum p_{i,i} \hat{w_i} $$

Note that shear forces are not considered for this barostat implementation. If shear forces are needed, the Berendsen or CRescale methods should be used.

### Stochastic Velocity rescaling

The velocity rescaling thermostat was implemented following its description in [Bussi, Donadio & Parrinello (2007)](https://doi.org/10.1063/1.2408420). This scheme is very similar in spirit to the stochastic cell rescaling pressure coupling, taking as a reference the Berendsen thermostat and adding a stochastic term to allow the correct sampling of isothermal ensembles. The construction of the stochastic term requires some careful random number generation, described in the appendix (eq. A7, in particular gives the expression for the overall velocity scaling factor) in the paper mentioned in this section.

### Triclinic boundary

The inclusion of shear stresses in the barostat lead to a rewrite of the Triclinic boundary. It was found that, due to numerical errors, triclinic boundaries produced spurious results if they happened to have a orthogonal configuration of vectors. Now, due to the shear forces at play, triclinic boundaries starting from skewed conformations can move towards cuboid-type conformations; or simulations may be decided to be started from cuboid-like conformations and let them evolve to skewed cases. Numerical safeguards against vanishingly small and/or exploding terms are added to ensure that the triclinic boundaries are always valid regardless of their particular geometry.

### Molecule PBC unwrap

Functionality was added to save coordinate and trajectory files that keep the connectivity of the molecules intact. This allows for easier visualization of simulation results. In brief, molecules are scaled to fractional coordinates, an adjacency list for connected atoms is built using the system topology, a breadth-first search is performed over all the atoms reconstructing the entire molecules and finally this reconstructed coordinates are scaled back to cartesian space.

### Loggers

The addition of the virial and pressure as a tensors implied the modification of the correspondent property loggers. Now both `VirialLogger` and `PressureLogger` return matrices instead of scalars. The user is responsible of the appropriate manipulation of the matrices in order to obtain the specific magnitudes they may be looking for (*e.g.* $\frac{1}{3} \mathrm{tr}\left[ \mathbf{P} \right]$ for the scalar pressure). 

The `TrajectoryWriter` logger was also modified to take an extra Symbol argument, `correction`, to decide if molecules should be PBC unwrapped or not. The currently accepted values are `:pbc` and `:wrap`, respectively.

### Tests

Unit tests were written for the isotropic, semiisotropic and anisotropic variants of the Berendsen, CRescale and MonteCarlo barostat. The unit test are run both for the CPU and GPU cases. In general, they follow the exact same design philosophy as the previous equivalent unit tests. The differences lie on checking the correctness of individual pressure components instead of the overall scalar pressure in the non-isotropic cases; and the correctness of the volume of the simulation box instead of the lengths of the boundaries.

### DIFF
--- a/docs/src/documentation.md
+++ b/docs/src/documentation.md
@@ -1107,6 +1107,8 @@ needs_virial(c::MyCoupler) = (truth = true, steps = c.n_steps)
 # In case you do NOT need the virial
 needs_virial(c::MyCoupler) = (truth = false, steps = Inf)
 ```
+The use of the [`virial`](@ref) tensor allows for non-isotripic pressure control. Molly follows the [definition](https://docs.lammps.org/compute_stress_atom.html) of LAMMPS, and takes into account specific and pairwise interactions;
+as well as the contribution of the K-space of the [`Ewald`](@ref) and [`PME`](@ref) methods.
 
 ## Loggers
 
@@ -1123,7 +1125,9 @@ The available loggers are:
 - [`VolumeLogger`](@ref)
 - [`DensityLogger`](@ref)
 - [`VirialLogger`](@ref)
+- [`ScalarVirialLogger`](@ref)
 - [`PressureLogger`](@ref)
+- [`ScalarPressureLogger`](@ref)
 - [`DisplacementsLogger`](@ref)
 - [`TrajectoryWriter`](@ref)
 - [`StructureWriter`](@ref)
--- a/src/coupling.jl
+++ b/src/coupling.jl
@@ -11,6 +11,8 @@ export
     CRescaleBarostat,
     MonteCarloBarostat
 
+default_seed() = rand(RandomDevice(), UInt64) # For stochastic couplers
+
 """
     apply_coupling!(system, coupling, simulator, neighbors=nothing, step_n=0;
                     n_threads=Threads.nthreads(), rng=Random.default_rng())
@@ -73,7 +75,7 @@ end
 needs_virial(c::ImmediateThermostat) = (truth = false, steps = Inf)
 
 @doc raw"""
-    VelocityRescaleThermostat(temperature, coupling_const; n_steps = 1, seed = 42)
+    VelocityRescaleThermostat(temperature, coupling_const; n_steps = 1, seed::UInt64 = default_seed())
 
 The stochastic velocity rescaling thermostat. See:
 
@@ -103,7 +105,7 @@ struct VelocityRescaleThermostat{T, C, N, S}
     seed::S
 end
 
-function VelocityRescaleThermostat(temperature, coupling_const; n_steps = 1, seed = 42)
+function VelocityRescaleThermostat(temperature, coupling_const; n_steps = 1, seed::UInt64 = default_seed())
     return VelocityRescaleThermostat(temperature, coupling_const, n_steps, seed)
 end
 
@@ -500,7 +502,7 @@ end
 
 function CRescaleBarostat(pressure::Union{PT, AbstractArray{PT}}, coupling_const;
                            coupling_type=:isotropic, compressibility=4.6e-5u"bar^-1",
-                           max_scale_frac=0.1, seed = 42, n_steps=1) where {PT}
+                           max_scale_frac=0.1, seed::UInt64 = default_seed(), n_steps=1) where {PT}
 
     if !(coupling_type ∈ (:isotropic, :semiisotropic, :anisotropic))
         throw(ArgumentError(ArgumentError("coupling_type must be :isotropic, :semiisotropic, or :anisotropic")))
--- a/src/energy.jl
+++ b/src/energy.jl
@@ -82,7 +82,10 @@ form, is defined as:
 ```
 where ``\bf{r_i}`` and ``\bf{f_i}`` are the position and force vectors,
 respectively, acting on the i``^{th}`` atom. In Molly.jl, we implement
-the [virial definition used in LAMMPS](https://docs.lammps.org/compute_stress_atom.html). 
+the [virial definition used in LAMMPS](https://docs.lammps.org/compute_stress_atom.html),
+and take into account pairwise and specific interactions, as well as the K-space
+contribution of the [`Ewald`](@ref) and [`PME`](@ref) methods, computed as indicated
+in the [original paper](https://doi.org/10.1063/1.470117).
 """
 function virial(sys)
     _, virial = forces(sys; needs_virial = true) # Force recomputation


ext/MollyCUDAExt.jl
```diff
@@ -515,7 +561,7 @@ function force_kernel!(
                 aϵ_j = CUDA.shfl_sync(0xFFFFFFFF, aϵ_j, laneid() + a, warpsize())
                 
                 atoms_j_shuffle = Atom(atype_j, aindex_j, amass_j, acharge_j, aσ_j, aϵ_j)
-                dr = vector(coords_j, coords_i, boundary)
+                dr = vector(coords_i, coords_j, boundary) # Sign flip (Giuseppe's kernel uses j, i) needed for virial, does not affect forces as only norm(dr) matters for them
```
jgreener64, line 564: Whilst none of the built-in interactions do, a user interaction can theoretically pay more attention to the i/j order. Ideally `dr` would stay as-is and the sign would be flipped where required for the virial.

src/energy.jl
```diff
@@ -23,29 +25,87 @@ end
...
+"""
+function kinetic_energy_tensor!(sys)
+    K = zero(sys.kin_tensor)  # same shape and units as target (energy)
+    fill!(K, zero(eltype(K)))
+    for (m, v) in zip(from_device(sys.masses), from_device(sys.velocities))
+        # m: mass per particle, v: velocity with units
+        K .+= m .* (v * transpose(v))  # units: energy
+    end
+    # Store the kinetic-energy tensor (½ Σ m v vᵀ), convertible to desired units
+    sys.kin_tensor .= 0.5 .* uconvert.(sys.energy_units, K)
```
jgreener64: `uconvert.(sys.energy_units, K) ./ 2` to avoid the type conversion for f32.

jgreener64: You could also just empty the `kin_tensor` and fill it directly to avoid allocating `K`.

src/force.jl
```diff
@@ -179,21 +191,36 @@ zero_forces(sys) = ustrip_vec.(zero(sys.coords)) .* sys.force_units
     forces(system, neighbors=find_neighbors(sys), step_n=0; n_threads=Threads.nthreads())
 
 Calculate the forces on all atoms in a system using the pairwise, specific and
-general interactions.
+general interactions. This call also populates the [`virial`](@ref) tensor of
+the system.
 """
-function forces(sys; n_threads::Integer=Threads.nthreads())
-    return forces(sys, find_neighbors(sys; n_threads=n_threads); n_threads=n_threads)
+function forces(sys; Virial::Bool = false, n_threads::Integer=Threads.nthreads())
```
jgreener64: For style, arguments not in the type domain should be lower case. Since they should also not clash with function names, you might want to rename this to `calc_virial`/`needs_virial` or similar.

src/interactions/ewald.jl
```diff
@@ -11,34 +11,55 @@ AtomsCalculators.@generate_interface function AtomsCalculators.potential_energy(
...
 
 AtomsCalculators.@generate_interface function AtomsCalculators.forces!(fs,
                                             sys,
                                             inter::AbstractEwald;
+                                            Virial = false,
                                             n_threads::Integer=Threads.nthreads(),
                                             kwargs...)
-    pe = ewald_pe_forces!(fs, sys, inter; n_threads=n_threads)
+    if Virial
+        virial = zeros(DefaultFloat, 3, 3) .* sys.energy_units
```
jgreener64: You can access the system float type from its type parameters, e.g. `System{<:Any, <:Any, T}`.

src/interactions/ewald.jl
```diff
@@ -831,31 +929,31 @@ end
 grad_safe_fft!( charge_grid, fft_plan ) = fft_plan  * charge_grid
 grad_safe_bfft!(charge_grid, bfft_plan) = bfft_plan * charge_grid
 
-function ewald_pe_forces!(Fs, inter::PME{T}, atoms, coords, boundary, force_units, energy_units,
-                          calculate_forces=true; n_threads::Integer=Threads.nthreads()) where T
+function ewald_pe_forces!(Fs, Vir, inter::PME{T}, atoms, coords, boundary, force_units, energy_units,
+                          ::Val{Virial}, calculate_forces=true; n_threads::Integer=Threads.nthreads()) where {T, Virial} # Here is the entry point, add Virial bool flag
```
jgreener64: Comments like this can be removed.

src/simulators.jl
```diff
@@ -683,12 +783,34 @@ end
         @warn "NoseHoover is not currently compatible with constraints, " *
               "constraints will be ignored"
     end
+    
+    if sim.coupling isa NoCoupling
```
jgreener64: This can move to a separate function.

Also it's probably worth checking that a smaller `vir.steps` is a factor of the current `needs_vir_steps` and throwing an error if not. Otherwise the two couplers could need the virial on different steps.

test/acemd.jl
```diff
@@ -27,7 +27,7 @@ using Test
 
     simulator = VelocityVerlet(
         dt=1.0u"fs",
-        coupling=AndersenThermostat(temp, 1.0u"ps"),
+        coupling=(AndersenThermostat(temp, 1.0u"ps"),),
```
jgreener64, line 30: We should still allow coupling to be a single struct too. You can use dispatch where required to do this, e.g.
```julia
f(c) = ...

function f(cs::Union{Tuple, NamedTuple})
    for c in cs
        f(c)
    end
end
```

src/coupling.jl
```diff
@@ -133,346 +224,532 @@ function apply_coupling!(sys, thermostat::BerendsenThermostat, sim, neighbors=no
...
+    x isa Number && unit(x) == NoUnits ? true : (uconvert(u"bar", x); true)
+catch
+    false
 end
 
-function apply_coupling!(sys, barostat::BerendsenBarostat, sim, neighbors=nothing,
-                         step_n::Integer=0; n_threads::Integer=Threads.nthreads(), kwargs...)
-    if !iszero(step_n % barostat.n_steps)
-        return false
+_isibar(x) = try
```
jgreener64: You can check `dimension` to avoid the `try`/`catch` here.

src/coupling.jl
```diff
@@ -133,346 +224,532 @@ function apply_coupling!(sys, thermostat::BerendsenThermostat, sim, neighbors=no
...
+        
+        if !all(_isibar.(compressibility))
+            throw(ArgumentError("semiisotropic: compressibility must have 1/pressure units"))
+        end
+
+
+        P_units = unit(pressure[1])
+        K_units = unit(compressibility[1])
+
+        px  = ustrip(uconvert(P_units, pressure[1]))
```
jgreener64: `ustrip(P_units, pressure[1])` does this in one call.

src/coupling.jl
```diff
@@ -133,346 +224,532 @@ function apply_coupling!(sys, thermostat::BerendsenThermostat, sim, neighbors=no
...
+                            κxy, κy,  κyz,
+                            κxz, κyz, κz) .* K_units
+
+        return BerendsenBarostat(P, coupling_const, coupling_type, κ, FT(max_scale_frac), n_steps)
+    end
+end
+
+function apply_coupling!(sys, barostat::BerendsenBarostat{PT, CT, ST, ICT, FT}, sim, neighbors=nothing,
+                         step_n::Integer=0; n_threads::Integer=Threads.nthreads(), kwargs...) where {PT, CT, ST, ICT, FT} # Avoids uconvert calls in method
+    if step_n % barostat.n_steps != 0; return false; end
```
jgreener64: Just style but put on multiple lines or use `step_n % barostat.n_steps != 0 && return false`.

src/coupling.jl
```diff
@@ -133,346 +224,532 @@ function apply_coupling!(sys, thermostat::BerendsenThermostat, sim, neighbors=no
...
+    
+    if barostat.coupling_type == :isotropic
+        for d in 1:D
+            α = (barostat.compressibility[d,d] * dt) / (D*τp)
+            s = 1 + α * (Pavg - barostat.pressure[d,d])
+            μ[d,d] = clamp(s, 1 - barostat.max_scale_frac, 1 + barostat.max_scale_frac)
+        end
+
+    elseif barostat.coupling_type == :semiisotropic
+        @assert D == 3 "Cannot apply semi-isotropic in 2D."
```
jgreener64: Prefer `ArgumentError` over `@assert` as assertions can be turned off at higher optimisation levels.

src/coupling.jl
```diff
@@ -483,156 +760,328 @@ does not actively control the temperature.
...
+                recompute_forces = true
+                barostat.n_accepted += 1
+            else
+                sys.coords .= old_coords
+                sys.boundary = old_boundary
+            end
+            barostat.n_attempted += 1
+
+            # Modify size of volume change to keep accept/reject ratio roughly equal
+            if barostat.n_attempted >= 10
```
jgreener64: Any way to move this last part out of the branch to avoid repetition?

src/spatial.jl
```diff
@@ -956,57 +1108,120 @@ function molecule_centers(coords::AbstractGPUArray,
...
+    
+    μinv = inv(μ)
+
     if ignore_molecules || isnothing(sys.topology)
-        sys.boundary = scale_boundary(sys.boundary, scale_factor)
-        sys.coords .= scale_vec.(sys.coords, (scale_factor,))
-    elseif sys.boundary isa TriclinicBoundary
-        error("scaling coordinates by molecule is not compatible with a TriclinicBoundary")
+        # box
+        B  = SMatrix{D,D}(ustrip(boxmatrix(sys.boundary)))
```
jgreener64: This gives a warning in the tests:
```
`┌ Warning: `ustrip(A::AbstractArray{T}) where T <: Number` is deprecated, use `ustrip.(A)` instead.`
```

src/coupling.jl
```diff
@@ -62,12 +64,99 @@ struct ImmediateThermostat{T}
...
+
+"""
+struct VelocityRescaleThermostat{T, C, N, S}
+    temperature::T
+    coupling_const::C
+    n_steps::N
+    seed::S
+end
+
+function VelocityRescaleThermostat(temperature, coupling_const; n_steps = 1, seed = 42)
```
jgreener64: Choose a random seed by default, also for the barostat.

docs/src/documentation.md
```diff
@@ -1098,6 +1098,16 @@ end
...
+Note that as some of the coupling methods need the [`virial`](@ref) to properly work, any newly defined custom coupler
+should be accompanied by a function that tells Molly if it also needs the virial tensor and the amount of simulation
+steps between integrations of the coupler.
+```julia
+# In case you need the virial
+needs_virial(c::MyCoupler) = (truth = true, steps = c.n_steps)
+# In case you do NOT need the virial
+needs_virial(c::MyCoupler) = (truth = false, steps = Inf)
+```
+
```
jgreener64, line 1112: Briefly mention that the virial takes into account pairwise and specific forces here, maybe with a link to the LAMMPS docs or similar, and PME with a link to the relevant paper.

jgreener64 (approved):
Great, click merge after making the two further changes.

PR #220, opened 2025-10-17
DOCS: Documentation update to provide example of KIM Portable Models in Molly.jl

ipcamit:
Added example for running `KIMPortableModels` in Molly.jl.


ipcamit:
Trying to rename the package! To be more in line with other KIM offerings. Will resubmit after 3-4 days once new package is approved.

PR #221, opened 2025-10-27
Example for using KIM_API.jl package

ipcamit:
This PR adds the example for using KIM_API.jl package for KIM-API based general interactions. It is a rename of KIMPortableModels previously.


jgreener64:
Great, thanks for this.

PR #222, opened 2025-10-28
Added soft-core potentials (Beutler et al. & Gapsys et al.)

ERovers:
I have added four functions of which two replace the previous soft-core potentials (LennardJonesSoftCore and CoulombSoftCore). 

- Added the soft-core potentials described in Beutler et al. (1994) in Chem. Phys. Lett. and called these functions LennardJonesSoftCoreBeutler and CoulombSoftCoreBeutler, these were similar to the previous soft-core potentials, but I reversed the λ-scaling so that for λ =1, the interaction is fully turned on and for λ=0, the interaction is turned off.
- Added the soft-core potentials described in Gapsys et al. (2012) in JCTC and called these functions LennardJonesSoftCoreGapsys and CoulombSoftCoreGapsys.
- Added a new atom type 'Atom_L' in preparation for atoms that will be controlled with λ. (Essentially the same as the 'Atom' type but with an added λ variable). However, it is not being used yet.
- Updated the test/interactions.jl to include tests for these functions.
- Updated the test/simulation.jl to run with LennardJonesSoftCoreBeutler in the HREMD.
- Updated the soft-core potentials section in examples.md to showcase the difference between these two types of soft-core potentials and the effect of λ on both.
- Updated documentations.md where soft-core potentials were mentioned.
- Updated cutoffs.jl by exporting the force_cutoff function to enable plotting the magnitude of the forces. While it is not used in the examples.md, it was helpful when I was plotting the forces generally.

Important note: These interactions are for appearing and disappearing atoms only, not for transformations between different atoms (e.g. a alchemical transformation from carbon to hydrogen atom).


src/interactions/lennard_jones.jl
```diff
@@ -263,9 +263,152 @@ end
...
+    else
+        return fdr
+    end
+end
+
+function pairwise_force(inter::LennardJonesSoftCoreGapsys, r, (dr, C12, C6))
+    R = inter.α*sqrt(cbrt(((26/7)*(C12/C6)*(1-inter.λ))))
+    invR = 1/R
+    if r >= R
+        return (((12*C12)/r^13)-((6*C6)/r^7))
```
jgreener64: To save computation you can set `r6 = r^6` then use `r6 * r` and `r6 * r6 * r` as appropriate. This also applies elsewhere.

src/types.jl
```diff
@@ -277,6 +277,32 @@ function inject_atom(at, at_data, params_dic)
     )
 end
 
+"""
+    Atom_L(; <keyword arguments>)
```
jgreener64: For now don't add this to Molly, as we may decide to have lambda in the main atom type.

src/cutoffs.jl
```diff
@@ -5,7 +5,8 @@ export
     DistanceCutoff,
     ShiftedPotentialCutoff,
     ShiftedForceCutoff,
-    CubicSplineCutoff
+    CubicSplineCutoff,
+    force_cutoff
```
jgreener64: Don't export, you can access with `Molly.force_cutoff` or `using Molly: force_cutoff`.

src/interactions/coulomb.jl
```diff
@@ -108,59 +109,73 @@ function pairwise_pe(::Coulomb, r, (ke, qi, qj))
 end
 
 @doc raw"""
-    CoulombSoftCore(; cutoff, α, λ, p, use_neighbors, σ_mixing, weight_special, coulomb_const)
+    CoulombSoftCoreBeutler(; cutoff, α, λ, use_neighbors, σ_mixing, weight_special, coulomb_const)
 
-The Coulomb electrostatic interaction between two atoms with a soft core.
+The Coulomb electrostatic interaction between two atoms with a soft core, used for appearing and disappearing of atoms based on the potential described in Beutler et al. 1994 (Chem. Phys. Lett.).
```
jgreener64: Link to the DOI of the paper, like in src/simulators.jl line 358. Also link where it is mentioned in the docs.

PR #223, opened 2025-10-31
Setup branch - More robust system building from force field files

allexandre97:
# Following OpenMM philosophy

Until now, Molly used  the chemfiles Julia bindings to infer the system topology from structure files. In general, this worked well but had some quirks when non-standard residues were present, for terminal residues and for water molecules. 

In this PR this was changed such that Chemfiles is only used to do a preliminary read of the structure file to obtain some basics like the atom names and positions. From this, we roughly follow the pipeline implemented into OpenMM to build the system topology.

## System canonicalization

OpenMM has a dictionary of variants of residue and atom names commonly found on PDB files. After the first Chemfiles parsing of the structure file, we query the determined names against this dictionary to use the canonical naming. This dictionary is can be found in `src/data/force_fields/pdbNames.xml`. This dictionary can be extended at the time the user defines the forcefield, by passing an optional keyword argument to `MolecularForcefield` pointing to their custom renaming definitions. This should be an .xml file with an equivalent layout as the one mentioned above.

## Bonds determination

Once the system has been canonically named, the residues are queried against a dictionary of bonds present in standard residues. These residues include all of the aminoacids, nucleic acids and water. The dictionary is built in such a way that it details how each residue can be linked to its previous and next residues, and includes special naming for the atoms that are present when they are in a terminal position. The dictionary can be found in `src/data/force_fields/residues.xml`. This dictionary can also be extended at the time the user defines the forcefield, in the same way as before.

## Template matching

With the bonds in place, one can then iterate over all of the residues present in the system, and compare their molecular graphs with the ones defined in the forcefield. When a match is found, the corresponding parameters are assigned to the atoms, bonds, angles, etc.


src/residues.jl
```diff
@@ -0,0 +1,720 @@
+# --- Elements dictionary ---
+const ELEMENT_SYMBOLS = Dict{String,Symbol}(
```
jgreener64: `Symbol(string)` should work.

src/residues.jl
```diff
@@ -0,0 +1,720 @@
...
+    "In"=>:In,"Sn"=>:Sn,"Sb"=>:Sb,"Te"=>:Te,"I"=>:I,"Xe"=>:Xe,
+    "Cs"=>:Cs,"Ba"=>:Ba,"La"=>:La,"Ce"=>:Ce,"Pr"=>:Pr,"Nd"=>:Nd,"Pm"=>:Pm,"Sm"=>:Sm,"Eu"=>:Eu,"Gd"=>:Gd,"Tb"=>:Tb,"Dy"=>:Dy,"Ho"=>:Ho,"Er"=>:Er,"Tm"=>:Tm,"Yb"=>:Yb,"Lu"=>:Lu,
+    "Hf"=>:Hf,"Ta"=>:Ta,"W"=>:W,"Re"=>:Re,"Os"=>:Os,"Ir"=>:Ir,"Pt"=>:Pt,"Au"=>:Au,"Hg"=>:Hg,
+    "Tl"=>:Tl,"Pb"=>:Pb,"Bi"=>:Bi,"Po"=>:Po,"At"=>:At,"Rn"=>:Rn,
+    "Fr"=>:Fr,"Ra"=>:Ra,"Ac"=>:Ac,"Th"=>:Th,"Pa"=>:Pa,"U"=>:U,"Np"=>:Np,"Pu"=>:Pu,"Am"=>:Am,"Cm"=>:Cm,"Bk"=>:Bk,"Cf"=>:Cf,"Es"=>:Es,"Fm"=>:Fm,"Md"=>:Md,"No"=>:No,"Lr"=>:Lr,
+    "Rf"=>:Rf,"Db"=>:Db,"Sg"=>:Sg,"Bh"=>:Bh,"Hs"=>:Hs,"Mt"=>:Mt,"Ds"=>:Ds,"Rg"=>:Rg,"Cn"=>:Cn,"Nh"=>:Nh,"Fl"=>:Fl,"Mc"=>:Mc,"Lv"=>:Lv,"Ts"=>:Ts,"Og"=>:Og,
+    "?"=>:X
+)
+
+const PERIODIC_TABLE = Symbol[
```
jgreener64: This may be available in the PeriodicTable package, which we import.

src/residues.jl
```diff
@@ -0,0 +1,720 @@
...
+
+            if haskey(standard_bonds, res_name)
+                for bond in standard_bonds[res_name]
+                    external = false
+                    if startswith(bond[1], "-") && i > 1
+                        external = true
+                        fromResidue = i-1
+                        fromAtom = bond[1][2:end]
+                        
+                        ext_ind = findfirst(x->x==fromAtom, resids[fromResidue].atom_names)
```
jgreener64: I think `isequal(fromAtom)` can replace `x->x==fromAtom`.

src/setup.jl
```diff
@@ -485,371 +583,324 @@ function System(coord_file::AbstractString,
...
                 implicit_solvent=:none,
                 kappa=0.0u"nm^-1",
-                rename_terminal_res::Bool=true,
-                grad_safe::Bool=false) where AT <: AbstractArray
-    if dist_buffer < zero(dist_buffer)
-        throw(ArgumentError("dist_buffer ($dist_buffer) should not be less than zero"))
-    end
+                grad_safe::Bool=false) where {AT<:AbstractArray}
+
+    dist_buffer < zero(dist_buffer) && throw(ArgumentError("dist_buffer ($dist_buffer) should not be less than zero"))
```
jgreener64: Generally keep lines to 100 character.

src/residues.jl
```diff
@@ -0,0 +1,720 @@
...
+            end
+        end
+    end
+
+    return sort(bonds)
+
+end
+
+# Builds disulfide bonds given some geometric criteria
+function create_disulfide_bonds(coords, boundary, canon_system, bonds)
```
jgreener64, line 276: Whether to do this could be a keyword argument, default `true`, during setup.

src/setup.jl
```diff
@@ -485,371 +583,324 @@ function System(coord_file::AbstractString,
...
-              "($(2 * dist_cutoff)), this can lead to unphysical simulations"
+    if center_coords
+        coords = coords .- (mean(coords),) .+ (box_center(boundary_used),)
     end
+    
+    canonical_system = canonicalize_system(top, resname_replacements, atomname_replacements)
+
+    top_bonds = create_bonds(canonical_system, standard_bonds)
+    top_bonds = create_disulfide_bonds(coords, boundary_used, canonical_system, top_bonds)
+    top_bonds = read_connect_bonds(coord_file, top_bonds, canonical_system)
```
jgreener64: Reading the CONECT records works for PDB files but not for other formats. Is it possible to extract extra bonds from the Chemfiles topology (which should include the CONECT bonds)?

jgreener64 (commented):
Looks great and should allow much more user-friendly setup.

It might be good to add a couple of new tests:
- A ligand with bonds defined as CONECT records and the custom template file provided.
- A non-PDB format like MOL2.

Some of the docs will need to change too.

I can try to dig up a few systems I have so we can check the forces/energies match OpenMM.

test/basic.jl
```diff
@@ -239,6 +239,82 @@ end
...
+@testset "System Setup" begin
+
+    FT = Float64
+    AT = Array
+
+    dat_dir = joinpath(data_dir, "OpenMM_Refs/")
+
+    ff = MolecularForceField(
+        FT,
+        joinpath.(ff_dir, ["ff99SBildn.xml", "tip3p_standard.xml", "his.xml"])...;
```
jgreener64: I think his.xml can now be removed completely from the tests and deleted.

test/basic.jl
```diff
@@ -239,6 +239,82 @@ end
...
+    for name in names
+
+        dat_file = joinpath(dat_dir, "$(name).dat")
+        pdb_file = joinpath(dat_dir, "$(name).pdb")
+
+        sys = System(pdb_file, ff;
+             array_type = AT,
+             nonbonded_method = :pme,
+             approximate_pme  = false,
+             disulfide_bonds  = true)
```
jgreener64, line 301: Might be worth testing the `disulfide_bonds=false` path on one of the structures and check that the number of bonds is as expected.

src/setup.jl
```diff
@@ -460,9 +650,9 @@ Gromacs files.
     `:none`, `:obc1`, `:obc2` and `:gbn2`.
 - `kappa=0.0u"nm^-1"`: the kappa value for the implicit solvent model if one
     is used.
-- `rename_terminal_res=true`: whether to rename the first and last residues
-    to match the appropriate atom templates, for example the first (N-terminal)
-    residue could be changed from "MET" to "NMET".
+- `disulfide_bonds=false`: wether or not to look for disulfide bonds between CYS
```
jgreener64: Default `true` is probably safer.

test/basic.jl
```diff
@@ -239,6 +239,82 @@ end
...
+
+    ff_custom = MolecularForceField(
+        joinpath.(ff_dir, ["ff99SBildn.xml", "tip3p_standard.xml", "his.xml", "gaff.xml", "imatinib.xml", "imatinib_frcmod.xml"])...;
+        units=true,
+        custom_residue_templates = joinpath(data_dir, "imatinib_topo.xml")
+    )
+
+    boundary = CubicBoundary(Inf*u"nm")
+    sys_mol2 = System(joinpath(data_dir, "imatinib.mol2"), ff; boundary = boundary)
+    sys_pdb  = System(joinpath(data_dir, "imatinib.pdb"),  ff_custom; boundary = boundary)
```
jgreener64: Also test the PDB with CONECT records so we know we can read those.

jgreener64: If there is a setup path that errors due to e.g. missing bond definitions you could test that with `@test_throws`.

PR #229, opened 2025-11-17
Adding bias potentials and cv functions

theaschulze:
Added functionality to run biased simulations. 

A CV function and a bias potential are individually defined and combined in the struct BiasPotential, which is passed to `AtomsCalculators.potential_energy` and `AtomsCalculators.forces!` as a general interaction type. 

Most code was added to `src/bias`, which contains the two files `bias.jl` and `cv.jl`. In `bias.jl`, methods to calculate the potential energy and forces that result from the added bias are defined, while `cv.jl` contains methods to calculate values of a number of CVs (distance, RMSD, Rg). The force calculation is split into calculating the derivative of the bias potential with respect to the CV and calculating the gradient of the CV with respect to the coordinates. The latter is by default done with autodiff, and `cv_gradient` is hence defined in `MollyEnzymeExt.jl`.

The function `kabsch` was added to `analysis.jl`. `kabsch` is needed in the calculate_cv method for the RMSD CV. 

All related tests are in `test/bias.jl` and new documentation under  `Biased simulations` in `documentation.md`.

### DIFF
--- a/ext/MollyEnzymeExt.jl
+++ b/ext/MollyEnzymeExt.jl
@@ -78,9 +78,7 @@ function Molly.cv_gradient(cv_type, coords, atoms, boundary, velocities) # this 
     u = only(unit_arr)
 
     # if coords (and hence d_coords) are unitful, correct units
-    if unit(d_coords[1][1]) == u"nm"
-        d_coords = d_coords .* u ./ u"nm"^2
-    end
+    d_coords = d_coords .* u ./ unit(d_coords[1][1])^2
         
     return d_coords, cv_val_ustrip * u
 end
--- a/src/analysis.jl
+++ b/src/analysis.jl
@@ -50,8 +50,8 @@ function kabsch(coords_1::AbstractArray{SVector{D, T}},
     trans_1 = mean(coords_1)
     trans_2 = mean(coords_2)
 
-    p = Molly.from_device(reshape(reinterpret(T, coords_1), D, n_atoms)) .- repeat(reinterpret(T, trans_1), 1, n_atoms)
-    q = Molly.from_device(reshape(reinterpret(T, coords_2), D, n_atoms)) .- repeat(reinterpret(T, trans_2), 1, n_atoms)
+    p = from_device(reshape(reinterpret(T, coords_1), D, n_atoms)) .- repeat(reinterpret(T, trans_1), 1, n_atoms)
+    q = from_device(reshape(reinterpret(T, coords_2), D, n_atoms)) .- repeat(reinterpret(T, trans_2), 1, n_atoms)
     
     cov = p * transpose(q)
     svd_res = svd(ustrip.(cov))
@@ -68,15 +68,13 @@ function kabsch(coords_1::AbstractArray{SVector{D, T}},
     return p_rot_reshaped, q_reshaped # return p centered and rotated, q centered
 end
 
-"""
-    ref_kabsch(coords_1, coords_2)
+#=  ref_kabsch(coords_1, coords_2)
 
 Wrapper function to return only the translated and rotated coordinates
 of coords_1 after superimposition of coords_1 and coords_2 by the Kabsch algorithm.
 
 Assumes the coordinates do not cross the bounding box, i.e. all
-coordinates in each set correspond to the same periodic image.
-"""
+coordinates in each set correspond to the same periodic image.=#
 function ref_kabsch(coords_1::AbstractArray{SVector{D, T}},
                     coords_2::AbstractArray{SVector{D, T}}) where {D, T}
     p_rot, _ = kabsch(coords_1, coords_2)
--- a/src/bias/bias.jl
+++ b/src/bias/bias.jl
@@ -129,12 +129,12 @@ function AtomsCalculators.potential_energy(
     )
 
     if bias.cv_type.correction == :pbc
-        coords = Molly.unwrap_molecules(sys) 
+        coords = unwrap_molecules(sys) 
     else
         coords = sys.coords 
     end
 
-    cv_sim = calculate_cv(bias.cv_type, Molly.from_device(coords), Molly.from_device(sys.atoms), sys.boundary, Molly.from_device(sys.velocities); kwargs...) 
+    cv_sim = calculate_cv(bias.cv_type, from_device(coords), from_device(sys.atoms), sys.boundary, from_device(sys.velocities); kwargs...) 
     energy = potential_energy(bias.bias_type, cv_sim; kwargs...) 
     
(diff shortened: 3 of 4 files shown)


src/bias/cv.jl
```diff
@@ -0,0 +1,119 @@
+# Calculate collective variables
+
+export
+    CalcDist,
+    CalcRg,
+    CalcRMSD,
+    calculate_cv
+    #calculate_cv_ustrip!, # dont export
```
jgreener64: Remove these lines.

src/bias/cv.jl
```diff
@@ -0,0 +1,119 @@
...
+    atom_idx_2::Int
+    dist_type::Symbol=:pbc_dist   
+    correction::Symbol=:wrap
+end
+
+function calculate_cv(cv::CalcDist, coords, atoms, boundary, args...; kwargs...) 
+    coords_1 = coords[cv.atom_idx_1]
+    coords_2 = coords[cv.atom_idx_2]
+    if cv.dist_type == :pbc_dist
+        dist_val = distances([coords_1,coords_2], boundary)[2]
```
jgreener64: `norm(vector(coords_1, coords_2, boundary))` is faster.

ext/MollyEnzymeExt.jl
```diff
@@ -24,6 +24,7 @@ EnzymeRules.inactive(::typeof(visualize), args...) = nothing
 EnzymeRules.inactive(::typeof(place_atoms), args...) = nothing
 EnzymeRules.inactive(::typeof(place_diatomics), args...) = nothing
 EnzymeRules.inactive(::typeof(read_frame!), args...) = nothing
+EnzymeRules.inactive(::typeof(kabsch), args...) = nothing
```
jgreener64: In general I don't think this should be inactive, but your use of it may be okay to be inactive in which case you could have a wrapper function (`ref_kabsch`?) that is marked inactive.

src/bias/bias.jl
```diff
@@ -0,0 +1,163 @@
...
+```
+where $s$ and $s_{target}$ are system and target CV values, respectively. 
+"""
+struct SquareBias{K, CVT}
+    k::K
+    cv_target::CVT             
+end
+
+function potential_energy(sb::SquareBias, cv_sim; kwargs...)  
+    pe = 0.5 * sb.k * (cv_sim - sb.cv_target)^2
```
jgreener64: `0.5` promotes `Float32` to `Float64`, here you could use ` / 2`.

src/analysis.jl
```diff
@@ -31,6 +32,42 @@ periodic boundary conditions.
 """
 distances(coords, boundary) = norm.(displacements(coords, boundary))
 
+"""
+    kabsch(coords_1, coords_2)
```
jgreener64, line 36: You could delete the corresponding code from `rmsd` and call this function to remove duplication.

src/bias/cv.jl
```diff
@@ -0,0 +1,119 @@
...
+Bias the distance between two atoms in a system. 
+
+# Arguments
+- `atom_idx_1::Int`: index of the first atom.
+- `atom_idx_2::Int`: index of the second atom.
+- `dist_type::Symbol=:pbc_dist`: distances are evaluated taking periodic boundary conditions into account. Should be `:dist` if PBCs should not be considered.
+- `correction::Symbol=:wrap`: the system is wrapped in the unit cell. Should be `:pbc` to unwrap molecules prior to the distance calculation. 
+"""
+@kwdef struct CalcDist   
+    atom_idx_1::Int
```
jgreener64: Would it be possible to allow multiple indices here and calculate the distance between the centers of mass?

ext/MollyEnzymeExt.jl
```diff
@@ -56,4 +57,31 @@ function EnzymeRules.reverse(config, ::Const{typeof(Molly.grad_safe_bfft!)}, dre
...
+        Duplicated(coords, d_coords),  
+        Const(atoms), 
+        Const(boundary), 
+        Const(velocities)
+    )
+    
+    u = only(unit_arr)
+
+    if unit(d_coords[1][1]) == u"nm"
+        d_coords = d_coords .* 1u"nm^-1"
```
jgreener64: I don't think this should be required.

src/bias/bias.jl
```diff
@@ -0,0 +1,164 @@
...
+    )
+
+    if bias.cv_type.correction == :pbc
+        coords = Molly.unwrap_molecules(sys) 
+    else
+        coords = sys.coords 
+    end
+
+    # gradient of cv with respect to coordinates
+    d_coords, cv_sim = Molly.cv_gradient(bias.cv_type, Molly.from_device(coords), Molly.from_device(sys.atoms), sys.boundary, Molly.from_device(sys.velocities))
```
jgreener64: The `Molly.` prefix is not required within the package (just the extensions and tests).

src/bias/cv.jl
```diff
@@ -0,0 +1,253 @@
...
+
+@doc raw"""
+    CalcMaxDist
+
+Bias the maximum distance between two groups of atoms. 
+
+# Arguments
+- `calc_type::Symbol`: Should be `:closest` to calculate distances between closest periodic images and `:raw` otherwise. 
+"""
+struct CalcMaxDist
```
jgreener64: Give these four structs the default of `:closest` with `@kwdef`.

ext/MollyEnzymeExt.jl
```diff
@@ -56,4 +57,32 @@ function EnzymeRules.reverse(config, ::Const{typeof(Molly.grad_safe_bfft!)}, dre
...
+        Duplicated(coords, d_coords),  
+        Const(atoms), 
+        Const(boundary), 
+        Const(velocities)
+    )
+    
+    u = only(unit_arr)
+
+    # if coords (and hence d_coords) are unitful, correct units
+    if unit(d_coords[1][1]) == u"nm"
```
jgreener64: I think you can replace these 3 lines with
```julia
d_coords = d_coords .* u ./ unit(d_coords[1][1])^2
```
which works for no units and non nm cases.

src/bias/cv.jl
```diff
@@ -0,0 +1,253 @@
...
+    CalcMinDist,
+    CalcMaxDist,
+    CalcComDist,
+    CalcSingleDist,
+    CalcDist,
+    CalcRg,
+    CalcRMSD,
+    calculate_cv
+
+@doc raw"""
```
jgreener64: Have these as comments rather than docstrings for non-exported functions or the docs build won't pass.

PR #232, opened 2025-11-26
Add missing λ-factor in force and potential for LennardJonesSoftCoreBeutler

Immi000:
This adds the missing lambda-factor referred to [here](https://github.com/JuliaMolSim/Molly.jl/pull/227#issuecomment-3517285284).

### DIFF
--- a/src/interactions/coulomb.jl
+++ b/src/interactions/coulomb.jl
@@ -321,7 +321,7 @@ end
     cutoff = inter.cutoff
     ke = inter.coulomb_const
     qi, qj = atom_i.charge, atom_j.charge
-    params = (ke, qi, qj, inter.σQ, inter.σ6_fac)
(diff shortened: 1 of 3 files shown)


jgreener64:
Thanks for this. Would you mind updating the docstring and making the analogous changes for `CoulombSoftCoreBeutler` too?

I guess the Gapsys soft core variants should also have this change?

jgreener64:
Thanks for the changes. Having a look it seems that OpenFE also applies the same λ weighting to the Gapsys variants (https://github.com/OpenFreeEnergy/openfe/blob/013b4b3213b8329339607296d371df8b8e207b92/openfe/protocols/openmm_rfe/_rfe_utils/relative.py#L810), so I think those should change as well.

jgreener64:
Great, thanks.

Issue #233, opened 2025-12-04
Comprehension with Enzyme can cause issues in reverse mode


Immi000:
I have recently run into an issue with differentiating simulations in reverse mode which involve a barostat, which kept failing with incomprehensible compilation errors:

<img width="987" height="257" alt="Image" src="https://github.com/user-attachments/assets/8fe5edc4-b050-4fd4-b2b9-1bb46a734b0c" />

Upon some closer inspection I found [this line](https://github.com/JuliaMolSim/Molly.jl/blob/9451266cdabfec506653767561daef46aa870e03/src/spatial.jl#L1149) in the `scale_coords!`-function. I am not exactly sure what goes wrong there, and maybe Enzyme.jl is still lacking some feature(s) in this regard, but in my experience comprehensions and reverse mode AD don't mix well. And indeed, replacing it with

```
scaled_coords = from_device(sys.coords)
scaled_coords .*= μ
sys.coords .= to_device(scaled_coords, AT)
```

fixed the issue.

The same thing will probably happen for the [scaling of the coordinates](https://github.com/JuliaMolSim/Molly.jl/blob/9451266cdabfec506653767561daef46aa870e03/src/spatial.jl#L1152) as well.

I know that my proposed fix above is not ideal as it copies the coordinates for each scaling and I am not sure why the scaling needs to happen off the device, but it does at least fix differentiability. Do you maybe have a better idea how this can be solved?

PR #234, opened 2025-12-08
Split MC-barostat implementation

Immi000:
This PR splits the implementation of the isotropic, semiisotropic and anisotropic Monte-Carlo-Barostat into separate functions. This should improve the compatibility with Enzyme AD as noted in [#233](https://github.com/JuliaMolSim/Molly.jl/issues/233).


jgreener64:
Great, thanks for this.

PR #238, opened 2026-01-25
test: added test for analysis.jl

sivasathyaseeelan:
This PR adds test for src/analysis.jl


test/analysis.jl
```diff
@@ -0,0 +1,67 @@
+using Molly
```
jgreener64: These imports shouldn't be required.

sivasathyaseeelan: I have refactored the imports!

PR #250, opened 2026-03-26
LINCS constraint algorithm

Immi000:
I have a proposal implementation of the [Linear Constraint Solver algorithm (LINCS)](https://doi.org/10.1002/(SICI)1096-987X(199709)18:12%3C1463::AID-JCC4%3E3.0.CO;2-H) as an alternative to RATTLE/SHAKE. I mostly used the [Gromacs implementation](https://manual.gromacs.org/current/reference-manual/algorithms/constraint-algorithms.html#lincs) as a reference and tried to match the style and interface of Molly's [shake.jl](https://github.com/JuliaMolSim/Molly.jl/blob/master/src/constraints/shake.jl) implementation where it made sense. I'll be happy to receive comments/feedback and if wanted I can also add some text on how to use it to the docs.

Usage is pretty much the same as with SHAKE: You create a `LINCS` object which needs an array of masses, distance constraints and angle constraints (and optionally velocity and distance tolerances) and pass it to the system constructor as a tuple in the `constraints`-argument. It supports the two Gromacs arguments `nrec` for the expansion order of the coupling matrix inversion approximation and `niter` for the number of iterations for the corrections of the rotational lengthening. The implementation is GPU compatible (tested with CUDA).

My main motivation for this implementation was the support for ring topologies, which as far as I could tell are not supported with SHAKE.


test/constraints.jl
```diff
@@ -331,7 +331,7 @@ end
             random_velocities!(sys, temp)
 
             simulate!(sys, simulator, 20)
-            @time simulate!(sys, simulator, 1000)
+            simulate!(sys, simulator, 1000)
```
jgreener64: If `LINCS` is added to setup it can be added to the protein test here.

src/constraints/lincs.jl
```diff
@@ -0,0 +1,655 @@
...
+    coords_a1 = sys.coords[data.atom1]
+    coords_a2 = sys.coords[data.atom2]
+    target_lengths = data.lengths .* unit_len
+    dr_norms = norm.(vector.(coords_a1, coords_a2, (sys.boundary,)))
+    max_err = maximum(ustrip.(abs.(dr_norms .- target_lengths)))
+
+    return max_err < ustrip(ca.dist_tolerance)
+end
+
+function check_velocity_constraints(sys::System{<:Any, <:Any, FT}, ca::LINCS) where FT
```
jgreener64, line 894: This doesn't appear to be used in the tests currently. Could it be added where `check_position_constraints` is called?

test/constraints.jl
```diff
@@ -354,3 +354,665 @@ end
...
+             for j in 1:n_atoms]
+    atom_masses = [atom_mass for _ in 1:n_atoms]
+
+    cons = LINCS(masses=atom_masses, dist_tolerance=T(1e-4)u"Å", vel_tolerance=T(1e-4)u"Å^2 * ps^-1",
+                 dist_constraints=dist_constraints)
+
+    boundary = CubicBoundary(T(200.0)u"Å")
+    lj = LennardJones(cutoff=ShiftedPotentialCutoff(r_cut), use_neighbors=true)
+
+    for AT in array_list
```
jgreener64: `array_list[2:end]`

test/constraints.jl
```diff
@@ -354,3 +354,665 @@ end
...
+        AT === Array && continue
+
+        neighbor_finder = GPUNeighborFinder(
```
jgreener64: You should add a check on `Molly.uses_gpu_neighbor_finder(AT)` like in test/basic.jl line 195.

test/constraints.jl
```diff
@@ -354,3 +354,665 @@ end
...
+            SVector(-0.4, -0.2, -0.3),
+        ]
+        dt = 0.002
+        xp = x .+ v .* dt
+        Molly.lincs_apply!(xp, x, data, ws, nothing)
+
+        @test lincs_check_constraints(xp, data; atol=1e-6)
+    end
+
+    @testset "zero displacement is no-op" begin
```
jgreener64: This is just preference but I find these inner `@testset` labels make the test output harder to read.

jgreener64 (commented):
Looks great. Would be nice to have as an option during setup from a file, you could add a `constraint_algorithm` option which can take `SHAKE_RATTLE` or `LINCS`. If `LINCS` is faster and supports `constraints=:hbonds` and `rigid_water=true` then it could be the default.

Would be interesting to compare the performance to SHAKE/RATTLE, e.g. for the 6mrr system on GPU with the above options.

You can also mention `LINCS` in the constraints section of the docs. In particular, it would be good to state which types of constraint clusters are compatible with LINCS.

src/constraints/lincs.jl
```diff
@@ -0,0 +1,994 @@
...
+    return LincsCouplingDense(coupled_indices, coupled_coef_arr, n_coupled_arr, max_coupled)
+end
+
+# --- Core algorithm ---
+
+@inline function lincs_bond_vector(coords, a1, a2, boundary)
+    return ustrip.(vector(coords[a2], coords[a1], boundary))
+end
+
+@inline function lincs_bond_vector(coords, a1, a2, ::Nothing)
```
jgreener64: Why is the no-PBC case required?

test/constraints.jl
```diff
@@ -331,7 +336,7 @@ end
             random_velocities!(sys, temp)
 
             simulate!(sys, simulator, 20)
-            @time simulate!(sys, simulator, 1000)
+            simulate!(sys, simulator, 1000)
 
             @test check_position_constraints(sys)
```
jgreener64: Add a `LINCS` version of the protein test.

docs/src/documentation.md
```diff
@@ -1557,6 +1571,20 @@ shake = SHAKE_RATTLE(
...
+Alternatively, using LINCS:
+```julia
+masses = [1.0u"g/mol", 1.0u"g/mol", 1.0u"g/mol", 1.0u"g/mol", 1.0u"g/mol", 1.0u"g/mol"]
+
+lincs = LINCS(
+    masses=masses,
+    dist_constraints=dist_constraints,
+    angle_constraints=angle_constraints,
+)
+# LINCS with 2 distance and 1 angle constraints (nrec=4, niter=1) (imlpicitly 5 distance constraints)
```
jgreener64: Typo.

src/setup.jl
```diff
@@ -437,7 +437,8 @@ function System(coord_file::AbstractString,
                 disulfide_bonds=true,
                 grad_safe::Bool=false,
                 strictness=default_strictness(),
-                force_separate_lj14=false) where {AT <: AbstractArray}
+                force_separate_lj14=false,
+                constraint_algorithm=LINCS) where {AT <: AbstractArray}
```
jgreener64, line 468: Add this to the docstring.

src/setup.jl
```diff
@@ -1325,6 +1329,68 @@ function exchange_constraints(T, bonds_all, angles_all, atoms_data, constraints_
...
+    end
+
+    if length(dist_constraints) > 0 || length(angle_constraints) > 0
+        lincs = LINCS(
+            masses=masses,
+            dist_tolerance=add_units(T(1e-6), u"nm", units),
+            vel_tolerance=add_units(T(1e-6), u"nm^2 * ps^-1", units),
+            dist_constraints=[dist_constraints...],
+            angle_constraints=[angle_constraints...],
+            nrec=6,
```
jgreener64: Is there a reason this is `nrec=6` and `niter=6` when the default elsewhere is different?

src/setup.jl
```diff
@@ -1325,6 +1329,68 @@ function exchange_constraints(T, bonds_all, angles_all, atoms_data, constraints_
     return constraints, bonds, angles
 end
 
+function exchange_constraints(T, bonds_all, angles_all, atoms_data, constraints_type,
```
jgreener64, line 1328: Most of this function repeats the above function, you could bring the final part where you create the constraints out of this function to use the same code.

jgreener64 (commented):
Some more comments.

src/constraints/lincs.jl
Immi000: It's not, it was a leftover I forgot to remove
Edit: It was used for some tests but I've added a boundary there instead, I don't think it should stay in the code

src/setup.jl
Immi000: I initially chose the values here such that they satisfy the constraints for the 6mrr protein benchmark as there is no way to specify these parameters at system creation when reading from file and I didn't want to add additional parameters to the system constructor. But you're right, it's confusing if they don't have the default values, so I have reverted that.

Currently, the best way to modify these is to recreate the constraint from its data (and I guess the same is true for SHAKE_RATTLE):
```
lincs = LINCS(
          masses=Array(masses(sys)),
          dist_constraints=sys.constraints[1].dist_constraints,
          angle_constraints=sys.constraints[1].angle_constraints,
          dist_tolerance=sys.constraints[1].dist_tolerance,
          vel_tolerance=sys.constraints[1].vel_tolerance,
          nrec=6,
          niter=6,
          iter_vel_correction=true,
      )
      sys.constraints = (Molly.setup_constraints!(lincs, sys.neighbor_finder, AT),)
```

I think it's fine this way for now but is still a bit unintuitive. Perhaps it could be useful to add a copy-constructor like the one for the `System` where `setup_constraints!` is called internally.

PR #258, opened 2026-05-26
Times Square Sampling

allexandre97:
# Times Square Sampling

## Overview

TSS is an enhanced sampling technique developed by the D. E. Shaw research group. In spirit, it is quite similar to AWH, in the sense that it defines a ladder of thermodynamic states, integrates one of the "rungs" for a bit using MD (or any similar sampler), and then decides which rung to sample next based on a reweighting-like scheme of the generated configurations along the thermodynamic ladder.

However, TSS is a much more sophisticated and involved method. Much of its implementation is dedicated to computational efficiency. Instead of working on the full thermodynamic ladder, TSS defines overlapping windows along it, and the reweighting scheme that allows both the estimation of the free energy and the decision of which rung should be sampled next works on a subset of all the available rungs. Moreover, the TSS paper describes a covariance-determinant method to fine-tune the asymptotic reference density over the thermodynamic states upon convergence. Rather than targeting a uniform distribution over the lambda ladder, it focuses the computational efforts in the regions where the ensemble changes rapidly as a function of $\lambda$, and which can therefore be inherently more noisy.

Moreover, the method has been designed to be able to run multiple replicas within one TSS run, further improving sampling; has history forgetting, such that the very early stages of the algorithm do not contribute to the final estimates; and comes with an internal self-consistent way to estimate uncertainties along the thermodynamic ladder.

## Implementation details

Molly closely follows the details described in the original paper introducing TSS. Since the technique is quite involved, instead of writing the code into a monolithic file, a new directory has been created in `src/free_energy/TSS` to contain it. It splits:

- graph and window construction
- local and global free-energy estimators
- history forgetting
- observables and adaptive target control
- single-window and windowed simulation paths
- TSS-specific PMF deconvolution

TSS is thus designed to work as a submodule for Molly, and its main entry point is in `src/free_energy/TSS.jl`, where the general orchestration happens.

### Exported namespace

- `TSSState`, akin to `AWHState`, takes a vector of `ThermoState` objects and creates a struct able to represent the state a given TSS simulation is in at any given point.
- `TSSSimulation`, akin to `AWHSimulation`, takes a constructed `TSSState` and builds upon it to actually run a simulation using Times Square Sampling.
- `TSSHistoryForgetting` interfaces with `TSSState` and informs the subsequent simulations on which fraction of early epochs to discard, and how long each epoch should be.
- `tss_grid_graph` is a helper to set up the TSS grid, windowed or not, over the thermodynamic space. It supports an arbitrary number of dimensions and both periodic and non-periodic dimensions.
- `tss_free_energy` is a helper taking a `TSSState` and reporting back the free-energy estimate along the lambda ladder.
- `tss_free_energy_uncertainties` is a helper taking a `TSSState` and running a jackknife to estimate the uncertainty of the free energy along the lambda ladder. It returns a `TSSJackknifeResult` struct, unexported to the namespace, that carries the relevant information.
- `PMFDeconvolution` is a struct and family of methods that implements the deconvolution method described by [Lindahl et al.](https://doi.org/10.1063/1.4890371) to obtain the unbiased PMF from the free-energy surface calculated for an extended ensemble that uses bias potentials. The method was originally designed to work with AWH, but it has been extended to work with TSS for this pull request. This struct can be passed as an argument when constructing the relevant AWH or TSS states.
- `pmf` is a helper that takes a `PMFDeconvolution` struct and returns a `PMF` result, defined in `src/free_energy/mbar.jl`. It is probably a good idea to move it to a more general file in the future. The result contains the deconvolved PMF and some extra data, such as the PMF grid, the uncertainty, etc.

### Homogenisation with AWH

Some of the core functionality of TSS is shared with AWH and, as previously mentioned, the deconvolution method to obtain the unbiased PMF was extended to work with TSS too. Therefore, a decent amount of work for this PR has been dedicated to creating the necessary interfaces for these two techniques to share. The shared machinery can be found in `src/free_energy/reweighting.jl` and `src/free_energy/pmf_deconvolution.jl`.

Ideally, this should be general enough and serve as a scaffold for future additions to the code.

### Tutorials

The file in `docs/src/free_energy.md` has been greatly reworked. Since changes in AWH were concomitant with the work done for TSS, the tutorials for that have also been rewritten. In general, we describe two cases: calculating a 2D PMF along the internal dihedrals of an alanine dipeptide, and calculating the free energy of solvation of ethanol in water. A new directory has been created in the Molly root at `scripts/`, which contains the scripts to run said simulations. Two PDB files have also been added to the `data/` directory, necessary to run the solvation case. The results of AWH and TSS match, which is a good sign and indicative of the implementations being correct. I am currently waiting for longer 100 ns simulations to finish, so the current version uses some sub-optimal images produced with 10 ns simulations. It is also worth mentioning that TSS is considerably faster than AWH, even when running 2 replicas at a time!

## Compatibility Notes

This PR changes the recommended AWH PMF workflow from constructor keywords such as `pmf_grid` to an explicit `PMFDeconvolution` object passed with `pmf=...`.

## Extra bits and pieces

- Added numerical safeguards for the soft-core potentials. In general, we now have better monitoring if two atoms completely overlap and guard against divisions by zero when that happens.
- Added some extra safeguards to some collective variables and bias potentials. We check for some geometric degeneracies, such as collinearity in torsion angles, and avoid spurious math that could cause infinite forces.


ext/MollyCUDAExt.jl
```diff
@@ -24,6 +24,8 @@ using CUDA
 using Atomix
 using KernelAbstractions
 
+const CUDA_CORE = isdefined(CUDA, :CUDACore) ? CUDA.CUDACore : CUDA
```
jgreener64: Add a note that this can be removed in future, is there a relevant link?

src/bias/bias.jl
```diff
@@ -78,6 +80,14 @@ function bias_gradient(sb::SquareBias, cv_sim)
     return sb.k * (cv_sim - sb.cv_target)
 end
 
+
+function _validate_flat_bottom_width(r_fb, label::AbstractString)
```
jgreener64: No leading underscore, and for other functions.

src/bias/bias.jl
```diff
@@ -215,6 +238,30 @@ struct BiasPotential{C, B}
...
+    _bias_all_finite(value) && return value
+    msg = "BiasPotential with CV $(typeof(bias.cv_type)) and bias " *
+          "$(typeof(bias.bias_type)) produced non-finite $(label)"
+    if !isnothing(cv_sim)
+        msg *= ", cv_sim=$(cv_sim)"
+    end
+    if !isnothing(max_abs_component)
+        msg *= ", max_abs_component=$(max_abs_component)"
+    end
+    throw(ArgumentError(msg * "."))
```
jgreener64: This could use `error` instead (later we may define Molly error types).

src/interactions/ewald.jl
```diff
@@ -73,9 +73,10 @@ end
     return charge(atom) * electrostatic_lambda(scheduler, atom, Val(T))
 end
 
-# Passing vec_ij and r, and not having calculate_forces as a Val, was required to avoid allocations
+# IMPORTANT Passing vec_ij and r, and not having calculate_forces as a Val, was required to avoid allocations
 function excluded_interactions_inner!(Fs, vir, atoms, coords, boundary, α, f_div_ϵr, scheduler,
-                            i, j, vec_ij, r, calculate_forces, ::Val{T}, ::Val{atomic},
+                            i, j, vec_ij, r,
```
jgreener64: The change in these functions is just for the argument order?

src/interactions/coulomb.jl
```diff
@@ -44,7 +44,7 @@ function Base.:+(c1::Coulomb, c2::Coulomb)
     )
 end
 
-function inject_interaction(inter::Coulomb, params_dic)
+function inject_interaction(inter::Coulomb, params_dic::AbstractDict)
```
jgreener64: Is this type annotation required (and elsewhere)? If not then remove.

test/interactions.jl
```diff
@@ -135,7 +135,7 @@
     )
     @test isapprox(
         potential_energy(inter, dr14, a1, a1),
-        45.35853723577515u"kJ * mol^-1";
+        25.90746475905439u"kJ * mol^-1";
```
jgreener64, line 138: Was this wrong before?

test/tss.jl
```diff
@@ -0,0 +1,1209 @@
+function make_tss_thermo_states(; n_atoms=6, n_states=3)
```
jgreener64, line 4: Any way to trim this file whilst still testing the main features? 1,200 lines is a lot.

src/constraints/constraints.jl
```diff
@@ -432,7 +432,20 @@ end
 abstract type ConstraintKernelData{D, N, M} end
 
 n_constraints(::ConstraintKernelData{<:Any, N}) where {N} = N
+n_constraints(::Type{<:ConstraintKernelData{<:Any, N}}) where {N} = N
 n_atoms_cluster(::ConstraintKernelData{<:Any, <:Any, M}) where {M} = M
+n_atoms_cluster(::Type{<:ConstraintKernelData{<:Any, <:Any, M}}) where {M} = M
+
+host_constraint_clusters(constraint_clusters) = constraint_clusters
+host_constraint_clusters(constraint_clusters::StructArray) =
```
jgreener64, line 509: In brief, what is the need for this?

src/constraints/constraints.jl
```diff
@@ -432,7 +432,20 @@ end
...
+    replace_storage(Array, constraint_clusters)
+
+function n_dof_lost(D::Integer,
```
jgreener64: Don't we have this function somewhere else? Maybe it is for a different case.

src/free_energy/pmf_deconvolution.jl
```diff
@@ -0,0 +1,356 @@
...
+            )
+        end
+        dest[bin_i] = isfinite(log_den) ? lwf - log_den : -T(Inf)
+    end
+    return dest
+end
+
+function _pmf_positive_log(value, label::AbstractString, ::Type{T}) where T
+    v = T(value)
+    isfinite(v) && v > zero(T) ||
```
jgreener64: Use `if` if multi-line.

src/free_energy/TSS/windows.jl
```diff
@@ -0,0 +1,965 @@
...
+
+function Base.show(io::IO, graph::TSSGraph)
+    print(io, "TSSGraph with ", _tss_count(graph.n_states, "state"), ", ",
+          _tss_count(length(graph.windows), "window"), ", ",
+          _tss_count(length(graph.rung_edges), "edge"))
+end
+
+Base.show(io::IO, ::MIME"text/plain", graph::TSSGraph) = show(io, graph)
+
+struct _TSSGraphEdge
```
jgreener64: No underscore.

src/free_energy/TSS.jl
```diff
@@ -0,0 +1,16 @@
+export
+    TSSHistoryForgetting,
+    TSSState,
+    TSSSimulation,
+    tss_grid_graph,
```
jgreener64, line 5: Can some of these be non-exported? Later I'm thinking of removing some other exports from the package but marking them as `public` so they are still user-facing.

jgreener64 (commented):
Looks great, though I can't pretend to have read it all. I have left some comments and might try and work through it more later.

ext/MollyCUDAExt.jl
allexandre97: I have added a comment on top explaining why its necessary at the moment. The bigger issue here is that the exported `Const` method for CUDA seems to conflict with something else, therefore annotation is necessary. Also, `shfl_recurse` was mistakenly unexported, they have opened and merger a PR solving that, but I believe its not available yet in the Julia repos

src/bias/bias.jl
allexandre97: Cleaned up

src/bias/bias.jl
allexandre97: Changed to `error` here.

I've made extensive use of `ArgumentError` for the TSS implementation, should I use `error` there too maybe?

src/interactions/ewald.jl
allexandre97: Here I've propagated the use of `::Val{calculate_forces}` to the topmost layer of the method calls, that is the main difference. I forgot to push that change, it's going to be in the next commit addressing these PR commits.

src/interactions/coulomb.jl
allexandre97: This is stale code from a branch that I have where I have changed the `inject_interactions` to work with arrays instead of dictionaries, so annotation was necessary there. I copypasted some stuff from that branch and forgot to remove this here. I'll clean it up.

test/tss.jl
allexandre97: I agree, I let codex loose writing the tests form me. I will trim as much as possible.

src/constraints/constraints.jl
allexandre97: This was necessary because calling `deepcopy(sys::System)` on a `System` using `CuArray` and with constraints would error. These methods move things from/to device as required, since if the clusters are stored as a `StructArray` with GPU storage, directly iterating them can error with a "scalar index disallowed"

src/constraints/constraints.jl
allexandre97: Yes, this is an overload of the generic method. Again, necessary to avoid iterating over GPU-backend data types. This method handles handles vectors whose element type is a concrete `ConstraintKernelData` type, in which case all clusters have the same number of atoms, so we compute the dof loss from element type times the vector length.

src/free_energy/TSS/windows.jl
allexandre97: Removed all underscores

src/free_energy/TSS.jl
allexandre97: I exported what was necessary to run the example files in the `scripts` directory, but we can have a chat about what to export and what not to export. I agree that the namespace is becoming quite bloated now

src/bias/bias.jl
allexandre97: Also, this required a change in `test/bias.jl` line 489. Now the test chacks if the case throws an `ErrorException` instead of an `ArgumentError`

src/free_energy/pmf_deconvolution.jl
allexandre97: Changed all these flow control idioms to follow the explicit `if` path

src/constraints/constraints.jl
jgreener64: Okay, move it under the other `n_dof_lost` function.

src/constraints/lincs.jl
```diff
@@ -998,6 +998,8 @@ function setup_constraints!(lincs::LINCS, neighbor_finder, arr_type)
     end
 
     if arr_type <: AbstractGPUArray
+        lincs.lincs_data.atom1 isa arr_type && return lincs
```
jgreener64: `if arr_type <: AbstractGPUArray && !(lincs.lincs_data.atom1 isa arr_type)`

src/force.jl
```diff
@@ -60,6 +60,7 @@ force_gpu(inter, ci, cj, ck, cl, bnd, ai, aj, ak, al, fu, vi, vj, vk, vl, sn, da
 force_gpu(inter, ci, cj, ck, cl, cm, bnd, ai, aj, ak, al, am, fu, vi, vj, vk, vl, vm, sn, data) = force(inter, ci, cj, ck, cl, cm, bnd, ai, aj, ak, al, am, fu, vi, vj, vk, vl, vm, sn, data)
 
 @inline zero_pairwise_force(dr, force_units) = ustrip.(zero(dr)) * force_units
+@inline zero_pairwise_energy(dr, energy_units) = ustrip(zero(dr[1])) * energy_units
```
jgreener64: Minor but this should probably go in src/energy.jl.

test/interactions.jl
jgreener64: Still wondering about this.

test/tss_pmf_deconvolution.jl
```diff
@@ -0,0 +1,203 @@
+function make_tss_pmf_thermo_states(; n_atoms=4, n_states=4)
```
jgreener64: May as well just have one test/tss.jl file.

src/free_energy/AWH.jl
```diff
@@ -73,37 +70,76 @@ mutable struct AWHState{T, P, S, I, B, PR, SPI, SSI, SGI}
     stats::AWHStats{T}
 end
 
+function Base.getproperty(state::AWHState, name::Symbol)
```
jgreener64, line 98: I think there is a cleaner way to do this with metaprogramming, looping over tuples of symbols. Or change things round so it isn't needed.

jgreener64 (commented):
Okay great, merge after making the below changes. And do check that all the code here is required, as it is a lot to maintain.

src/constraints/constraints.jl
allexandre97: Moved

src/constraints/lincs.jl
allexandre97: Changed

src/free_energy/AWH.jl
allexandre97: Changed to the suggested meta-programming approach

test/interactions.jl
allexandre97: This change is due to what the default value for alpha is in  the `LennardJonesSoftCoreGapsys` potential. It was set to be 1, changed it to 0.85 since that is the standard that GROMACS uses.

src/force.jl
allexandre97: Moved it to src/energy.jl

test/tss_pmf_deconvolution.jl
allexandre97: Merged everything into one file.

PR #259, opened 2026-05-28
feat: Add Lux.jl-based Machine Learning Potentials integration

rainerrodrigues:
Hi team! I am preparing a proposal to expand Machine Learning Potentials in Molly. As a proof of concept, I have implemented a native interface for explicit-parameter networks using Lux.jl. Unlike the Flux example in the docs (which mutates internal state and predicts forces directly), this implementation separates architecture, parameters, and state (model, ps, st), evaluating scalar potential energy and utilizing Zygote.jl to compute analytical, energy-conserving forces.

What this PR does:

- Adds the MachineLearningPotential struct to src/interactions/machine_learning.jl.

- Implements potential_energy and forces evaluation pipelines.

- Adds a 3-atom unitless test suite verifying energy and force calculations.

- Adds basic documentation to differentiable.md.

Looking forward to any feedback on the architecture!


jgreener64:
Thanks for this. We could certainly think about ways to make interacting with Lux models either. However, this implementation doesn't pass in the element of each atom, doesn't account for the boundary and probably won't work with units. I think we should also move away from Zygote, which is increasingly unmaintained. Maybe this could go in the examples.

The question is whether it's possible to get something suitably generic here, or whether there is use in making the user define these functions themselves to be explicit and avoid issues.

rainerrodrigues:
Thanks for the excellent feedback, @jgreener64! That makes total sense. This implementation definitely glosses over critical edge cases like periodic boundaries, chemical species embeddings, and Unitful dimensions that a production-ready interface requires. I also really appreciate the note regarding the ecosystem shift toward Enzyme over Zygote for performance.
I'd love to move this code over to the examples/ directory as a standalone script showing users how to explicitly write a custom Lux potential for a simple system.

rainerrodrigues:
@jgreener64 Where should I go about putting this in the examples section in the docs, assuming I fix the suggested changes to this code?

jgreener64:
It could go after "Python ASE calculator" in docs/src/examples.md.

rainerrodrigues:
@jgreener64  It looks like the latest CI run failed on a memory allocation check in test/protein.jl:140 (4 <= 3). Since this PR didn't touch the Amber protein simulation code, I assume this is a transient/flaky benchmark issue on the CI runner, but let me know if I need to rebase!

test/Project.toml
```diff
@@ -15,13 +15,17 @@ GLMakie = "e9467ef8-e4e7-5192-8a1a-b1aee30e663a"
 GPUArrays = "0c68f7d7-f131-5f86-a1c3-88cf8149b2d7"
```
jgreener64, line 15: No changes here.

docs/src/differentiable.md
```diff
@@ -432,6 +432,29 @@ The [code is available](https://github.com/greener-group/GB99dms).
 Molly.jl was also used to code [reversible simulation](https://doi.org/10.1073/pnas.2426058122), an extension of differentiable simulation with RAD where the gradients are calculated explicitly.
 This means the memory cost is constant in step number.
 
+## Machine Learning Potentials
```
jgreener64: No changes here.

docs/src/examples.md
```diff
@@ -608,6 +608,84 @@ sys = System(
 potential_energy(sys) # -2080.2391023908813 eV
 ```
 
+## Lux.jl machine learning potential
+[Lux.jl](https://lux.csail.mit.edu) models can be integrated into Molly simulations as explicit-parameter neural network potentials. This approach separates the network architecture from its parameters and state (`model`, `ps`, `st`), which ensures type-stability and facilitates high-performance scientific machine learning.
+
+A complete, runnable script demonstrating how to define a custom `MachineLearningPotential` struct, evaluate potential energy, and compute analytical, energy-conserving forces using `Zygote.jl` can be found in the repository at:
```
jgreener64: No need to include the separate script if it is here, remove this reference.

jgreener64 (commented):
Looks good, the CI failure is unrelated.

docs/src/examples.md
rainerrodrigues: Alright, I make the changes

rainerrodrigues: @jgreener64 I think it's good to go for merging.

PR #260, opened 2026-06-01
Add native ANI-2x machine-learning potential: CPU + GPU energy and analytic forces (GSoC 2026)

sivasathyaseeelan:
## Summary

Native Julia **ANI-2x** neural-network potential for Molly.jl, developed for GSoC 2026. It runs the whole AEV + per-element neural-network pipeline in pure Julia (no Python runtime, unlike the existing `ASECalculator` path), with **energy and forces on both CPU and GPU** (Apple Metal today, CUDA-portable via KernelAbstractions).

### What this delivers

- **Energy** on CPU (threaded, zero-allocation AEV) and on-device (`compute_ani_energy_ka`).
- **Forces via a single analytic path**: `AtomsCalculators.forces!` → `compute_ani_forces_ka`, an analytic backward (forward AEV → manual NN VJP `∂E/∂G` → backward radial/angular AEV kernels `∂E/∂r`, atomic equal-and-opposite scatter). Runs on CPU (KA CPU backend) and GPU. `forces(sys)` works for both CPU and Metal-backed systems.
- Validated against **TorchANI** and finite differences to **~1e-6 eV/Å**, with ΣF ≈ 0.
- 8-member ensemble averaging; minimum-image convention via `Molly.vector` (Cubic + Triclinic).
- **19/19** ANI testsets pass; Aqua clean.

### How ANI flows through the code

```mermaid
flowchart TD
    A["sys.coords (with units)"] -->|"strip units → Å"| B["coords in Å + species_idx"]
    B --> AEV

    subgraph AEV["AEV build · per atom, over neighbours within cutoff"]
      direction TB
      FC["cutoff f_C — Eq. 2"]
      FC --> R["radial G^R — Eq. 3"]
      FC --> AN["angular G^A — Eq. 4"]
    end

    AEV --> M["AEV matrix (n_atoms × 1008)"]
    M -->|"per element: one matmul"| NN["element NNs → E_i + self-energy"]
    NN -->|"Σ_i E_i — Eq. 1 · average over 8 networks"| EH["E (Hartree)"]
    EH -->|"× 27.2114"| EV["E (eV)"]

    B -. "analytic backward: NN VJP (∂E/∂G) + backward AEV kernels (∂E/∂r)" .-> FF["forces F = −∇E (eV/Å)"]
    EV -.-> FF

    classDef out fill:#e8f5e9,stroke:#2e7d32;
    class EV,FF out;
```

Equations follow ANI-1 (Smith et al., *Chem. Sci.* 2017); ensemble averaging follows ANI-1x (Smith et al., *JCP* 2018); element set follows ANI-2x (Devereux et al., *JCTC* 2020). Every equation-implementing function cites its equation.

### Performance (Apple M3, 6mrr slices)

- **Full 6mrr protein (15,954 atoms):** on-device energy **158 ms**, forces **409 ms**.
- **vs TorchANI CPU** at the full protein: energy **~23×** faster, forces **~11×** faster. (TorchANI's optimised CPU kernels win at small N; Molly scales far better.) On the Apple GPU, TorchANI has no usable path at all (PyTorch-MPS lacks float64 and a 5-D scan in the angular AEV), so Molly's native Metal implementation stands alone. A true GPU-to-GPU comparison needs CUDA.
- **CPU forces** (analytic, single path) are **7–14× faster** than the earlier Enzyme AD path (138 vs 1829 ms at 1000 atoms).
- Full benchmark suite, tables and figures in `benchmark/ani_benchmarks.md`.

### Architecture

- `src/interactions/ml_potentials.jl` — core: `AbstractMLPotential`, `ANIPotential`, scalar AEV helpers, public stubs.
- `ext/MollyLuxExt.jl` (Lux + HDF5) — CPU AEV, energy, HDF5 weight loader.
- `ext/MollyKALuxExt.jl` (KernelAbstractions + Lux + HDF5) — GPU-portable AEV kernels, on-device energy (`compute_ani_energy_ka`), the analytic forces (`compute_ani_forces_ka`) and the single `AtomsCalculators.forces!`.

### Development notes

Built up over GSoC weeks 1–8: infrastructure and AEV correctness (validated to ~1e-8 vs TorchANI), ensemble averaging, zero-allocation threaded AEV, GPU neighbour-list kernels (O(N²)/O(N³) → O(N·k)/O(N·k²)), on-device energy, then the on-device analytic forces. Per review, forces were **consolidated to the single analytic path**: the earlier finite-difference and Enzyme reverse-mode force paths were removed (Enzyme remains only for Molly's unrelated general AD). Recent additions: device-parameter caching, ensemble Metal forces (~1.7× a single member), and the TorchANI head-to-head. Full history is in the commits.

## Test plan

- Generate reference data: `python test/torchani_reference.py`
- Run the ANI tests: `julia --project -t8 -e 'using Molly, Lux, HDF5, KernelAbstractions; include("test/ml_potentials.jl")'`
- **19/19 testsets pass** (AEV, loading, forward pass, energy, N₂/H₂O forces vs TorchANI, ensemble, neighbour-list consistency, NVE MD, 6mrr 50-atom and full 15,954-atom energy, KA AEV kernel, threaded≡serial AEV, KA neighbour-list kernel, on-device energy, minimum-image, on-device forces with finite-difference + TorchANI cross-checks).
- Forces now require `KernelAbstractions` (the single analytic path); energy needs `Lux`, `HDF5`.
- Benchmarks: `julia --project -t8 benchmark/run_ani_benchmarks.jl` (see `benchmark/ani_benchmarks.md`).


test/torchani_reference.py
```diff
@@ -0,0 +1,282 @@
+#!/usr/bin/env python3
```
jgreener64, line 1: Maybe move this to `test`.

src/types.jl
```diff
@@ -1828,6 +1832,49 @@ end
...
+# Concrete subtypes (ANIPotential, etc.) are defined in ext/MollyLuxExt.jl.
+abstract type AbstractMLPotential end
+
+"""
+    ANIPotential(path; force_units, energy_units, T, ensemble_idx)
+
+Load an ANI neural network potential from an HDF5 file exported by
+`scripts/torchani_reference.py`. Requires `Lux` and `HDF5` to be loaded.
+"""
+struct ANIPotential{M, PS, ST, SP, P, SE, D, F, E} <: AbstractMLPotential
```
jgreener64: This can be in a new file in `src/interactions`.

src/types.jl
```diff
@@ -1828,6 +1832,49 @@ end
 
 function update_ase_calc! end
 
+# Base type for ML interatomic potentials.
+# Concrete subtypes (ANIPotential, etc.) are defined in ext/MollyLuxExt.jl.
+abstract type AbstractMLPotential end
```
jgreener64: Only add this if it is useful for dispatch.

src/types.jl
```diff
@@ -1828,6 +1832,49 @@ end
...
+end
+
+"""
+    compute_aevs(coords, species_indices, neighbors, boundary, aev_params, n_species)
+
+Compute atomic environment vectors (AEVs) for all atoms.
+Returns a `(n_atoms, aev_length)` matrix.
+Implementation is in ext/MollyLuxExt.jl.
+"""
+function compute_aevs end
```
jgreener64: These can either be fully in the extension, or fully in the new interaction file and imported in the extension.

ext/MollyLuxExt.jl
```diff
@@ -0,0 +1,346 @@
...
+                      boundary,
+                      p,          # aev_params NamedTuple
+                      n_species::Int) where {D,T}
+    n_atoms      = length(coords)
+    n_rad_per    = length(p.η_R) * length(p.r_s_R)    # outer product: EtaR × ShfR
+    n_ang_per    = length(p.η_A) * length(p.r_s_A) * length(p.θ_s)  # EtaA × ShfZ × ShfA
+    n_pairs      = n_species * (n_species + 1) ÷ 2
+    aev_len      = n_species * n_rad_per + n_pairs * n_ang_per
+
+    aevs    = zeros(T, n_atoms, aev_len)
```
jgreener64: Does this allocate memory every call? It would be good to pre-allocate the required buffers so that `potential_energy` does not allocate memory.

jgreener64 (commented):
Looks like a great start. Ideally we would test on a larger system like `data/6mrr_equil.pdb` which includes more element types and periodic boundaries.

I think to merge this PR we would also want Enzyme being used to get the forces, and an example in the docs.

ext/MollyLuxEnzymeExt.jl
```diff
@@ -0,0 +1,184 @@
...
+    xval = tape
+    d = xval >= zero(xval) ? one(xval) : exp(xval / typeof(xval)(0.1f0))
+    return (nothing, dret.val * d)
+end
+
+# ============================================================================
+# Shared utilities
+# ============================================================================
+
+function _lue_strip_boundary(b::CubicBoundary)
```
allexandre97: Just stylistic, but avoid using underscores to start name methods, even if they are meant to be unexported / private

ext/MollyLuxEnzymeExt.jl
```diff
@@ -0,0 +1,184 @@
...
+# coords_mat: (3, N) Float32 — the only Active argument.
+# All other args are Const so Enzyme doesn't need to differentiate through them.
+#
+# Batched by species: gather each element's AEV rows into one (aev_len, n_s) matrix
+# and call Lux.apply once per element rather than once per atom. This cuts the number
+# of Lux.apply calls Enzyme has to trace from n_atoms to n_species, which both speeds
+# up the reverse pass and reduces allocation. `group_atoms[s]` lists the atom indices
+# of species s; the gather/permutedims are differentiable so forces stay exact.
+function _ani_energy_for_ad(
+    coords_mat        :: AbstractMatrix{Float32},
```
allexandre97: I understand that you are hard-coding this in `Float32` because the ANI-2.x weights are in that precision. However, Molly does allow to run simulation in any Float type. In other places you dispatch on `sys::System{D, AT, T}` with arbitrary `D` and `T`, but then cast back to `Float32`. 

If the force path intentionally only works in `Float32` you should explicitly document it, and do not dispatch on arbitrary `T`. If not, you should propagate the generic `T` and do something like `E = zero(eltype(coords_mat))`

src/types.jl
```diff
@@ -1828,6 +1834,75 @@ end
...
+
+"""
+    ANIPotential(path; force_units, energy_units, T, ensemble_idx)
+
+Load an ANI neural network potential from an HDF5 file exported by
+`scripts/torchani_reference.py`. Requires `Lux` and `HDF5` to be loaded.
+
+By default all ensemble members are loaded and energies are averaged.
+Pass `ensemble_idx` to load only a specific member (wrapped in a length-1 vector).
+"""
```
allexandre97: In `ext/MollyLuxEnzymeExt.jl you` force a `CubicBoundary`. If that is intended, clearly document that this potential can only be used with cubic boundaries, since Molly does support other geometries.

ext/MollyLuxEnzymeExt.jl
```diff
@@ -0,0 +1,184 @@
...
+                                   sys::System{D, AT, T},
+                                   inter::ANIPotential;
+                                   kwargs...) where {D, AT, T}
+    Ha_to_eV = T(27.211396132)
+
+    coords_f32  = _lue_f32coords(sys.coords)
+    species_idx = [inter.species_map[ad.element] for ad in sys.atoms_data]
+    nbrs        = get(kwargs, :neighbors, nothing)
+    bdy         = _lue_strip_boundary(sys.boundary)
+    bdy_f32     = CubicBoundary(Float32.(bdy.side_lengths))
```
allexandre97: Do ANI potentials only work with Cubic boundaries?

ext/MollyLuxExt.jl
```diff
@@ -0,0 +1,631 @@
...
+        if isnothing(neighbors)
+            for j in 1:n_atoms
+                j == atom_i && continue
+                dr = coords[j] - coords[atom_i]
+                norm(dr) < r_c_max || continue
+                n_nbrs += 1
+                sc.nbr_coords[n_nbrs]  = coords[j]
+                sc.nbr_species[n_nbrs] = Int(species_indices[j])
+            end
+        else
```
allexandre97: Should this no-neighbor path use `vector(coords[atom_i], coords[j], boundary)` like the neighbor-list path below? As written, this stores the raw coordinate. Later `_radial_aev!` / `_angular_aev!` subtract `coord_i` from `nbr_coords`, so atoms close across a periodic boundary will be missed unless a neighbor finder is used.

You could we either coalesce here with `nbr_coords[n] = coords[atom_i] + vector(...)`, or document that periodic ANI systems require a neighbor finder.

ext/MollyLuxExt.jl
```diff
@@ -0,0 +1,631 @@
...
+    split     = n_species * n_rad_per
+
+    nchunks = clamp(length(buf.scratch), 1, max(n_atoms, 1))
+    if nchunks == 1
+        _aev_chunk!(buf, 1:n_atoms, buf.scratch[1], coords, species_indices,
+                    neighbors, boundary, p, n_species, r_c_R, r_c_A, r_c_max, split, aev_len)
+    else
+        base = n_atoms ÷ nchunks
+        rem  = n_atoms % nchunks
+        Threads.@threads :static for c in 1:nchunks
```
allexandre97, line 373: Why is `:static` needed here? Does this give better performance than with load balancing?

ext/MollyLuxExt.jl
```diff
@@ -0,0 +1,631 @@
...
+    fcj_buf     = Vector{T}(undef, n_atoms)
+    drj_buf     = Vector{SVector{D,T}}(undef, n_atoms)
+    ok_buf      = Vector{Bool}(undef, n_atoms)
+
+    for atom_i in 1:n_atoms
+        n_nbrs = 0
+        if isnothing(neighbors)
+            for j in 1:n_atoms
+                j == atom_i && continue
+                dr = coords[j] - coords[atom_i]
```
allexandre97: Same issue as above, this could violate minimum image convention. Fix or document that either neighbors are required or PBC systems are not supported

ext/MollyLuxExt.jl
```diff
@@ -0,0 +1,631 @@
...
+        b   = Float32.(read(lg["bias"]))                  # (out,)
+        out_sz, in_sz = size(W)
+        act = k < n_dense ? _celu01 : identity
+        push!(lux_layers, Lux.Dense(in_sz => out_sz, act))
+        push!(ps_pairs, Symbol("layer_", k) => (weight = W, bias = b))
+    end
+
+    model = Lux.Chain(lux_layers...)
+    ps    = NamedTuple(ps_pairs)
+    _, st = Lux.setup(Xoshiro(0), model)
```
allexandre97, line 485: Add a comment as to why you hard code the RNG. I understand that it is because you are discarding the parameters, and therefore the RNG is not relevant. Is there a cleaner way of doing this?

ext/MollyLuxExt.jl
```diff
@@ -0,0 +1,631 @@
...
+    Ha_to_eV = T(27.211396132)
+    E_eV = E_ha * Ha_to_eV
+
+    if sys.energy_units == NoUnits
+        return E_eV
+    elseif dimension(sys.energy_units) == u"𝐋^2 * 𝐌 * 𝐍^-1 * 𝐓^-2"
+        return uconvert(sys.energy_units, E_eV * Unitful.Na * u"eV")
+    else
+        return uconvert(sys.energy_units, E_eV * u"eV")
+    end
```
allexandre97: This pattern appears elsewhere, you could define a helper to avoid repeating code. I would also turn `Ha_to_eV` into a constant outside the method since it is used by several functions

ext/MollyLuxExt.jl
```diff
@@ -0,0 +1,631 @@
...
+end
+
+# Forces via finite differences (fallback when Enzyme is not loaded).
+function AtomsCalculators.forces!(fs,
+                                  sys::System{D, AT, T},
+                                  inter::ANIPotential;
+                                  kwargs...) where {D, AT, T}
+    h = T(1e-4)   # Å — must be T (Float64) for numerical precision
```
allexandre97: In line with other comments, `T` in principle can be any Float type. If this is not intended to be the case for this PR, clearly document it and add appropriate guards.

ext/MollyKALuxExt.jl
```diff
@@ -0,0 +1,333 @@
...
+using StaticArrays, LinearAlgebra
+
+# ============================================================================
+# Scalar helpers (device-compatible — no closures, no allocation)
+# ============================================================================
+
+# Smooth cosine cutoff f_C — [ANI-1, Smith et al. Chem. Sci. 2017] Eq. (2).
+@inline function _ka_cosine_cutoff(r::T, r_c::T) where T
+    r < r_c ? T(0.5) * (one(T) + cos(T(π) * r / r_c)) : zero(T)
+end
```
allexandre97: This appears to duplicate `Molly.cosine_cutoff` from `MollyLuxExt.jl` with the same body. Since the cutoff is a simple scalar function and has no Lux/HDF5 dependency, could we define it once in core Molly and reuse it here?

ext/MollyKALuxExt.jl
```diff
@@ -0,0 +1,333 @@
...
+# ============================================================================
+# KernelAbstractions AEV kernel — one thread per central atom
+# ============================================================================
+
+# Compute radial + angular AEV for atom `atom_i` from a dense distance table.
+# GPU-portable form of _radial_aev! / _angular_aev! — same equations:
+#   radial block  = [ANI-1] Eq. (3) G^R,   angular block = [ANI-1] Eq. (4) G^A.
+# Uses the O(N²) all-pairs approach: each thread scans all n_atoms neighbours.
+# For neighbour-list systems: call _aev_kernel_nl! instead (see below).
+@kernel inbounds=true function _aev_kernel!(
```
allexandre97: This method suffers from the issue of potential violation of minimum image convention. The distance between atoms is calculated without the `Molly.vector(...)` method. In fact, you don't even take the boundary as a method.

If this is intended, and PBC are not supported, document it clearly and add the necessary guards.

ext/molly_lux_ka.jl
```diff
@@ -0,0 +1,333 @@
...
+# Public interface: compute_aevs_ka — GPU-portable AEV computation
+# ============================================================================
+
+"""
+    Molly.compute_aevs_ka(coords, species_indices, p, n_species;
+                          backend=nothing, neighbors=nothing, workgroup=256)
+
+GPU-portable AEV computation using KernelAbstractions. One thread per atom.
+`coords` can be a CPU Vector{SVector} or a GPU array (CuArray, ROCArray, etc.).
+Pass `backend=MetalBackend()`/`CUDABackend()` to run on GPU; defaults to the
```
allexandre97: This seems to diverge to how we dispatch depending on array type elsewhere in Molly. Users can select an execution mode by doing something like `System(...; array_type = CuArray)`. In fact, this function already defaults to `KernelAbstractions.get_backend(coords)`. 

Moreover, `backend=...` does not appear to move `coords` or `species` to the GPU, so the docstring may be misleading.

allexandre97 (commented):
Overall looks good! The only pressing issue that I found is that you consistently compute distances directly as the difference between the position of two atoms, instead of using the `vector` method defined in Molly. If the system has periodic boundaries, doing it with the subtraction method can violate the minimum image convention. If this is intended to be used only on systems without PBC you should clearly document it and guard against it.

sivasathyaseeelan:
Thanks @jgreener64 and @allexandre97 for the thorough reviews! I've pushed three commits addressing all the points:

- `657bd347` — minimum-image in all-pairs paths + shared `cosine_cutoff`
- `6c3cb566` — move ML potentials to a core interactions file + docs + shared helpers + script move
- `10cd92f7` — drop leading underscores from internal helper names

All 17 ANI tests still pass.

### Correctness / PBC
- **All-pairs paths stored raw coordinates (minimum-image, MollyLuxExt & KA kernel).** The CPU all-pairs branches (`compute_aevs`, the buffered chunk loop) now build displacements with `vector(coords_i, coords_j, boundary)` and store the imaged position, so they respect the minimum-image convention. The KA GPU kernels intentionally use raw differences — documented at the top of `MollyKALuxExt.jl` (non-periodic / pre-imaged use; periodic systems use the CPU path).

### Structure / reuse
- **`cosine_cutoff` duplicated in the KA ext.** Now defined once in core Molly (`src/interactions/ml_potentials.jl`, no Lux/HDF5 dep) and reused by both extensions; `_ka_cosine_cutoff` removed.
- **ANIPotential struct in `src/types.jl` / stubs placement.** Moved `AbstractMLPotential`, `ANIPotential`, `celu01`, `cosine_cutoff` and the public stubs into a new `src/interactions/ml_potentials.jl`, included from `Molly.jl` — alongside the extensions that implement them. `AbstractMLPotential` is kept as the shared supertype for dispatch.
- **Repeated `Ha_to_eV` / unit-conversion blocks.** Hoisted into `const HARTREE_TO_EV` + `ani_energy_to_units` / `ani_force_to_units` helpers in core; all three extensions use them.
- **`scripts/torchani_reference.py`.** Moved to `test/torchani_reference.py`; all references updated.
- **Leading underscores on private methods.** Renamed all internal ANI helpers to match Molly's convention (e.g. `_radial_aev!` → `radial_aev!`).
- **Pre-allocation.** `potential_energy` uses the reusable `AEVBuffers` (0 heap allocations after warmup); this is unchanged from the earlier commits.

### Precision / boundaries (documented)
- **`Float32` hardcoding vs generic `T`.** The ANI-2x weights are `Float32`, so the energy/force paths run in `Float32` internally regardless of the coordinate type. This is now documented in the `ANIPotential` docstring rather than implying full generic-`T` support.
- **CubicBoundary only.** Documented that periodic systems require a neighbour finder and that the boundary handling assumes `CubicBoundary`.
- **`compute_aevs_ka` `backend=` docstring.** Clarified that `coords`/`species` must already be on the target device and that `backend` defaults to `KernelAbstractions.get_backend(coords)`.

### Minor
- **Hard-coded RNG (`Xoshiro(0)`).** Commented: the randomly-initialised parameters are discarded and replaced by the HDF5 weights, so the seed only makes construction deterministic.
- **`Threads.@threads :static`.** Commented: it fixes chunk c → `scratch[c]` (each chunk owns its buffer); it's not for load balancing (chunks are equal-sized, `nchunks == nthreads`).

Happy to adjust any of these further.

mitkotak:
Was pointed to this PR by @abhijeetgangan. Cool work ! I maintain the [JAX equivalent](https://github.com/mitkotak/bio-mlff/) of what you are doing.

Not sure how much optimization work has been done already but there were 2 useful optimization tricks that I found useful while writing the JAX version:

- Since the AEV scales with the number of species, you can shrink the size of your [radial AEV](https://github.com/mitkotak/bio-mlff/blob/c67d395406c1eb092e1c45cbce81ca6262e3b5b1/biomlff/ani.py#L322-L331) and [angular AEV](https://github.com/mitkotak/bio-mlff/blob/c67d395406c1eb092e1c45cbce81ca6262e3b5b1/biomlff/ani.py#L377-L388) by only computing the active species (For e.g. H and O for water). Since the AEV kernels are memory-bound this helps a lot. 


- The second optimization is in avoiding the big concatenate of radial AEV and angular AEV. Instead you can apply the weights of the first layer to both the [AEVs individually and then sum them up](https://github.com/mitkotak/bio-mlff/blob/c67d395406c1eb092e1c45cbce81ca6262e3b5b1/biomlff/ani.py#L407-L413). Essentially doing the first layer by hand but it avoids materialzing the big AEV embedding in memory by contracting them with the weights immediately.


I have tried keeping the code as readable as possible but let me know if there's any confusion. I would be curious to see how much of this can be done automatically with Enzyme.

mitkotak:
Also if you do end up comparing with JAX at some point, I was able to run it on my laptop with https://github.com/tillahoffmann/jax-mps

sivasathyaseeelan:
Hi @mitkotak, thank you so much for taking the time to look through this and for the thoughtful suggestions. That's really kind of you. bio-mlff looks fantastic, and it's genuinely helpful to have a well-written reference to learn from, so I'll be reading through it carefully.

Both of your tricks are lovely, and neither is in here yet.

**1. Active-species AEV.** This is a really nice idea. We currently materialize the full-width AEV (all species slots), so for something like water most of the columns just sit at zero. Since our AEV kernels are memory-bound as well, restricting to the active species should help both the kernel work and the NN input width. It also looks orthogonal to the species-batched NN we already do (one `Lux.apply` per element), so hopefully it composes cleanly. I'd love to prototype it on the neighbour-list kernel.

**2. Fused first layer.** I really like this one too, especially for the on-device path, where we currently build the full `(n_atoms × aev_len)` matrix before running the NN. Folding `W_1` into the radial/angular accumulation and summing would avoid ever materializing it. It also lines up nicely with the analytic backward we recently added (the NN VJP already touches the first layer), so I'll look into fusing both directions. Thank you for spelling it out so clearly.

On the Enzyme question, interestingly we ended up going the other way. We had an Enzyme reverse-mode forces path but recently replaced it with a hand-written analytic backward (a VJP through the AEV and the per-element NNs), mainly because Enzyme can't differentiate the KernelAbstractions/Metal kernels. On CPU the analytic path also turned out quite a bit faster for us, so the manual route worked out well here. That said, I'd be very curious to hear about your experience with AD overhead on the JAX side.

Thank you also for the jax-mps pointer, which is really timely. We recently found that TorchANI doesn't have a usable Apple-GPU path (PyTorch-MPS lacks float64 and hits a 5-D scan in the angular AEV), whereas Molly runs the whole model natively on Metal. A JAX-on-Metal cross-check would be a lovely apples-to-apples comparison, so I'll definitely give jax-mps a try.

Thanks again, I really appreciate you sharing these.

mitkotak:
> Enzyme can't differentiate the KernelAbstractions/Metal kernels

Interesting. Do you know why ? (Have a very surface level understanding of Enzyme ecosystem)

sivasathyaseeelan:
> Interesting. Do you know why ? (Have a very surface level understanding of Enzyme ecosystem)

Enzyme works on LLVM IR. It reads the LLVM code your function compiles to and builds the gradient from it. On the CPU this is easy: Julia compiles your code to normal host LLVM, and Enzyme differentiates that. That's what our old CPU forces used, straight through the AEV and the neural nets.

On the GPU it breaks down for two reasons:

1. Enzyme's GPU support is built for CUDA and AMDGPU. Those compile through LLVM in a way Enzyme can hook into. Metal is different: Metal.jl compiles kernels to Apple's own format (AIR / Metal shaders), and Enzyme has no support for that target. So it cannot differentiate a Metal kernel at all.
2. Even on CUDA, a KernelAbstractions kernel is hard for Enzyme because of the atomic scatters, shared memory, and barriers (`@atomic`, `@localmem`, `@synchronize`). Those aren't plain arithmetic, so a gradient doesn't just fall out.

So we wrote the backward pass by hand instead: the NN VJP plus the backward radial and angular AEV kernels. The same kernels run on CPU, Metal, and CUDA. It's exact, and it's actually faster than Enzyme even on the CPU (about 7 to 14x at 1000 atoms).

mitkotak:
> Enzyme because of the atomic scatters, shared memory, and barriers (@atomic, @localmem, @synchronize). Those aren't plain arithmetic, so a gradient doesn't just fall out.

Gotcha. Even if you can convert a scatter into a reduction with padding, so (n_atoms, max_atoms) ? At least that's how I have been staying out of JAX's gather/scatter path wherever possible. Was wondering if a similar rewrite can used incase of Enzyme.

(Sorry if I am hijacking your PR)

src/types.jl
```diff
@@ -23,7 +23,15 @@ export
...
-    ASECalculator
+    ASECalculator,
+    AbstractMLPotential,
+    ANIPotential,
+    compute_aevs,
+    compute_aevs_ka,
+    compute_ani_energy_ka,
+    compute_ani_forces_ka,
+    cosine_cutoff,
+    celu01
```
allexandre97: Do not need to export all of these, the Molly namespace is already quite cluttered. Those methods are internal to the things you have implemented in this PR. Export only things that an end user would have to use frequently in the scripts they would write.

docs/src/examples.md
```diff
@@ -518,6 +518,43 @@ save("polymer_angle.png", f)
 
 There is [an example](https://acesuit.github.io/ACEmd.jl/stable/molly) of using ACE potentials in Molly via [ACEmd.jl](https://github.com/ACEsuit/ACEmd.jl).
 
+## ANI neural network potentials
+
+[`ANIPotential`](@ref) is a native Julia implementation of the [ANI-2x](https://doi.org/10.1021/acs.jctc.0c00121) neural network potential (H, C, N, O, S, F, Cl), with no Python runtime dependency.
+It becomes available when [Lux.jl](https://github.com/LuxDL/Lux.jl) and [HDF5.jl](https://github.com/JuliaIO/HDF5.jl) are loaded; loading [Enzyme.jl](https://github.com/EnzymeAD/Enzyme.jl) in addition gives single-pass reverse-mode AD forces (otherwise forces fall back to finite differences).
```
allexandre97: Is this still the case? Or is the fallback now to analytical forces?

ext/molly_lux_ka.jl
```diff
@@ -0,0 +1,945 @@
...
+    n_eta_R    :: Int,
+    n_shf_R    :: Int,
+    n_eta_A    :: Int,
+    n_shf_r    :: Int,
+    n_th       :: Int,
+    split      :: Int,    # boundary between radial/angular in AEV row
+)
+    atom_i = @index(Global, Linear)
+
+    if atom_i <= n_atoms
```
allexandre97: Why is this guard needed?

ext/molly_lux_ka.jl
```diff
@@ -0,0 +1,945 @@
...
+
+# Convert a Molly NeighborList (flat half-pairs (i,j,special)) into a host CSR
+# (offsets, symmetrised indices). This is the production path: it *consumes the
+# neighbours produced by the system's neighbour finder* (CellListMapNeighborFinder on
+# CPU, DistanceNeighborFinder on GPU) rather than building its own list. `neighbors.list`
+# may be device-resident (GPU finder) — `Array(...)` brings the pairs to the host for the
+# O(total_pairs) counting sort; offsets/indices are then uploaded to the compute backend.
+function nl_to_csr(neighbors, n_atoms::Int)
+    npairs = length(neighbors)                 # = neighbors.n
+    host   = Array(neighbors.list)             # host copy (no-op if already on CPU)
```
allexandre97: Use `Molly.from_device()` and `Molly.to_device()` to move things between CPU / GPU

ext/molly_lux_ka.jl
```diff
@@ -0,0 +1,945 @@
...
+    ps_dev_all, _ = ani_nn_dev_params(pot, dev)               # cached on-device weights
+    E = 0.0
+    for ens_i in 1:n_ens
+        for s in 1:n_species
+            g = groups[s]; isempty(g) && continue
+            sym   = Symbol(idx_to_elem[s])
+            batch = permutedims(aevs[idx_dev[s], :])          # (aev_len, n_s)
+            ps_d  = getfield(ps_dev_all[ens_i], sym)
+            E_s, dEdbatch = nn_energy_and_grad(getfield(pot.model, sym), ps_d, batch)
+            E += Float64(E_s) + Float64(pot.self_energies[s]) * length(g)
```
allexandre97: The model has weights in `Float32` but here you upcast the energy to `Float64`. Is this really necessary? If so, why?

ext/molly_lux_ka.jl
```diff
@@ -0,0 +1,945 @@
...
+        KernelAbstractions.synchronize(ka_backend)
+    end
+
+    return dcoords
+end
+
+# End-to-end on-device ANI forces (eV/Å). Forward AEV → NN VJP (∂E/∂G) → backward AEV kernels
+# (∂E/∂r) → F = -∂E/∂r · Ha→eV, ensemble-averaged. This is the single ANI forces path — the
+# AtomsCalculators.forces! method below dispatches here for both CPU (KA CPU backend) and GPU
+# (Metal/CUDA) systems. The former finite-difference and Enzyme reverse-mode paths are retired:
```
allexandre97: If paths are retired remove mentions to them in production comments.

ext/molly_lux_ka.jl
```diff
@@ -0,0 +1,945 @@
...
+# GPU array → Metal/CUDA), computes forces via the analytic backward, and accumulates into `fs`
+# in the system's units. Replaces the former finite-difference (MollyLuxExt) and Enzyme
+# reverse-mode (MollyLuxEnzymeExt) methods. Requires KernelAbstractions in addition to Lux/HDF5.
+function AtomsCalculators.forces!(fs, sys::System{D, AT, T}, inter::ANIPotential;
+                                  kwargs...) where {D, AT, T}
+    nbrs    = get(kwargs, :neighbors, nothing)
+    n_sp    = length(inter.species_map)
+    sp      = Int32[inter.species_map[ad.element] for ad in sys.atoms_data]
+    coords  = Molly.ustrip_vec.(u"Å", sys.coords)       # Å, unitless; stays on coords' device
+    species = Molly.to_device(sp, AT)
```
allexandre97: Is there a reason why you have to do this every time the forces are being called? If not, you should move the species to device just one and avoid repeating the operation unnecessarily

test/ml_potentials.jl
```diff
@@ -0,0 +1,897 @@
...
+        random_velocities!(sys, 300.0u"K"; rng=MersenneTwister(42))
+
+        E0     = total_energy(sys)
+        simulate!(sys, VelocityVerlet(dt=0.1u"fs"), 50)
+        E_final = total_energy(sys)
+
+        # NVE: total energy should be approximately conserved
+        drift = abs(ustrip(u"eV", E_final - E0)) / abs(ustrip(u"eV", E0))
+        @test drift < 0.05   # < 5% relative drift over 50 steps at 0.1 fs
+        @test norm(sys.coords[1] - sys.coords[2]) > 0.5u"Å"  # no explosion
```
allexandre97, line 343: Is this realistic? Energy conservation for 50 steps at 0.1 fs integration  timestep is not a hard test to pass. Maybe you could add some benchmarks

test/ml_potentials.jl
```diff
@@ -0,0 +1,897 @@
...
+
+        # Timing (second call — buffers warm, neighbor list computed)
+        t = @elapsed potential_energy(sys)
+        alloc = @allocated potential_energy(sys)
+        println("6mrr first $n atoms: E = $(round(typeof(E), E, sigdigits=8))")
+        println("  time (warm):  $(round(t*1000, sigdigits=3)) ms")
+        println("  allocs (warm): $alloc bytes")
+    end
+end
+
```
allexandre97: If you test all atoms in the following test you can remove the 50 atoms test maybe?

docs/src/api.md
```diff
@@ -8,6 +8,7 @@ Package extensions are used in order to reduce the number of dependencies:
 - To use [`visualize`](@ref), call `using GLMakie`.
 - To use [`ASECalculator`](@ref), call `using PythonCall`.
 - To use [`rdf`](@ref), call `using KernelDensity`.
+- To use [`ANIPotential`](@ref) energy, call `using Lux, HDF5`. Forces use a single analytic path (`AtomsCalculators.forces!` → [`compute_ani_forces_ka`](@ref)) that runs on CPU or GPU, so forces — and the on-device functions [`compute_aevs_ka`](@ref) / [`compute_ani_energy_ka`](@ref) / [`compute_ani_forces_ka`](@ref) — additionally need `using KernelAbstractions`. (There is no Enzyme or finite-difference forces path.)
```
allexandre97: No need to mention there is no Enzyme or finite-diffs path if the only production code will be the analytical one.

jgreener64:
Thanks for the input @mitkotak.

My understanding is that KernelAbstractions kernels should work with Enzyme on all backends. In practice bugs and compile times can be an issue. If we have an analytical derivative path, though, that will likely be faster and make better use of memory.

allexandre97 (commented):
Overall looks good. I still have some doubts as why you support the path that does not use `Molly.vector` to calculate the distance between two particles. No matter if systems are periodic, or not; neither small nor big, that should work and is more idiomatic within the Molly ecosystem.

Also, since you have implemented the analytical force path, make sure you clean the code and docs of stale references to FD and AD paths. What you could have is a test comparing analytical forces to AD forces, to ensure that the implementation is correct.

sivasathyaseeelan:
Hi @mitkotak, following up with what I found implementing your suggestions.

**Active-species AEV (your first suggestion).** I built it: the on-device energy and forces now materialise only the columns for species present and slice the first NN layer to match. It is exact (energy bit-identical to the full-width path, forces match to about 1e-6). Measured speedup on Apple Metal, reduced vs full-width:

| N atoms | water (2 species) energy | water forces | 6mrr (7 species) energy | 6mrr forces |
|--------|--------|--------|--------|--------|
| 1000   | 1.01x | 1.01x | 1.00x | 1.00x |
| 2000   | 1.04x | 1.04x | 1.01x | 1.01x |
| 4000   | 1.07x | 1.06x | 1.02x | 1.02x |
| 8000   | **1.12x** | **1.11x** | 1.03x | 1.03x |

At the full system the absolute full-width vs reduced times are:
- **6mrr, 15,954 atoms (all 7 species):** energy 156.0 to 150.8 ms (1.03x), forces 410.6 to 402.2 ms (1.02x). Almost nothing, since every species is present.
- **water, 15,490 atoms (2 species):** energy 131.2 to 124.3 ms (1.05x), forces 342.4 to 330.7 ms (1.04x).

So up to about 1.12x for water at mid-size (the AEV narrows from 1008 to 128 columns, roughly 8x), tapering to 1.03 to 1.05x at the full system, and essentially nothing for the full 7-species protein.

**Why the win is small for us.** Our AEV kernel is a neighbour-list kernel: each thread loops over an atom's actual neighbours, which are always present species, so it already only computes and writes their columns. The full-width AEV just carried extra zero columns the kernel never spent time on. That differs from a dense framework like your JAX version, where the AEV is computed over all species slots with masking, so restricting to active species removes real compute. For us the reduction only trims already-zero output columns and the first-layer matmul.

**Fused first layer (your second suggestion).** I profiled before building it. On a water system at 8000 atoms on Metal the breakdown is:

| component | time | share |
|--------|--------|--------|
| AEV kernel | 29.5 ms | 85% |
| first-layer matmul (what fusion removes) | 1.7 ms | 5% |
| total energy | 34.7 ms | |

So fusing the first layer has a hard ceiling of about 1.05x even if it were free, and in practice it would move that 5% into the already-dominant AEV kernel as many short scattered updates, which would likely be slower. So I did not pursue it.

**For reference, the numbers these sit on top of** (full 6mrr protein, 15,954 atoms, this machine, single ensemble member): Molly on Metal does energy in about 149 ms and forces in about 396 ms, versus TorchANI CPU at about 2653 ms energy and 3235 ms forces (roughly 18x and 8x). On a water slice of the same size (15,490 atoms) Molly Metal is about 127 ms energy and 330 ms forces, about 20x and 10x faster than TorchANI CPU.

**Decision.** I have kept the active-species and fused-layer changes out of the PR for now to keep it focused on the core energy/forces path for review and merge, and I am treating them as possible follow-ups. The most useful thing your suggestions gave me was the profiling that pinpointed the real bottleneck, the AEV kernel itself (that 85%), which is where I will look next for a real speedup. They also surfaced a latent caching bug I have since fixed.

Thank you again for the ideas and the clear reference code, really appreciate it.

src/types.jl
sivasathyaseeelan: Done. Exports are now just `ANIPotential` and `AbstractMLPotential`. The rest (`compute_aevs`, `compute_aevs_ka`, `compute_ani_energy_ka`, `compute_ani_forces_ka`, `cosine_cutoff`, `celu01`) are unexported; the extension imports `celu01`/`cosine_cutoff` explicitly so they stay out of the public namespace.

docs/src/examples.md
sivasathyaseeelan: Updated. Forces are now a single analytic path, available with just `using Lux, HDF5` (KernelAbstractions is a core Molly dependency). Removed the Enzyme / finite-difference wording from the example.

ext/molly_lux_ka.jl
sivasathyaseeelan: Added a comment: KernelAbstractions rounds the launch up to a whole number of workgroups, so the last group can spawn threads with `atom_i > n_atoms`; the guard makes them no-op.

ext/molly_lux_ka.jl
sivasathyaseeelan: Done, `nl_to_csr` now uses `Molly.from_device(neighbors.list)` instead of `Array(...)`.

ext/molly_lux_ka.jl
sivasathyaseeelan: Intentional, and now commented. The summands stay Float32 (the NN outputs); only the reduction is done in Float64. Summing ~1e4 per-atom terms plus large per-species self-energies in Float32 loses precision, so a Float32 running sum drifts; the Float64 accumulation keeps the full-protein energy within <0.001% of the TorchANI reference. It does not change the model precision.

ext/molly_lux_ka.jl
sivasathyaseeelan: Removed the retired-paths wording from the production comment.

ext/molly_lux_ka.jl
sivasathyaseeelan: The species vector is derived from `sys.atoms_data`, which is passed in fresh on every `forces!` call, so there is no persistent place to cache it without keying on the system (and it would go stale if the atoms change). In practice the cost is negligible: it is a small `Int32` vector plus one upload, versus the AEV kernel which is about 85% of the runtime (profiled separately), so per call this is microseconds against hundreds of milliseconds. I have left it per-call for now, but happy to add a system-keyed cache if you would prefer.

test/ml_potentials.jl
sivasathyaseeelan: Agreed, this in-suite test is only a fast smoke check. The stringent conservation test lives in `benchmark/ani_trajectory.jl`: a 300-atom 6mrr slice over 3000 steps at 0.5 fs conserves total energy to |ΔE|/|E0| ≈ 1.5e-7. I added a comment on the test pointing to it.

test/ml_potentials.jl
sivasathyaseeelan: Done, removed the 50-atom test. The full-protein test (all 15,954 atoms, cross-validated against TorchANI) exercises the same large-system path and subsumes it.

docs/src/api.md
sivasathyaseeelan: Removed. api.md no longer mentions the Enzyme / finite-difference path.

mitkotak:
Thanks for looking into this. Yup AEV being the bottleneck checks out with what I see on GPUs. Be careful about not overfitting your intuitions too much on Metal vs NVIDIA/AMD GPU since the ratio of compute/memory differs [widely](https://hazyresearch.stanford.edu/blog/2024-11-28-tk-mlx). Though looking for locality opportunities in the AEV kernel should give you net gain on all.

Curious about why you kept the AEV and neighborlist in one kernel since I would imagine that keeping them separate would let you do a fast topk for the neighbor without the compute of the AEV stalling you. Not familiar with Metal tricks but atleast on GPUs combining them together has always blown away my register budgets.

Also for baselines if not too much of a hassle can you compare with the ANI-JAX I have both on Metal and CPU (though fair to compare with the original reference implementation as well). It should be a much stronger baseline and profiling how JAX is doing things might help you see how your kernels can be better.

sivasathyaseeelan:
Thanks @mitkotak, this is helpful.

**On Metal vs NVIDIA/AMD.** Fair point, and thanks for the ThunderKittens/MLX link. I will keep the optimisations portable (locality and occupancy that help everywhere) rather than tuning to Metal's compute/memory ratio, and re-profile on CUDA before drawing conclusions once I have an NVIDIA GPU access.

**On the AEV and the neighbour list.** They are already separate. Molly's neighbour finder (`DistanceNeighborFinder`, or `CellListMapNeighborFinder` for large systems) builds the neighbour list, which I hand to the AEV kernel as CSR (offsets + flattened indices); the kernel only consumes it. The all-pairs kernel that scans every atom is just a fallback for systems with no finder attached, so neighbour selection never stalls on AEV compute.

Where your register point does bite is inside the AEV kernel itself: it is one thread per atom carrying the full radial + angular accumulation, which is heavy and consistent with it being the bottleneck. On Metal I already have a write-reduced variant (one workgroup per atom, accumulate the AEV row in threadgroup memory, one coalesced global write) that cuts global traffic. Splitting the per-atom kernel further (a distance/cutoff pass, then radial and angular as separate passes) for occupancy is exactly the locality work I want to try next, and per your first point I will confirm it helps on CUDA too, not just Metal.

**On an ANI-JAX baseline.** Happy to, bio-mlff is a stronger comparison than TorchANI and profiling how JAX schedules things should be useful. I will add it on CPU for sure, and attempt Metal via jax-metal / the jax-mps route (that may hit the same unsupported-op limits I saw with TorchANI on MPS, but I will try). I will report back with the numbers alongside the TorchANI reference.

Thanks again for the detailed feedback.

sivasathyaseeelan:
Redid the CPU comparison at matched thread counts (1 and 8 threads) so it is apples-to-apples, since the three implementations thread very differently. M3, 6mrr slices, min-of-N. Metal is thread-independent so it is shown once.

**Energy (ms):**

| N atoms | Molly t1 | Molly t8 | TorchANI t1 | TorchANI t8 | bio-mlff t1 | bio-mlff t8 | Molly Metal |
|---------|----------|----------|-------------|-------------|-------------|-------------|-------------|
| 1000    | 96.5     | 28.4     | 47.0        | 29.3        | 45.3        | 45.5        | 70.1        |
| 8000    | 397.5    | 188.5    | 774.1       | 498.1       | 596.3       | 591.3       | 84.0        |
| 15,954  | 1546     | **437.8**| 3354        | 2394        | 2417        | 2671        | **145.6**   |

**Forces (ms):**

| N atoms | Molly t1 | Molly t8 | TorchANI t1 | TorchANI t8 | bio-mlff t1 | bio-mlff t8 | Molly Metal |
|---------|----------|----------|-------------|-------------|-------------|-------------|-------------|
| 1000    | 371.9    | 130.2    | 94.3        | 56.2        | 83.2        | 82.5        | 179.9       |
| 8000    | 1386     | 646.1    | 1035        | 633.1       | 1347        | 1080        | 211.4       |
| 15,954  | 5027     | **1322** | 4543        | 2870        | 6957        | 7733        | **389**     |

Threading behaviour, which is the interesting part:
- Molly scales well with threads: energy ~3.5x from t1 to t8 (1546 to 438 ms at 16k), forces ~3.8x (5027 to 1322).
- TorchANI scales modestly (~1.4x energy, ~1.6x forces).
- bio-mlff (JAX/XLA on CPU) is essentially thread-insensitive here: the XLA CPU threadpool did not respond to `intra_op_parallelism_threads` (t1 ~ t8, and a touch slower at 16k from noise). You would know better than me whether there is a better knob for XLA-CPU parallelism.
- At t8, Molly is the fastest CPU implementation at scale for both energy (438 vs 2394 vs 2671) and forces (1322 vs 2870 vs 7733). At t1, Molly leads on energy; TorchANI edges it on forces at 16k (4543 vs 5027).
- Metal is far ahead of any CPU configuration (146 ms energy / 389 ms forces at the full protein).

(bio-mlff had to run on CPU with `JAX_PLATFORMS=cpu`, since with `jax-metal` installed JAX defaults to the Metal device and then fails to compile the ANI graph, as noted above.)

jgreener64:
Nice one. I think benchmarking on CUDA is important here, both for comparing to other software and getting the right tradeoffs in the kernels.

I can also benchmark on various NVIDIA GPUs (4090, A6000, 2080ti) if you want wider coverage, let me know what script to run.

mitkotak:
> (bio-mlff had to run on CPU with JAX_PLATFORMS=cpu, since with jax-metal installed JAX defaults to the Metal device and then fails to compile the ANI graph, as noted above.)

What issue were you seeing with jax-mps ?

sivasathyaseeelan:
Good question, and it turned out well. The failure I described was with Apple's `jax-metal` plugin (0.1.1), not your `jax-mps`. `jax-metal` registers a Metal device but its PJRT plugin is pinned to an old jaxlib and cannot deserialize the StableHLO that `jaxlib` 0.10.2 (what bio-mlff pulls) emits (`unknown attribute code: 22`, StableHLO v1.13.7).

`jax-mps` works. On Python 3.13 with jax/jaxlib 0.10.2 and jax-mps 0.10.10, bio-mlff runs on the Apple GPU via MLX with `JAX_PLATFORMS=mps`, so there is now a real Metal-to-Metal comparison. M3, 6mrr slices:

Energy (Metal, ms):

| N atoms | Molly Metal | bio-mlff jax-mps |
|---------|-------------|------------------|
| 1000    | 70.2        | 13.2             |
| 2000    | 75.5        | 32.1             |
| 5000    | 78.5        | 124.8            |
| 8000    | 84.0        | 269.8            |
| 15,954  | 145.9       | 1246             |

Forces (Metal, ms):

| N atoms | Molly Metal | bio-mlff jax-mps |
|---------|-------------|------------------|
| 1000    | 181.1       | 32.8             |
| 2000    | 193.5       | 66.5             |
| 8000    | 213.1       | 447.5            |
| 15,954  | 394         | 1790             |

MLX has much lower launch overhead, so jax-mps is well ahead at small N (energy 13 vs 70 ms at 1000). But Molly's Metal kernels are nearly flat with system size (70 to 146 ms energy from 1000 to 16k) while jax-mps grows steeply, so Molly pulls ahead past a few thousand atoms and is about 8.5x on energy and 4.5x on forces at the full 6mrr protein. Molly's roughly 70 ms fixed small-N overhead is the obvious thing for me to chase next, and MLX's dispatch efficiency is a good thing to learn from.

Thanks for pushing on this, jax-mps is a much better Metal baseline than I expected to get.

sivasathyaseeelan:
Thanks Dr. Joe(@jgreener64)! Agreed, CUDA is the important next step, both for the software comparison and for getting the kernel tradeoffs right.

Tim (@maleadt) has set me up with GPU access on cyclops, so I will write the CUDA benchmark scripts and run them there myself first, then share the exact steps (and the numbers) with you. Once that is working I would love to take you up on the wider coverage across the 4090 / A6000 / 2080ti, since seeing the compute/memory tradeoffs across a few NVIDIA generations would be really useful for tuning the kernels.

Will follow up with the scripts and results.

jgreener64:
Superseded by #281.

PR #262, opened 2026-06-09
Remove unused NxN matrix in OBC calculations

davide-grheco:
While looking through the OBC implicit solvent implementation I noticed that born_radii_and_grad allocates a zeros N×N matrix on every step that is never written to. This PR removes it and returns nothing instead, dispatching to a new CPU and GPU code path that skips the unused term.

I am aware that this duplicates quite a bit of code, but I wasn't sure how you'd prefer to handle that.
There is a small benchmark I just used to check if what I did actually has an effect, it can be removed if it is not necessary.
I don't know how much this actually affects real simulation performance, possibly the difference will be negligible.

Before
```
n_atoms   alloc (bytes)   median (ms)
------------------------------------
50        51712           0.12
100       181312          0.54
200       682528          2.50
400       2639840         10.21
```

After
```
n_atoms   alloc (bytes)   median (ms)
------------------------------------
50        11472           0.11
100       21136           0.53
200       38840           2.11
400       75712           8.76
```

### DIFF
--- a/src/interactions/implicit_solvent.jl
+++ b/src/interactions/implicit_solvent.jl
@@ -641,6 +641,11 @@ end
 get_i1(x) = @inbounds x[1]
 get_i2(x) = @inbounds x[2]
 
+struct EmptyIgrads end
+
+Base.getindex(::EmptyIgrads, i, j) = EmptyIgrads()
+Base.:-(x, ::EmptyIgrads) = x
+
 function born_radii_sum(or, offset, I, α, β, γ)
     radius = or + offset
     ψ = I * or
@@ -665,13 +670,11 @@ function born_radii_and_grad(inter::ImplicitSolventOBC{T}, coords, boundary) whe
         end
         Is[i] = I
     end
-    I_grads = zeros(eltype(Is), length(Is), length(Is)) ./ unit(inter.dist_cutoff)
-
     Bs_B_grads = born_radii_sum.(inter.offset_radii, inter.offset, Is,
                                  inter.α, inter.β, inter.γ)
     Bs      = get_i1.(Bs_B_grads)
     B_grads = get_i2.(Bs_B_grads)
-    return Bs, B_grads, I_grads
+    return Bs, B_grads, EmptyIgrads()
 end
 
 function born_radii_and_grad(inter::ImplicitSolventOBC, coords::AbstractGPUArray, boundary)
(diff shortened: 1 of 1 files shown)


jgreener64:
Yes this is worth fixing. I don't think you need to duplicate the code though. Instead of using `nothing` you could have a new struct `EmptyIgrads` that wraps a zero unitful number. If you define `getindex` for this struct to return the zero number then I think the rest of the code works as-is.

davide-grheco:
Hello, this is indeed a better idea. Rather than returning a zero unitful number I decided to define Base.- to return the input. This works as of now as it is only used in `de = bi * (t3 - ig) * rinv` and it makes the struct definition a bit simpler as it would otherwise have to be parametric on the unit type.
If you prefer I can indeed make it parametric and return unitful zero, which would simplify the use of the struct in cases in which you need to multiply or sum it to something else without having to define all those functions.

I did not remove the duplicated GPU kernel. I am not particularly experienced in GPU operations using Julia. I guess duplication could be removed by making the @Const(I_grads) in the signature not Const, but I do not know what that change would lead to.

benchmark/obc_alloc_benchmark.jl
```diff
@@ -0,0 +1,35 @@
+using Molly
```
jgreener64: I don't think we need this file.

src/interactions/implicit_solvent.jl
```diff
@@ -641,6 +641,13 @@ end
 get_i1(x) = @inbounds x[1]
 get_i2(x) = @inbounds x[2]
 
+struct EmptyIgrads end
+
+Base.getindex(::EmptyIgrads, i, j) = EmptyIgrads()
+Base.:-(x, ::EmptyIgrads) = x
+
+get_ig(I_grads, i, j) = @inbounds I_grads[i, j]
```
jgreener64: This function is no longer needed.

src/interactions/implicit_solvent.jl
```diff
@@ -973,6 +979,23 @@ function gbsa_force_2_gpu(coords::AbstractArray{SVector{D, C}}, boundary, dist_c
     return fs_mat
 end
 
+function gbsa_force_2_gpu(coords::AbstractArray{SVector{D, C}}, boundary, dist_cutoff, Bs, B_grads,
```
jgreener64: I'd prefer to remove the `@Const` annotation than duplicate the kernels, it doesn't change much.

jgreener64 (commented):
Looks good, just a few things.

PR #264, opened 2026-06-12
Constraints contribution to virial

allexandre97:
# Constraint virial support across for supported constraints and several integrators

Molly was lacking the constraint contribution to the virial, which made impossible the use of virial-derived barostats if constraints were to be used. This PR addresses mainly that issue and some other stuff that was derivative of it.

 ## Main changes:

- Adds `ConstraintApplicationContext` plumbing so position and velocity constraint application can optionally accumulate virial contributions. This is needed since the form of the contribution to the virial depends on the type of constraint being applied.
- Implements constraint virial accumulation for SHAKE/RATTLE and LINCS on CPU and GPU paths.
- Adds constraint virial buffers, validity tracking, and merge logic so constraint virials are combined with interaction virials before pressure/virial logging or pressure coupling.
- Updates `VelocityVerlet`, `DPDVelocityVerlet`, `Verlet`, `StormerVerlet`, and `Langevin` to request virials based on both couplers and virial/pressure loggers, and to pass the right constraint context during integration. `StormerVerlet` documented clearly stating that it CAN use constraints, but it is not suitable for use with coupling algorithms.
- Adjusts barostat coupling to use a pressure kinetic tensor captured after final velocity constraints when needed.
- Updates pressure/virial loggers to use valid cached total virials and to error for constrained systems when a valid total virial is unavailable.
- Makes constrained atoms also contribute to hydrogen mass repartitioning.
- Moves hydrogen mass repartitioning before constraint generation so constrained atom masses/inverse masses reflect repartitioned masses.
- Adds CUDA compatibility updates for `CUDA.CUDACore` naming.
- Adds tests for new behaviour.


ext/MollyCUDAExt.jl
```diff
@@ -127,7 +134,7 @@ effective_force_maxregs_override(config::Molly.CUDALaunchConfig) =
 
 function autotune_key(sys::System{D, <:CuArray}, pairwise_inters, force_maxregs_override) where D
     dev = CUDA.device()
-    sm_count = CUDA.attribute(dev, CUDA.CU_DEVICE_ATTRIBUTE_MULTIPROCESSOR_COUNT)
+    sm_count = CUDA.attribute(dev, CUDACore.CU_DEVICE_ATTRIBUTE_MULTIPROCESSOR_COUNT)
```
jgreener64: I think with recent CUDA.jl tags this file can now remain unchanged bar this line. We can drop support for CUDA.jl v4.

src/constraints/constraints.jl
```diff
@@ -53,6 +53,31 @@ struct AngleConstraint{D, I}
     end
 end
 
+abstract type ConstraintApplicationKind end
```
jgreener64: Perhaps `AbstractConstraintType`, `PositionConstraint` and `VelocityConstraint`?

src/force.jl
```diff
@@ -162,32 +162,143 @@ Base.:+(x::SpecificForce3Atoms, y::SpecificForce3Atoms) = SpecificForce3Atoms(x.
...
+const INVALID_BUFFER_STEP = -1
+
+mutable struct BufferValidity
+    interaction_virial_step::Int
+    constraint_virial_step::Int
+    total_virial_step::Int
+    pressure_step::Int
+end
+
+BufferValidity() = BufferValidity(
```
jgreener64, line 203: Add a very brief comment about what this is.

src/force.jl
```diff
@@ -162,32 +162,143 @@ Base.:+(x::SpecificForce3Atoms, y::SpecificForce3Atoms) = SpecificForce3Atoms(x.
...
+    INVALID_BUFFER_STEP,
+    INVALID_BUFFER_STEP,
+    INVALID_BUFFER_STEP,
+    INVALID_BUFFER_STEP,
+)
+
+invalidate_interaction_virial!(v::BufferValidity) = (v.interaction_virial_step = INVALID_BUFFER_STEP; v)
```
jgreener64: I prefer multi-line functions here.

src/simulators.jl
```diff
@@ -230,6 +230,154 @@ function VelocityVerlet(; dt, coupling=nothing, remove_CM_motion=1)
...
+    return true, smin
+end
+
+same_logger_observable(observable, name::Symbol) =
+    isdefined(@__MODULE__, name) && observable === getfield(@__MODULE__, name)
+
+function logger_virial_interval(logger)
+    if hasproperty(logger, :observable) && hasproperty(logger, :n_steps)
+        observable = getproperty(logger, :observable)
+        if same_logger_observable(observable, :virial_wrapper) ||
```
jgreener64: Not sure I like this inspection of structs, does it work with user-defined loggers? It may be better to define functions on loggers that a user can extend.

src/simulators.jl
```diff
@@ -230,6 +230,154 @@ function VelocityVerlet(; dt, coupling=nothing, remove_CM_motion=1)
...
+
+function merge_constraint_virial_if_needed!(buffers, sys, step_n::Integer,
+                                            needs_virial::Bool)
+    if needs_virial && length(sys.constraints) > 0
+        merge_constraint_virial!(buffers, sys, step_n)
+    end
+    return buffers
+end
+
+function merge_zero_constraint_virial_if_needed!(buffers, sys, step_n::Integer,
```
jgreener64: Shorter name.

src/simulators.jl
```diff
@@ -373,48 +551,73 @@ end
...
         sys.coords .= wrap_coords.(sys.coords, (sys.boundary,))
         place_virtual_sites!(sys)
 
-        # Velocity prediction for dissipative force evaluation
-        sys.velocities .= velocities_t .+ accels_t .* λ_dt
+        velocities_half .= sys.velocities
 
-        # Compute forces at new positions with predicted velocities
-        forces!(forces_t_dt, sys, neighbors, buffers, Val(needs_vir), step_n; n_threads=n_threads)
+        # DPD dissipative forces depend on velocity. Temporarily use the
```
jgreener64, line 671: Is this done like this elsewhere?

src/simulators.jl
```diff
@@ -531,20 +749,53 @@ end
 
 The Störmer-Verlet integrator.
 
-The velocity calculation is accurate to O(dt).
-
-Does not currently work with coupling methods that alter the velocity.
-Does not currently remove the center of mass motion.
+This is a coordinate-history integrator: after the first step, coordinates are
```
jgreener64: Very long for a docstring, move most to comments.

src/simulators.jl
```diff
@@ -531,20 +749,53 @@ end
...
-- `coupling::C=nothing`: the coupling which applies during the simulation.
+- `coupling::C=nothing`: no coupling. Non-empty coupling methods throw an
+    `ArgumentError` when the simulation starts.
 """
 @kwdef struct StormerVerlet{T, C}
     dt::T
     coupling::C = nothing
 end
 
+stormer_verlet_coupling_supported(::Nothing) = true
```
jgreener64: Is this used anywhere? Could just remove since we document that it doesn't work with coupling methods that alter the velocity.

test/constraints.jl
```diff
@@ -491,6 +491,121 @@ function lincs_center_of_mass(x, masses_val)
     return sum(masses_val[i] * x[i] for i in eachindex(x)) / total
 end
 
+function make_lincs_no_unit_buffer_system(coords, masses_val)
```
jgreener64, line 461: This is 1000 extra lines of tests, trim them if possible so that each one tests something different.

src/simulators.jl
```diff
@@ -261,37 +412,59 @@ end
...
+            apply_velocity_constraints!(sys; context=vel_context, n_threads=n_threads)
+            merge_constraint_virial_if_needed!(buffers, sys, step_n, needs_vir_step)
+            if needs_vir_step
+                kinetic_energy_tensor!(buffers.kin_tensor, sys)
+                pressure_kin_tensor = copy(buffers.kin_tensor)
+            end
+        end
 
         if !iszero(sim.remove_CM_motion) && step_n % sim.remove_CM_motion == 0
             remove_CM_motion!(sys)
```
jgreener64, line 531: This alters the velocities, is this okay with your stored kinetic energy tensor?

src/simulators.jl
```diff
@@ -242,15 +390,18 @@ end
...
     sys.coords .= wrap_coords.(sys.coords, (sys.boundary,))
     place_virtual_sites!(sys)
     !iszero(sim.remove_CM_motion) && remove_CM_motion!(sys)
     neighbors = find_neighbors(sys, sys.neighbor_finder, nothing, init_step, true;
                                n_threads=n_threads)
     forces_t, forces_t_dt = zero_forces(sys), zero_forces(sys)
     buffers = init_buffers!(sys, n_threads)
-    forces!(forces_t, sys, neighbors, buffers, Val(needs_vir), init_step; n_threads=n_threads)
+    needs_vir_init = needs_virial_on_step(needs_vir, needs_vir_steps, init_step)
+    forces!(forces_t, sys, neighbors, buffers, Val(needs_vir_init), init_step;
```
jgreener64: Does anything special need to be done for step zero? Is the virial accurate here?

src/constraints/shake.jl
```diff
@@ -756,7 +999,12 @@ end
 function apply_position_constraints!(sys::System,
                                      ca::SHAKE_RATTLE,
                                      r_pre_unconstrained_update;
+                                     context=nothing,
                                      kwargs...)
+    context = isnothing(context) ? default_shake_position_constraint_context() : context
+    validate_shake_position_virial_context(context)
+    coords_before = context.needs_virial ? copy(sys.coords) : nothing
```
jgreener64: Is this allocation okay? Could use a buffer. Also for velocities.

jgreener64 (commented):
Looks good from what I can understand, but this is a complex change. I trust you to have confidence in each line even if it was written by AI.

As I understand the `recompute_forces` order issue still needs to be thought about.

jgreener64:
We could use finite differencing to test our instantaneous `virial` call, with and without constraints, against the derivative of the potential energy with respect to the box.

src/constraints/constraints.jl
allexandre97: These structs describe the application of the constraint rather than the constraint itself, but I do agree that the naming is a bit ambiguous. I will change them to:

`AbstractConstraintApplication`
`PositionConstraintApplication`
`VelocityConstraintApplication`

src/force.jl
allexandre97: This basically tracks which step the cached virial / pressure buffers are valid for. This is need, for example, for constrained systems with  coupling. After coupling, forces may be recomputed, but recompute_forces_after_coupling! intentionally does not recompute a constrained-system total virial.

src/force.jl
allexandre97: Changed

src/simulators.jl
allexandre97: You're right, this is suboptimal. I moved these functions to `src/loggers.jl` now instead they take a logger as an argument and dispatch on its type. A user can extend and overload the functions depending on their needs, that way we do not need to care about what fields the loggers are carrying.

src/simulators.jl
allexandre97: Changed to `merge_zero_constraint_virial!`

src/simulators.jl
allexandre97: This is only done for `DPDVelocityVerlet`, but an equivalent pattern is already present in the master branch: temporarily set `sys.velocities` to the predicted velocity before calling `forces!`, because DPD forces depend on velocity and `forces!` reads velocities from the system. The state is restored before coupling / logging. The current branch keeps the same idea, but expressed relative to the half-step velocity introduced by the constraints refactor.

Maybe we could, in principle, avoid this by telling `forces!` to accept an optional velocity array override, but that would be a bigger refactor.

src/simulators.jl
allexandre97: Trimmed and moved into comments

src/simulators.jl
allexandre97: Changed these bits and pieces such that only `coupling=nothing` is allowed for `StormerVerlet`. Any other argument will throw `ArgumentError`

test/constraints.jl
allexandre97: Trimmed tests

src/simulators.jl
allexandre97: You're right, this was technically wrong, I suppose it wasn't being problematic because COM removal did not happen every step. I changed where COM removal is called in `VelocityVerlet` and `DPDVelocityVerlet`. It's still called before couplers and loggers and after applying constraints, but now also before the kinetic tensor snapshot used by pressure coupling.

src/simulators.jl
allexandre97: Following what we discussed, the zero-step issue is solved by takihng a first small (0.5 fs tstep) fake verlet-like half step, computing constraints and virial there, and reverting back to initial state. 

`pressure`  and `virial` methods extended to take an integrator to apply the same logic so that we can get the constraint contribution. If no integrator is provided, defaults to verlet-like small step.

src/constraints/shake.jl
allexandre97: Added buffers to avoid allocations

ext/MollyCUDAExt.jl
allexandre97: I have tried updating CUDA to the latest version and I still need the `const CUDA_CORE = isdefined(CUDA, :CUDACore) ? CUDACore : CUDA` guard. What can change is using `DEVICE_ATTRIBUTE_MULTIPROCESSOR_COUNT` instead, which is public.

The reason I need the guard is because of namespace conflict. CUDA makes `Const` public but it conflicts with another definition under the same name. Therefore, it needs annotation, and I think in CUDA 5 it is `CUDA` but in v6 is `CUDACore`

docs/src/documentation.md
```diff
@@ -1403,17 +1408,30 @@ Base.values(logger::MyLogger) = logger.history
 Then, define the logging function that is called every step by the simulator:
 ```julia
 function Molly.log_property!(logger::MyLogger,
-                                sys,
-                                neighbors,
-                                step_n;
-                                n_threads=Threads.nthreads(),
-                                kwargs...)
+                             sys,
```
jgreener64, line 1411: If the order was `sys`/`neighbors`/`step_n`/`buffers` then this would match the energy and force functions.

docs/src/documentation.md
```diff
@@ -1434,14 +1452,14 @@ Many times, a logger will just record an observation to an `Array` containing a
 For this purpose, you can use the [`GeneralObservableLogger`](@ref) without defining a custom logging function.
 Define your observation function as
 ```julia
-function my_observable(sys::System, neighbors, step_n; n_threads::Integer, kwargs...)
+function my_observable(sys::System, buffers, neighbors, step_n; n_threads::Integer, kwargs...)
```
jgreener64: Same argument as before about argument order.

src/constraints/constraints.jl
```diff
@@ -53,6 +53,51 @@ struct AngleConstraint{D, I}
...
+    needs_virial::Bool
+    step_n::Int
+    dt::DT
+    virial_scale::S
+    buffers::B
+    coords_buffer::CB
+    velocities_buffer::VB
+end
+
+function ConstraintApplicationContext(kind::K;
```
jgreener64: I think this can be done via `@kwdef` on the struct.

src/constraints/lincs.jl
```diff
@@ -735,13 +776,40 @@ function lincs_apply!(coords, old_coords, data::LincsData, ws::LincsWorkspace,
...
 end
 
-function lincs_vel_apply!(velocities, coords, data::LincsData, ws::LincsWorkspace, boundary)
+lincs_apply!(coords, old_coords, data::LincsData, ws::LincsWorkspace, boundary) =
+    lincs_apply!(coords, old_coords, data, ws, boundary, default_position_constraint_context())
+
+function accumulate_lincs_position_virial!(data::LincsData, ws::LincsWorkspace,
+                                           context)
+    context.needs_virial || return context
+    context.kind isa PositionConstraintApplication ||
```
jgreener64: Use `if` if multi-line.

src/interactions/ewald.jl
```diff
@@ -246,14 +246,12 @@ end
 function ==(a::Ewald, b::Ewald)
     return a.dist_cutoff == b.dist_cutoff &&
            a.error_tol   == b.error_tol   &&
-           a.excluded_pairs == b.excluded_pairs &&
```
jgreener64, line 249: My bad!

src/interactions/lennard_jones.jl
```diff
@@ -152,55 +152,139 @@ The potential energy is defined as
...
+    # Representative C6- and C12-like terms. Used to determine both the 
+    # physical units of the accumulated quantities and the factor types.
+    term_6_example  = at.ϵ * at.σ^6
+    term_12_example = at.ϵ * at.σ^12
+
+    # The correction requires summing over N(N + 1) / 2 atom pairs. For large
+    # systems this can involve millions of additions. Accumulating in Float32
+    # can cause loss of precision, so we will do the accumulation always in 
+    # Float64
+    Tacc = promote_type(
```
jgreener64: `Tacc = Float64` then?

src/interactions/lennard_jones.jl
```diff
@@ -152,55 +152,139 @@ The potential energy is defined as
...
+        Float64,
+        typeof(ustrip(term_6_example)),
+        typeof(ustrip(term_12_example)),
+        typeof(ustrip(dist_cutoff)),
+    )
+
+    # Accumulate unitless numerical values in the higher-precision type.
+    ϵσ6_sum  = zero(Tacc)
```
jgreener64: Is there a value to accumulating without units?

src/interactions/lennard_jones.jl
```diff
@@ -152,55 +152,139 @@ The potential energy is defined as
...
+    n_pairs_acc = Tacc((n_atoms * (n_atoms + 1)) ÷ 2)
+
+    # Restore the units.
+    ϵσ6_mean =
+        (ϵσ6_sum / n_pairs_acc) * unit(term_6_example)
+
+    ϵσ12_mean =
+        (ϵσ12_sum / n_pairs_acc) * unit(term_12_example)
+
+    # Evaluate the final correction factors in the high-precision accumulator
```
jgreener64: Quite verbose, reduce some of the comments.

src/interactions/lennard_jones.jl
```diff
@@ -152,55 +152,139 @@ The potential energy is defined as
...
-struct LJDispersionCorrection{F}
-    factor::F
+struct LJDispersionCorrection{F6, F12}
+    factor_6::F6
+    factor_12::F12
 end
 
-function LJDispersionCorrection(atoms, dist_cutoff, σ_mix=LorentzMixing(),
+Base.getproperty(dc::LJDispersionCorrection, name::Symbol) =
+    name === :factor ? getfield(dc, :factor_6) + getfield(dc, :factor_12) : getfield(dc, name)
```
jgreener64: I'd prefer to remove all the places we call `.factor`.

src/spatial.jl
```diff
@@ -607,16 +591,19 @@ function wrap_coords(v, boundary::TriclinicBoundary)
...
     # Bound in x-axis
-    v_wrap -= bv[1] * floor((v_wrap[1] - dx - (v_wrap[2] - dy) / boundary.tan_a_b) * rs[1])
+    v_wrap -= bv[1] * floor(
+        (v_wrap[1] - dx - (v_wrap[2] - dy) * boundary.cot_a_b) * rs[1],
+    )
     return v_wrap
 end
 
 # Unwrap coordinates so that molecules are not split over the boundary
+# Ensures that the center of geometry of bonded clusters of molecules
```
jgreener64: Bonded clusters of atoms?

test/basic.jl
```diff
@@ -1,3 +1,4 @@
+if false == true
```
jgreener64: Presumably left in from testing.

test/basic.jl
```diff
@@ -876,3 +928,331 @@ end
...
+
+        bmat_original = Molly.boxmatrix(sys.boundary)
+        bmat_new = F * bmat_original
+
+        boundary_new = TriclinicBoundary(bmat_new)
+
+        # Apply the same affine transformation to every coordinate.
+        coords_new = [F * coord for coord in sys.coords]
+
+        sys_out = typeof(sys)(
```
jgreener64: You can probably use the `System` constructor from another `System` here.

test/basic.jl
```diff
@@ -876,3 +928,331 @@ end
...
+                # Enzyme differentiates the stored tail energy (U6 + U12), while
+                # Molly's pressure virial uses the mechanical LJ tail, 2U6 + 4U12.
+                correction += (2 * U6 + 4 * U12) - (U6 + U12)
+            end
+        end
+
+        return correction * I
+    end
+
+    data_dir = joinpath(dirname(pathof(Molly)), "..", "data")
```
jgreener64: These variables are already available.

test/basic.jl
```diff
@@ -876,3 +928,331 @@ end
         end
     end
 end
+
+end
+
+
+@testset "Virial Correctness" begin
```
jgreener64: I think this should move to test/gradients.jl. It still gets run, it just means that Enzyme errors won't stop the non-Enzyme tests.

src/simulators.jl
```diff
@@ -284,38 +486,67 @@ end
         cons_vel_storage = zero(sys.velocities)
     end
     dt_div2 = sim.dt / 2
-    dt_sq_div2 = sim.dt^2 / 2
 
     progress = setup_progress(n_steps, show_progress)
     for step_n in (init_step + 1):(init_step + n_steps)
+        needs_vir_step = needs_virial_on_step(needs_vir, needs_vir_steps, step_n)
+        pressure_kin_tensor = nothing
```
jgreener64: Ideally this would be the same type throughout the step, could set before the step loop and zero if required each step.

src/simulators.jl
```diff
@@ -397,49 +634,81 @@ end
...
 
+        # Remove drift after the final velocity constraints/virial accumulation
+        # and before the kinetic tensor snapshot used for pressure coupling.
         if !iszero(sim.remove_CM_motion) && step_n % sim.remove_CM_motion == 0
             remove_CM_motion!(sys)
         end
-        recompute_forces = apply_coupling!(sys, buffers, sim.coupling, sim, neighbors, step_n;
-                                           n_threads=n_threads, rng=rng)
+
+        if using_constraints
```
jgreener64: These two conditions can be one line (and elsewhere).

src/simulators.jl
```diff
@@ -555,20 +844,36 @@ end
...
-
-Does not currently work with coupling methods that alter the velocity.
-Does not currently remove the center of mass motion.
+Position constraints are supported. Coupling methods are intentionally
+unsupported.
 
 # Arguments
 - `dt::T`: the time step of the simulation.
-- `coupling::C=nothing`: the coupling which applies during the simulation.
+- `coupling::C=nothing`: no coupling. Other values throw an `ArgumentError`
```
jgreener64: If we are going to disallow coupling for Stormer-Verlet we can just remove the coupling field from the struct.

src/energy.jl
```diff
@@ -98,17 +99,28 @@ The [virial definition from LAMMPS](https://docs.lammps.org/compute_stress_atom.
...
 [`Ewald`](@ref) and [`PME`](@ref) methods computed as indicated in
 [Essmann et al. 1995](https://doi.org/10.1063/1.470117).
-Contributions from constraints, implicit solvent methods and bias potentials are ignored.
+Contributions from implicit solvent methods and bias potentials are ignored.
+For constrained systems, constraint contributions are approximated using a
+deterministic small-step constraint preview.
 Compatible with virtual sites apart from [`OutOfPlaneSite`](@ref).
 
-To calculate the scalar virial, see [`scalar_virial`](@ref).
+Passing a simulator for constrained systems uses the same deterministic preview
```
jgreener64: I don't quite get what this means, it doesn't seem that giving sim does anything? If it's not required, remove along with the functions at the end of simulators.jl.

jgreener64 (commented):
Looks good, merge after addressing comments.

docs/src/documentation.md
allexandre97: APIs now use sys, neighbors, step_n, buffers

docs/src/documentation.md
allexandre97: Same as above, changed API to that order

src/constraints/constraints.jl
allexandre97: Removed constructor, now struct is built with keywords

src/constraints/lincs.jl
allexandre97: Changed

src/interactions/lennard_jones.jl
allexandre97: Changed

src/interactions/lennard_jones.jl
allexandre97: Updated to keep units attached

src/interactions/lennard_jones.jl
allexandre97: Made comments more concise

src/energy.jl
allexandre97: This was added in case at some point we want to support passing a specific / custom integrators to calculate the constraints contribution to the virial in this path. The first implementation I did of this required passing a simulator, in fact; now it just falls back to using a verlet step no matter what.

I removed it now, we can always implement this again if we find a good use case.

src/interactions/lennard_jones.jl
allexandre97: Removed

src/simulators.jl
allexandre97: Changed so we only allocate for the tensor once, before the loop

src/simulators.jl
allexandre97: Changed to `if using_constraints && needs_vir_step`

src/simulators.jl
allexandre97: Removed coupling from `StormerVerlet`

src/spatial.jl
allexandre97: My bad, that was a typo

test/basic.jl
allexandre97: Yeah... sorry you had to see such aberration

test/basic.jl
allexandre97: Moved to other file

test/basic.jl
allexandre97: Changed

PR #267, opened 2026-07-02
Support for Inf dielectric constant for soft core reaction field interactions

Immi000:
A while ago I [proposed support for conducting boundary conditions](https://github.com/JuliaMolSim/Molly.jl/pull/247) for the CoulombReactionField. This PR adds the same functionality to the recently added soft-core variants.


jgreener64:
Great, thanks.

PR #268, opened 2026-07-02
added CUDA kernel for random_velocities!

ywitzky:
I have added a CUDA kernel for random_velocities!()

Performance improvement has been measured for a dense LJ system with fixed average density that has been simulated with the  Langevin algorithm for varying number of particles and box sizes. Benchmarked on a NVIDIA GeForce RTX 4090 I get:

before changes: commit 0f19f896
simulation 0 took 1.884167053 for 100 atoms
simulation 2 took 1.780424146 for 400 atoms
simulation 4 took 2.333971938 for 1600 atoms
simulation 6 took 3.446783882 for 6400 atoms
simulation 8 took 8.438397874 for 25600 atoms
simulation 10 took 31.827900505 for 102400 atoms

random_velocities!: commit 92eba91d
simulation 0 took 1.235772652 for 100 atoms
simulation 2 took 1.212126355 for 400 atoms
simulation 4 took 1.446382735 for 1600 atoms
simulation 6 took 2.0191413 for 6400 atoms
simulation 8 took 3.58989918 for 25600 atoms
simulation 10 took 11.58562941 for 102400 atoms

### DIFF
--- a/ext/MollyCUDAExt.jl
+++ b/ext/MollyCUDAExt.jl
@@ -19,7 +19,8 @@ module MollyCUDAExt
 
 using Molly
 using Molly: from_device, box_sides, sorted_morton_seq!, sum_pairwise_forces_gpu,
-             sum_pairwise_forces_nonl, sum_pairwise_potentials_gpu, volume
+             sum_pairwise_forces_nonl, sum_pairwise_potentials_gpu, volume, random_velocities!, remove_CM_motion!
+
 using CUDA
 using Atomix
 using KernelAbstractions
@@ -2341,5 +2342,41 @@ function pairwise_force_kernel_nonl!(forces::AbstractArray{T}, coords_var, veloc
 
     return nothing
 end
+
+import Molly.random_velocities! 
+
+function random_velocities!(vels::CuArray, sys::Molly.AbstractSystem, temp; rng)
+    T = typeof(convert(AbstractFloat, ustrip(temp)))
+    V = unit(eltype(eltype(vels))) ### avoids SVector
+    masses=Molly.masses(sys)
+
+    M = unit(eltype(masses))
+    units =  T(ustrip(uconvert(V^2,T(sys.k) * temp / M))) ### use square units
+
+    N = length(vels)
+    raw_ptr = CuPtr{T}(Base.pointer(vels))
+
+    GC.@preserve vels masses begin
+        flat_vels_raw = Base.unsafe_wrap(CuArray, raw_ptr   , (3*N,))
+
+        CUDA.@cuda threads=256 blocks=cld(N, 256) fastmath=true random_velocities_kernel!(flat_vels_raw,ustrip.(masses), N,units)
+    end
+    
+    return vels 
+end
+
+function random_velocities_kernel!(vels,masses, N,units)
+    i = threadIdx().x + (blockIdx().x - 1) * blockDim().x
+    if i <= N
+        @inbounds scale = sqrt(units/masses[i])
+        base_idx = 3 * (i - 1)
+
+        @inbounds vels[base_idx+1] = randn() * scale
+        @inbounds vels[base_idx+2] = randn() * scale
+        @inbounds vels[base_idx+3] = randn() * scale
+    end
+    return nothing
+end
+
 
 end
--- a/ext/MollyCUDAExt.jl
+++ b/ext/MollyCUDAExt.jl
@@ -2343,40 +2343,4 @@ function pairwise_force_kernel_nonl!(forces::AbstractArray{T}, coords_var, veloc
     return nothing
 end
 
-import Molly.random_velocities! 
-
-function random_velocities!(vels::CuArray, sys::Molly.AbstractSystem, temp; rng)
-    T = typeof(convert(AbstractFloat, ustrip(temp)))
-    V = unit(eltype(eltype(vels))) ### avoids SVector
-    masses=Molly.masses(sys)
-
-    M = unit(eltype(masses))
-    units =  T(ustrip(uconvert(V^2,T(sys.k) * temp / M))) ### use square units
-
-    N = length(vels)
-    raw_ptr = CuPtr{T}(Base.pointer(vels))
-
-    GC.@preserve vels masses begin
-        flat_vels_raw = Base.unsafe_wrap(CuArray, raw_ptr   , (3*N,))
-
-        CUDA.@cuda threads=256 blocks=cld(N, 256) fastmath=true random_velocities_kernel!(flat_vels_raw,ustrip.(masses), N,units)
-    end
-    
-    return vels 
-end
-
-function random_velocities_kernel!(vels,masses, N,units)
-    i = threadIdx().x + (blockIdx().x - 1) * blockDim().x
-    if i <= N
-        @inbounds scale = sqrt(units/masses[i])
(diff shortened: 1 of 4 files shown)


nhz2:
You might find https://github.com/medyan-dev/PhiloxRNG.jl helpful for even more performance

ext/MollyCUDAExt.jl
```diff
@@ -2342,4 +2343,40 @@ function pairwise_force_kernel_nonl!(forces::AbstractArray{T}, coords_var, veloc
...
+    V = unit(eltype(eltype(vels))) ### avoids SVector
+    masses=Molly.masses(sys)
+
+    M = unit(eltype(masses))
+    units =  T(ustrip(uconvert(V^2,T(sys.k) * temp / M))) ### use square units
+
+    N = length(vels)
+    raw_ptr = CuPtr{T}(Base.pointer(vels))
+
+    GC.@preserve vels masses begin
```
jgreener64: This seems a bit magical, is this guaranteed safe? Might be better to set one `SVector` per thread directly.

ext/MollyCUDAExt.jl
```diff
@@ -2342,4 +2343,40 @@ function pairwise_force_kernel_nonl!(forces::AbstractArray{T}, coords_var, veloc
...
+        flat_vels_raw = Base.unsafe_wrap(CuArray, raw_ptr   , (3*N,))
+
+        CUDA.@cuda threads=256 blocks=cld(N, 256) fastmath=true random_velocities_kernel!(flat_vels_raw,ustrip.(masses), N,units)
+    end
+    
+    return vels 
+end
+
+function random_velocities_kernel!(vels,masses, N,units)
```
jgreener64: Could you make this a KernelAbstractions kernel and put it in kernels.jl? Then it would work on all backends.

test/gpu_consistency.jl
```diff
@@ -536,6 +536,61 @@
...
+                atoms=AT(atoms),
+                coords=AT(coords),
+                boundary=boundary,
+                velocities=AT(velocities),
+                pairwise_inters=pairwise_inters,
+                loggers=(temp=TemperatureLogger(100),),
+                #neighbor_finder=neighbor_finder,
+            )
+
+            using Random, Statistics
```
jgreener64: These should already be imported here.

jgreener64 (commented):
Thanks for this. I had put this off because I couldn't see a way to use the `rng` option consistently on GPU. I think we are going to change the default `rng` to `nothing` though, which would give no guarantees about reproducibility and can use this path, with the old path being used when `rng` is set.

ext/MollyCUDAExt.jl
ywitzky: I have done that now with similar performance.

ywitzky:
@nhz2 I couldnt get any improvements with PhilanxRNG with in the explicit CUDA version. The rewrite with KernelAbstractions.jl makes PhilanxRNG less interesting to use, as PhilanxRNG is only implemented for CUDA.

src/kernels.jl
```diff
@@ -646,3 +646,18 @@ end
         end
     end
 end
+
+@kernel function random_velocities_kernel!(vels::AbstractVector{SVector{D, C}},@Const(masses::AbstractVector{M}), units, @Const(virtual_sites))  where {D,C,M}
+    i = @index(Global, Linear)
+    if i<= length(vels)
+        @inbounds scale = C(sqrt(units/masses[i]))
```
jgreener64: `units` here can be called `kT`.

src/kernels.jl
```diff
@@ -646,3 +646,18 @@ end
         end
     end
 end
+
+@kernel function random_velocities_kernel!(vels::AbstractVector{SVector{D, C}},@Const(masses::AbstractVector{M}), units, @Const(virtual_sites))  where {D,C,M}
```
jgreener64, line 650: The `virtual_sites` are currently unused here, you should have something like `scale = ... * !virtual_site_flags[i]` which would avoid branching.

src/kernels.jl
```diff
@@ -646,3 +646,18 @@ end
...
+    i = @index(Global, Linear)
+    if i<= length(vels)
+        @inbounds scale = C(sqrt(units/masses[i]))
+        @inbounds vels[i] = SVector{D,C}(randn() * scale,randn() * scale, randn() * scale)
```
jgreener64: This works for 3D but not for 2D. Could have a function with dispatch or maybe `randn(SVector{D, C})`.

test/gpu_consistency.jl
```diff
@@ -536,6 +536,59 @@
...
+                vel=deepcopy(velocities)
+                Molly.random_velocities!(vel, sys, temp)
+                mean_cpu = mean(mean(vel))
+                std_cpu = mean( std(vel))
+
+                vel= Molly.random_velocities!(velocities_gpu, sys2, temp; rng=cuRAND.LibraryRNG())
+                mean_gpu = mean(mean(vel))
+                std_gpu = mean( std(vel))
+
+                @test abs(mean_cpu) < 0.005u"nm * ps^-1"
```
jgreener64: You could also check that `mean(norm, vel)` is in an expected range.

jgreener64 (commented):
Great, a few more comments.

src/kernels.jl
ywitzky: Done. Added a similar kernel for coupling with Andersen thermostat as well. This gives similar speed ups. The second one would be unnecessary if we use an additional allocation which I tried to avoid.

jgreener64 (commented):
It's a technical point but I think it's better to explicitly set virtual site velocities to zero rather than skipping them.

Fix that and add the mean velocity test and this is ready for merge.

nhz2:
> @nhz2 I couldnt get any improvements with PhilanxRNG with in the explicit CUDA version. The rewrite with KernelAbstractions.jl makes PhilanxRNG less interesting to use, as PhilanxRNG is only implemented for CUDA.

I can make a follow up PR for this, but the idea is that PhiloxRNG provides a pure function `randn_f32` that maps a `(ctr0::UInt64, ctr1::UInt64, key::UInt64)` to a `NTuple{4, Float32}`. This function is supported by all platforms and is [even vendored into GPUArrays.jl](https://github.com/JuliaGPU/GPUArrays.jl/blob/c575bac619095e6448325388e996ec7b17b80fc5/src/host/random.jl#L77-L78)

Using `randn_f32` you only need one go of philox and two box muller transforms per velocity 3 vector. But it is important that `randn_f32` is in the same kernel as the scaling to get good performance.

`rng` can be used to get the `key` and `ctr1` using a pair of `rand(rng, UInt64)`. Then `ctr0` can be something like `i<<16`, where i is the atom index. This leaves you 16 bits to play with if you want to use more philox samples in the future. For example maybe you want to add a Float64 version. This scheme also means that the random velocities will be almost the same on all platforms (down to floating point implementation differences).

ywitzky:
@nhz2 Thanks for the reply. I tried with the functionality of GPUArrays now but couldnt get any improvements. Feel free to do a follow up PR.

@jgreener64 Only velocities are zeroed for random_velocites as the Andersen thermostat implementation for the cpu does not zero the virtual sites.

jgreener64:
Thanks, and @nhz2 do make a follow-up PR if you get a chance.

PR #270, opened 2026-07-11
Use PhiloxRNG

nhz2:
This PR uses PhiloxRNG for generating noise for `random_velocities`, `apply_Andersen_coupling_kernel!`, and `Langevin` and `LangevinSplitting` simulation as a followup to #268 

Here is a benchmark for Langevin dynamics without any forces.

```julia
using Molly
using CUDA
using Random
using StaticArrays

function runbench(; ftype::Type{T}=Float32, atype::Type{AT}=Array, n_atoms = 100000) where {T, AT}
    atom_mass = 10.0f0u"g/mol"
    box_len = T(cbrt(n_atoms/100)*2.0)*u"nm"
    boundary = CubicBoundary(box_len)
    temp = 100.0f0u"K"
    atoms = AT([Atom(mass=atom_mass, σ=0.3f0u"nm", ϵ=0.2f0u"kJ * mol^-1") for i in 1:n_atoms])
    coords = AT([rand(SVector{3,T})*box_len for i in 1:n_atoms])
    velocities = AT([random_velocity(atom_mass, temp) for i in 1:n_atoms])
    simulator = Langevin(;
        dt=0.002f0u"ps",
        temperature=temp,
        friction=T(1.0)u"ps^-1",
    )

    sys = System(
        atoms=atoms,
        coords=coords,
        boundary=boundary,
        velocities=velocities,
        # loggers=(
        #     temp=TemperatureLogger(typeof(1.0f0u"K"), 10),
        # ),
    )

    simulate!(deepcopy(sys), simulator, 20; show_progress=false) # Compile function
    n_iters = 10_000
    (CUDA.@elapsed simulate!(sys, simulator, n_iters; show_progress=false))/n_iters
end

for atype in (Array, CuArray)
    for n_atoms in (10_000, 100_000)
        for ftype in (Float32, Float64)
            println("$n_atoms atoms $ftype $atype")
            println(round(Int, runbench(;n_atoms, ftype, atype)*1E6), " μs/it")
            println()
        end
    end
end
```

Results PR vs master:

Julia Version 1.12.6
CPU: 16 × AMD Ryzen 7 9800X3D 8-Core Processor
NVIDIA GeForce RTX 3080

10000 atoms Float32 Array
93 μs/it vs 170 μs/it

10000 atoms Float64 Array
144 μs/it vs 192 μs/it

100000 atoms Float32 Array
905 μs/it vs 1723 μs/it

100000 atoms Float64 Array
1373 μs/it vs 1915 μs/it

10000 atoms Float32 CuArray
31 μs/it vs 158 μs/it

10000 atoms Float64 CuArray
88 μs/it vs 190 μs/it

100000 atoms Float32 CuArray
42 μs/it vs 1148 μs/it

100000 atoms Float64 CuArray
174 μs/it vs 1254 μs/it


src/kernels.jl
```diff
@@ -662,29 +662,144 @@ end
...
-            @inbounds scale = C(sqrt(units/masses[i]))
-            @inbounds vels[i] = SVector{D, C}(ntuple(i -> randn()* scale, Val(D)))
+    natoms = length(vels)%UInt64
+    ctr0 = i%UInt64
+    @inbounds if i <= length(vels)
+        if !virtual_sites[i]
+            scale = C(Base.FastMath.sqrt_fast(units / masses[i]))
+            vels[i] = @inline(randn_svec(SVector{D, FT}, ctr0, ctr1, key, natoms)) * scale
         else
-            @inbounds velocities[i] = SVector{D, C}(ntuple(i -> 0.0, Val(D)))
```
nhz2: I don't think this branch is covered by the tests, because this looks like a typo

src/kernels.jl
```diff
@@ -662,29 +662,144 @@ end
...
 end
 
-@kernel function random_velocities_kernel!(vels::AbstractVector{SVector{D, C}},@Const(masses::AbstractVector{M}), units, @Const(virtual_sites))  where {D,C,M}
+# `D` standard normals as an SVector of float type `FT`, from the Philox counter-based RNG.
+# Using a generated function to unroll for arbitrary dimension D. `Float32` gets a dedicated
+# method; any other `AbstractFloat` (including `Float64`) falls back to
+# drawing in `Float64` and converting.
+# After calling this advance ctr0 by num_philox_calls*natoms
+@generated function randn_svec(::Type{SVector{D, Float32}}, ctr0::UInt64, ctr1::UInt64,
+                               key::UInt64, natoms::UInt64) where {D}
```
nhz2: Where can I add unit tests for this? This function works for any D, but I'm missing tests for the case where D > 4.

jgreener64:
Thanks, I'll take a look next week. Funnily enough I am just changing the way we deal with randomness so I may get you to rebase this on top of new changes.

In brief the plan is to have default `rng=nothing`, i.e. letting Molly generate whatever random numbers it wants, which makes fast paths that don't respect `rng` like in https://github.com/JuliaMolSim/Molly.jl/pull/268 valid. Giving something else to `rng` would go via the old path of always generating the random numbers on CPU.

Perhaps this PR allows us to keep a default `rng` though.

src/kernels.jl
jgreener64: I didn't check that PR carefully enough, but I have fixed this locally and will push fixes next week.

src/kernels.jl
jgreener64: We only support `D` of 2 or 3 in general.

jgreener64: Tests could go in test/basic.jl.

src/simulators.jl
```diff
@@ -1037,7 +1038,17 @@ end
         apply_loggers!(sys, neighbors, init_step, buffers, initial_loggers; n_threads=n_threads)
     end
     accels_t = calc_accels.(forces_t, masses(sys))
-    noise = zero(sys.velocities)
+    vel_scales = Fill(sim.vel_scale, length(masses(sys)))
```
nhz2, line 1080: I'm using a `Fill` array so that `Langevin` and `LangevinSplitting` can share the same kernel code, even though `Langevin` has a single constant `vel_scale` for all atoms and `LangevinSplitting` has different `vel_scale` per atom.

src/kernels.jl
nhz2: Okay, I've removed the generated function and just hand implemented the 2 and 3 D cases.

nhz2:
That makes sense, though I think a good default is to derive from the `Random.default_rng()`. That way a global `Random.seed!` can be used to make a script or test reproducible.

jgreener64:
I just pushed some changes so you will need to update the PR.

Since your approach uses `rng` I think you can replace both random branches with yours. Then we can keep `Random.default_rng()` as the default, and reproducibility can happen by giving `rng` or by changing the global seed.

nhz2:
Okay I think I've merged in the changes correctly

src/kernels.jl
```diff
@@ -661,3 +661,116 @@ end
...
+    SVector{2, FT}(c1, c2)
+end
+@inline function randn_svec(::Type{SVector{3, FT}}, ctr0::UInt64, ctr1::UInt64, key::UInt64, natoms::UInt64) where {FT <: AbstractFloat}
+    c1, c2 = randn_f64(ctr0, ctr1, key)
+    ctr0 += natoms
+    c3, c4 = randn_f64(ctr0, ctr1, key)
+    SVector{3, FT}(c1, c2, c3)
+end
+
+# units is kT
```
jgreener64: Change the variable to `kT`.

src/kernels.jl
```diff
@@ -661,3 +661,116 @@ end
...
+# host
+function langevin_o_step!(
+        vels::AbstractVector{SVector{D, C}},
+        vel_scales::AbstractVector,
+        noise_scales::AbstractVector,
+        philox_ctr1::UInt64,
+        philox_key::UInt64,
+        ::Type{FT},
+    ) where {D, C, FT}
+    @assert eachindex(noise_scales) == eachindex(vels)
```
jgreener64: In general I prefer to avoid `@assert`s in the code since they are turned off in some situations. These can be assumed to be true here.

test/basic.jl
```diff
@@ -50,6 +50,22 @@
     @test 0.35              < std(vels_nounits_1)  < 0.55
     @test 0.35              < std(vels_nounits_2)  < 0.55
 
+    for D in 2:3
+        for FT in (Float32, Float64)
+            local natoms = UInt64(1_000)
```
jgreener64, line 55: Don't use `local` annotations unless required.

jgreener64 (commented):
Looks good, just a few comments.

Since the new code includes part of the random number generation algorithm, it would be worth adding a few tests to check that a `random_velocities` call looks sensible on CPU and GPU; i.e. it gives a reasonable spread of values without repetition and with an expected norm distribution.

You may need to update https://juliamolsim.github.io/Molly.jl/dev/documentation/#Randomness to reflect the new behaviour.

test/basic.jl
nhz2: The local is because these variables are local to only the inner most for loop not the whole test set. I don't want some potential future test in this testset that happens to be using these variable names to become connected to these variables.

nhz2:
I've added some more tests, and found a bug, I was using `Base.FastMath.sqrt_fast` on the GPU, but it turns out on CUDA this is not precise enough for Float64. It looks like it does the sqrt in Float32 and converts to Float64 at the end. I've changed back to regular sqrt.

jgreener64:
Looks good. The CI is failing (let me know if you can't see the logs), possibly due to a different seed in the barostat tests.

test/coupling.jl
```diff
@@ -465,6 +458,17 @@ end
             volume=VolumeLogger(n_log_steps),
         ),
     )
+    (sys, rng)
+end
+
+@testset "25 Argon atoms NVT baseline" begin
+    sys, rng = get_argon_sys(Array)
+    temp = 288.15u"K"
+    random_velocities!(sys, temp; rng=rng)
```
nhz2, line 467: I've added a call to `random_velocities!` so the simulation starts near the correct temp instead of needing to warm up.

This NVT test was also repeated at the start of each of the Monte Carlo barostat tests, so I factored it out into its own testset.

nhz2:
NotGradients tests are passing now.

test/simulation.jl
```diff
@@ -977,7 +979,7 @@ end
     @time simulate!(repsys, simulator, n_steps; assign_velocities=false, n_threads=1)
 
     efficiency = repsys.exchange_logger.n_exchanges / repsys.exchange_logger.n_attempts
-    @test efficiency > 0.1 # This is a fairly arbitrary threshold, but it's a good test for very bad cases
+    @test efficiency > 0.07 # This is a fairly arbitrary threshold, but it's a good test for very bad cases
```
jgreener64: Why is the drop required here, if you are setting the seed? Just checking this is not hiding a change in behaviour.

src/spatial.jl
```diff
@@ -862,39 +793,39 @@ Virtual sites are given a velocity of zero.
...
+function random_velocities!(vels::AbstractVector{SVector{D, C}}, sys::AbstractSystem, temp; rng=Random.default_rng()) where {D, C}
+    FT = float_type(sys)
+    ms = from_device(masses(sys))
+    vsf = from_device(sys.virtual_site_flags)
+    kT = sys.k * temp
+    ctr1 = rand(rng, UInt64)
+    key = rand(rng, UInt64)
+    natoms = UInt64(length(ms))
+    @inbounds @simd ivdep for i in eachindex(vels)
+        scale = ifelse(vsf[i], zero(C), C(Base.FastMath.sqrt_fast(kT/ms[i])))
```
jgreener64, line 809: Might be safer to just use `sqrt` here too? Unless you are sure this is safe, and it gives a speedup.

jgreener64 (commented):
Looks good, a couple more things.

Could you remove the unused functions, I think `andersen_coupling_kernel!` and `andersen_atoms_bump`.

Claude suggested the below, up to you if this requires changes.

> GPU consistency test tolerance is too tight (test/gpu_consistency.jl, "Langevin random numbers") — maximum(norm, vels_gpu_cpu .- vels_cpu) < eps(one(FT)) * vel_scale doesn't account for draws several σ in magnitude carrying 1-ulp differences. On this machine the CPU @simd ivdep loop in random_velocities! (src/spatial.jl:800-812) compiles the muladds in randn_svec differently from a scalar/kernel evaluation, producing diffs of 2.2e-16 nm/ps (F64, tolerance 1.1e-16) and 1.7e-7 (F32, tolerance 5.9e-8) — i.e. the test would fail here, and whether it passes is CPU/vectorization-dependent. Recommend scaling the bound by the max draw magnitude (up to 9.5σ for F64, 6.8σ for F32), e.g. 16*eps(one(FT))*vel_scale, or a per-element relative bound. The langevin_o_step! test's 4*eps(one(FT)) is marginal for the same reason (I measured 2 eps; values reach ~4–5, so a single ulp is 4–5 eps).

src/spatial.jl
nhz2: This is safe since kT and mass are positive and finite. The issue with regular sqrt is that it needs to throw an error if any mass is negative, which the compiler has a hard time dealing with. `Base.sqrt_llvm` would be better but it is an intrinsic so it doesn't work with units or other custom number types. Ideally Base would have something like https://github.com/JuliaMath/NaNMath.jl so we can get compiler friendly math that other number types can specialize.

nhz2:
I've removed the dead code, and adjusted the tolerances in the tests. The benchmark results are also updated in the initial post.

jgreener64:
Thanks for this!

Issue #269, opened 2026-07-11
Race condition on shared `rng` in `simulate_remd!`


nhz2:
IIUC any non default `rng` passed to `simulate_remd!` is getting shared by all the threads in:

https://github.com/JuliaMolSim/Molly.jl/blob/76d910e91b97bcda60618f40fe1a9ac914a756f9/src/simulators.jl#L1661-L1665

I think this is the cause of the test failure I am seeing when I run the Molly.jl tests:

```julia
  Temperature REMD: Error During Test at ~/github/Molly.jl/test/simulation.jl:833
    Got exception outside of a @test
    TaskFailedException
        nested task error: AssertionError: length(ints) == 501
        Stacktrace:
         [1] mt_setfull!(r::MersenneTwister, ::Type{UInt64})
           @ Random ~/packages/julias/julia-1.12.6/share/julia/stdlib/v1.12/Random/src/RNGs.jl:227
```

jgreener64:
You are right, and this will be fixed by the changes mentioned in https://github.com/JuliaMolSim/Molly.jl/pull/270. The plan is to not allow a specified `rng` for this path.

jgreener64:
If `rng=Random.default_rng()`, is this thread safe? In that case we may just throw an error on this path for other `rng`.

nhz2:
Yes `rng=Random.default_rng()` is threadsafe because it isn't an actually mutable rng object, just a token to tell rand functions to use the special task local rng system described in https://www.youtube.com/watch?v=6xpo07VuyBM

```julia-repl
julia> rng = Random.default_rng()
TaskLocalRNG()

julia> ismutable(rng)
false
```

jgreener64:
Thanks, fixed in https://github.com/JuliaMolSim/Molly.jl/commit/86bbcdacd85f3204e3404ce21bb262bb7b5a00d0.

PR #271, opened 2026-07-12
Avoid unnecessary `CuArray` calls in `to_device`

nhz2:
Calling `CuArray` on an existing `CuArray` makes a copy. This PR changes the `to_device` function to avoid this copy if the input is already the correct array type. This should match the behavior of `to_device` with `Array`.

I'm not familiar with Enzyme but with some help from codex GPT the Enzyme tests are passing.
There was a type instability and an issue with spawn https://github.com/EnzymeAD/Enzyme.jl/issues/3226


src/types.jl
```diff
@@ -1425,6 +1425,7 @@ from_device(x) = Array(x)
 from_device(x::StructArray) = replace_storage(Array, x)
 
 to_device(x::Array, ::Type{<:Array}) = x
+to_device(x::AT, ::Type{AT}) where {AT <: AbstractArray} = x
```
jgreener64, line 1449: I think the method above for `Array` can now be deleted?

test/simulation.jl
```diff
@@ -831,13 +831,13 @@ end
 end
 
 @testset "Temperature REMD" begin
-    rng = MersenneTwister(1)
```
jgreener64, line 834: Don't change this for now, will fix based on #269.

src/force.jl
```diff
@@ -920,34 +922,32 @@ function pairwise_forces_loop!(fs_nounits, fs_chunks, vir_nounits, vir_chunks, a
...
-                                                step_n)
-                        f_ustrip = checked_ustrip(f, force_units)
-                        fs_chunk[i] -= f_ustrip
-                        fs_chunk[j] += f_ustrip
-
-                        if needs_vir
-                            v = dr * transpose(f)
-                            vir_chunk .+= ustrip.(v)
-                        end
+        Threads.@threads for chunk_i in 1:n_threads
```
jgreener64: As I recall my benchmarking showed a slight improvement in speed using `@spawn` so I would prefer to wait for a fix to https://github.com/EnzymeAD/Enzyme.jl/issues/3226. We will make sure tests pass as we get closer to a release.

jgreener64 (commented):
Looks good, just a few comments.

jgreener64:
Great, thanks.

PR #272, opened 2026-07-17
Add support for Double Exponential Potential

allexandre97:
Extended force field parsing logic to read and load double exponential potentials. This is done by literal string matching, the string must be strictly compliant with a valid OpneMM string for the same functional form.

On top of the normal double exponential, I have also added the soft-core variant, following the `alpha` and `beta` mixing logic found in https://doi.org/10.1039/d3dd00070b


src/interactions/double_exponential.jl
```diff
@@ -0,0 +1,348 @@
...
+    \frac{\alpha\beta e^\alpha}{\alpha - \beta} e^{-\alpha r_{ij} / r_m}
+    - \frac{\alpha\beta e^\beta}{\alpha - \beta} e^{-\beta r_{ij} / r_m}
+\right] \frac{\vec{r}_{ij}}{r_{ij}}
+```
+where
+```math
+r_m = 2^{1/6}\sigma_{ij}
+```
+
+The default ``\alpha`` and ``\beta`` values are those predicted by Garnet.
```
jgreener64: I think we should not provide defaults here.

src/interactions/double_exponential.jl
```diff
@@ -0,0 +1,348 @@
...
+    dr, 
+    atom_i,
+    atom_j,
+    energy_units=u"kJ * mol^-1", 
+    special=false,
+    args...
+) where {C, T}
+
+    if shortcut_pair(inter.shortcut, atom_i, atom_j, special)
+        return ustrip(zero(dr[1])) * energy_units
```
jgreener64: `zero_pairwise_energy` now.

src/force_field.jl
```diff
@@ -730,6 +730,52 @@ function read_ff_xml!(ff_file, ff_param_array, atom_types, atom_type_order, attr
...
+                                    at.type, at.class, at.element, at.charge, at.mass, σ, ϵ, missing, missing)
+                            end
+                        end
+                        
+                    end
+
+                end
+            else
+                report_issue("CustomNonbondedForce without "*
+                             "energy=sqrt(epsilon1*epsilon2)*(((beta*exp(alpha))/(alpha-beta))*exp(-alpha*(r/((2^(1/6))*((sigma1+sigma2)/2))))-((alpha*exp(beta))/(alpha-beta))*exp(-beta*(r/((2^(1/6))*((sigma1+sigma2)/2))))) "*
```
jgreener64: Move the string out as a variable, also this misses quotes after `=`.

src/force_field.jl
```diff
@@ -730,6 +730,52 @@ function read_ff_xml!(ff_file, ff_param_array, atom_types, atom_type_order, attr
                     push!(nbfix_pairs, NBFixPair(type1, type2, class1, class2, σ, ϵ))
                 end
             end
+        
+        elseif entry_name == "CustomNonbondedForce"
+
+            if entry["energy"] == "sqrt(epsilon1*epsilon2)*(((beta*exp(alpha))/(alpha-beta))*exp(-alpha*(r/((2^(1/6))*((sigma1+sigma2)/2))))-((alpha*exp(beta))/(alpha-beta))*exp(-beta*(r/((2^(1/6))*((sigma1+sigma2)/2)))))"
```
jgreener64: Should also check that `bondCutoff="3"`.

src/setup.jl
```diff
@@ -1819,14 +1819,34 @@ function System(T, AT, atoms, coords, boundary_used, velocities, atoms_data, vir
...
+        )
+    elseif nonzero_epsilon_count != 0
+        DoubleExponential(
+            cutoff = DistanceCutoff(T(dist_cutoff)),
+            use_neighbors = using_neighbors,
+            α = T(global_params[1]),
+            β = T(global_params[2]),
+            σ_mixing = σ_mix,
+            ϵ_mixing = ϵ_mix,
+            weight_special = pi_weight_14_lj
```
jgreener64, line 1843: This will be zero since `bondCutoff="3"`. `CustomNonbondedForce` does not support 1-4 weighting.

src/force_field.jl
```diff
@@ -730,6 +730,52 @@ function read_ff_xml!(ff_file, ff_param_array, atom_types, atom_type_order, attr
...
+                                "CustomNonbondedForce with global parameters other than \"alpha\" or \"beta\" not supported",
+                                strictness
+                            )
+                        end
+                    elseif element.name == "Atom"
+
+                        σ = add_units(parse(T, element["sigma"]), u"nm", units)
+                        ϵ = add_units(parse(T, element["epsilon"]), u"kJ * mol^-1", units)
+                        if haskey(element, "class")
+                            push!(ljforce_atom_classes, AtomType{T, T, typeof(σ), typeof(ϵ)}(
```
jgreener64, line 765: This uses the same array as the `LennardJonesForce`, either use a different one or explicitly throw an error if `LennardJonesForce` and `CustomNonbondedForce` are found in the same file.

src/force_field.jl
```diff
@@ -850,7 +904,13 @@ function MolecularForceField(T::Type, ff_files::AbstractString...; units::Bool=t
     weight_14_coulomb = ff_param_array[2]
     # Use LennardJonesForce 1-4 weighting if present
     weight_14_lj = (ff_param_array[7] ? ff_param_array[6] : ff_param_array[4])
-    dispersion_correction = (isnothing(ff_param_array[8]) ? true : ff_param_array[8])
+    dispersion_correction = (isnothing(ff_param_array[12]) ? true : ff_param_array[12])
+
+    double_exp_alpha = (ff_param_array[9] ? ff_param_array[8] : zero(T))
```
jgreener64: Both of these need to be set in the XML file, should zero them both if either is unset.

jgreener64 (commented):
Looks good. Needs a test for a system against the OpenMM energy value (which currently will fail due to the 1-4 weighting). That will also test this setup path. You can add the OpenMM energy directly into the tests.

You can change the supported XML tags section of the docs. Also, CI currently fails.

jgreener64:
Great, merge when ready.

PR #273, opened 2026-07-17
Timing Reports

ejmeitz:
This PR adds a simple timing report when `timing=true` is passed to `simulate!`. This should have a negligible effect on runtime and make benchmarking and optimizing things easier. This is no replacement for microbenchmarks but is still valuable information to have.

I ran one of the constrained protein simulations from the tests to show the result for a simulation that used most of the features Molly has. The result looks like this:

image

Potential Issues/Improvements:
- Not sure how to handle REMD as there are multiple simulations. Right now its all 0 in the report except for the `Other` row.
- Compile time is still included and not explicitly broken out. Does not affect correctness just interpretability. 
- Break out long-range timing as that is often important

Also this was made with the help of Cursor, but I did look at all the code myself and made some manual tweaks. Things look good to me. Not sure what the state of Molly test suite is though. Locally I get an error related to a tmp file being deleted when its still in use. Pretty sure thats unrelated to what I changed.


jgreener64:
Whilst this could be useful for development, I'm not sure it's worth having in the code. It also seems like the AI has recoded a version of https://github.com/KristofferC/TimerOutputs.jl.

> Locally I get an error related to a tmp file being deleted when its still in use.

Do open an issue if this persists.

ejmeitz:
Yea its similar to that library, but I didn't want to add dependencies.

My main goal with this is to enable regression testing. Once 1.0 ships something like this will  make that easier.

I agree it would be nice if it could live in its own branch, but personally I like seeing this output from LAMMPS. To be fair the main reason its useful there is for tuning the number of domains in the decomposition to minimize communication time.

jgreener64:
There are some benchmarks for regression testing in the `benchmark` directory.

PR #274, opened 2026-07-17
Minor PME CPU Improvements

ejmeitz:
Looked for some quick changes to speed up PME on CPU. The `ustrip` optimization probably could be used for the GPU path as well. I just don't have any easy way to test that right now. I do not really see any way to improve the k-space as that just calls FFTW. On the real-space side I made the following changes:

1. `ustrip` `α`, volume, `recip_box`, and `f/ϵr` once outside the k-space loop, and run convolution without units. Apply units at the end.
2. Hoisted some calculations in charge spreading and force interpolation. This probably has modest improvements, but it was definitely wasted FLOPs.
3. Pass `num_threads` to fft plans (EDIT)

### Result (6mrr, ~16k atoms, 4 Julia threads)
Timing `Molly.ewald_pe_forces!` only, 1000 iterations after warmup:

≈ **16% faster** on the PME energy+forces path single thread

### Benchmark Code
```julia
using Molly

n_threads = 1
ff = MolecularForceField(
    Float64,
    joinpath(pkgdir(Molly), "data", "force_fields", "ff99SBildn.xml"),
    joinpath(pkgdir(Molly), "data", "force_fields", "tip3p_standard.xml"),
)
sys = System(
    joinpath(pkgdir(Molly), "data", "6mrr_equil.pdb"),
    ff;
    array_type=Array,
    nonbonded_method=:pme,
    center_coords=false,
)
pme = only(filter(i -> i isa PME, sys.general_inters))
Fs = Molly.zero_forces(sys)

# warmup
for _ in 1:3
    fill!(Fs, zero(eltype(Fs)))
    Molly.ewald_pe_forces!(Fs, nothing, sys, pme, Val(false); n_threads=n_threads)
end

n = 1000
samples_ms = Vector{Float64}(undef, n)
for i in 1:n
    fill!(Fs, zero(eltype(Fs)))
    t0 = time_ns()
    Molly.ewald_pe_forces!(Fs, nothing, sys, pme, Val(false); n_threads=n_threads)
    samples_ms[i] = (time_ns() - t0) / 1e6
end
μ = sum(samples_ms) / n
se = sqrt(sum((x - μ)^2 for x in samples_ms) / (n - 1)) / sqrt(n)
println("ms / ewald_pe_forces! = $(round(μ; digits=3)) ± $(round(se; digits=3))")
```

### DIFF
--- a/src/interactions/ewald.jl
+++ b/src/interactions/ewald.jl
@@ -402,8 +402,8 @@ function PME(dist_cutoff, atoms, boundary; error_tol=0.0005, order=5,
         pc_sum, pc_abs2_sum = nothing, nothing
     end
 
-    fft_plan  = plan_fft!(charge_grid)
-    bfft_plan = plan_bfft!(charge_grid)
+    fft_plan  = plan_fft!(charge_grid; num_threads = n_threads)
+    bfft_plan = plan_bfft!(charge_grid; num_threads = n_threads)
     bsm_x = to_device(bsplines_moduli[1], AT)
     bsm_y = to_device(bsplines_moduli[2], AT)
     bsm_z = to_device(bsplines_moduli[3], AT)
--- a/src/interactions/ewald.jl
+++ b/src/interactions/ewald.jl
@@ -668,14 +668,6 @@ end
     end
 end
 
-@inline function ustrip_recip_box(recip_box)
-    return SVector(
-        ustrip.(recip_box[1]),
-        ustrip.(recip_box[2]),
-        ustrip.(recip_box[3]),
-    )
-end
-
 # Unit-stripped CPU kernel: avoids Unitful overhead on every grid point.
 @inline function recip_conv_inner_nou!(vir_nou, charge_grid::Array{Complex{T}, 3}, bsm_x, bsm_y, bsm_z,
                                        recip_box, mesh_dims, f_div_ϵr, factor, boxfactor,
@@ -788,7 +780,7 @@ function recip_conv!(vir, buffer_virial, charge_grid::Array{Complex{T}, 3}, buff
     α_u = ustrip(α)
     factor = T(π)^2 / α_u^2
     boxfactor = T(π) * ustrip(volume(boundary))
-    recip_box_u = ustrip_recip_box(recip_box)
+    recip_box_u = ustrip_vec.(recip_box)
     length_unit = unit(box_sides(boundary, 1))
     f_u = ustrip(energy_units * length_unit, f_div_ϵr)
     esum = zero(T)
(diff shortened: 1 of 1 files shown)


ejmeitz:
Ran on a second computer. The effect is real just small and even smaller when you go to multiple threads. The parallel scaling is probably something to look at. PME should scale better than this.

master: 
- 40 cores: `ms / ewald_pe_forces! = 23.859 ± 0.113`
- 1 core: `ms / ewald_pe_forces! = 30.417 ± 0.004`

this branch:
- 40 cores: `ms / ewald_pe_forces! = 23.479 ± 0.095`
- 1 core: `ms / ewald_pe_forces! = 28.393 ± 0.003`

ejmeitz:
ok adding `num_threads` kwarg to fft plans helps scaling a lot:

```julia
fft_plan  = plan_fft!(charge_grid; num_threads = n_threads)
bfft_plan = plan_bfft!(charge_grid; num_threads = n_threads)
```

```
40 cores (before): ms / ewald_pe_forces! = 23.479 ± 0.095
40 cores (now): ms / ewald_pe_forces! = 10.714 ± 0.072
1 core: ms / ewald_pe_forces! = 28.393 ± 0.003
```

src/interactions/ewald.jl
```diff
@@ -402,8 +402,14 @@ function PME(dist_cutoff, atoms, boundary; error_tol=0.0005, order=5,
         pc_sum, pc_abs2_sum = nothing, nothing
     end
 
-    fft_plan  = plan_fft!(charge_grid)
-    bfft_plan = plan_bfft!(charge_grid)
+    if AT <: AbstractGPUArray
+        fft_plan  = plan_fft!(charge_grid)
+        bfft_plan = plan_bfft!(charge_grid)
+    else
+        fft_plan  = plan_fft!(charge_grid; num_threads=n_threads)
```
jgreener64, line 409: May also be worth trying the `flags` option at [plan_fft](https://juliamath.github.io/AbstractFFTs.jl/stable/api/#AbstractFFTs.plan_fft).

Out of interest, where on GitHub is the function that takes `num_threads`? I can't seem to find it.

jgreener64 (commented):
Thanks for looking at this. The FFT change and loop hoisting look good, I am less comfortable with the `ustrip` changes as they can hide errors. Is that related to the CI failures?

src/interactions/ewald.jl
ejmeitz: I saw it on [this ](https://juliamath.github.io/FFTW.jl/stable/fft/#FFTW.plan_r2r) function and just figured it might be on the other plans as well. It turned out to be there. The FFTW/AbstractFFT docs seem to be missing somethings.

There is also `FFTW.set_num_threads` which is not documented and does a similar thing I think.

ejmeitz: The default flag here is the same as what LAMMPS does so I suspect its optimal.

The only other flag we could try is FFTW.MEASURE which has bigger overhead but actually runs an FFTW whereas FFTW.ESTIMATE (the default) uses heuristics to pick the algorithm.

ejmeitz:
Removed the `ustrip` optimization

jgreener64:
Great, thanks.

PR #276, opened 2026-07-22
Use philox hash for DPD random force

### LINKED ISSUE #275: Statistical issues with the use of `Base.hash` in `dpd_gaussian`
The `Base.hash` function is good enough for use in `Dict`, but isn't statistically good for generating random numbers.

Here is an example of an unexpected correlation:

https://github.com/JuliaMolSim/Molly.jl/blob/ac668a98c73eb2eb137d5334c7138e47428f86fb/src/interactions/dpd.jl#L77-L86

```julia
using Statistics
using Random
i = 1
N = 10_000_000
base_noise_1 = randn(N)
base_noise_2 = randn(N)
noise_12 = Float64[dpd_gaussian(1, 2, n) for n in 1:N]
noise_13 = Float64[dpd_gaussian(1, 3, n) for n in 1:N]
```
```julia-repl
julia> cov(noise_13, noise_12)
0.6463430763848428

julia> cov(base_noise_1, base_noise_2)
-0.0003334227474072866
```

We can fix this by using `PhiloxRNG.randn_f32` here (for this use case I think `Float32` is good enough).

nhz2:
This PR fixes #275 by using the `randn_f32` from PhiloxRNG.jl instead of `Base.hash` which was not designed to have good statistical properties.

This PR also adds a `key` field to the `DPDInteraction` struct to change the random forces.

Using philox4x32_10 here is not ideal because it generates four random values when only one is needed. It is still reasonably fast, but there is probably room for improvement.

CC @sivasathyaseeelan

### DIFF
--- a/src/interactions/dpd.jl
+++ b/src/interactions/dpd.jl
@@ -1,7 +1,7 @@
 export DPDInteraction
 
 @doc raw"""
-    DPDInteraction(; a, γ, σ, r_c, dt, use_neighbors)
+    DPDInteraction(; a, γ, σ, r_c, dt, use_neighbors, key)
 
 The dissipative particle dynamics (DPD) interaction between two particles.
 
@@ -39,8 +39,8 @@ The conservative potential energy is
 V(r_{ij}) = \frac{a}{2} r_c \left(1 - \frac{r_{ij}}{r_c}\right)^2
 ```
 
-Deterministic pairwise random numbers are generated from a hash of the key argument,
-the particle indices, and the step number, ensuring momentum conservation and reproducibility.
+Pairwise random numbers are generated from a hash of the `key`,
+the particle indices, and the step number, ensuring momentum conservation.
 
 When using a neighbor list, set `dist_cutoff` to at least `1.5 * r_c` to provide a
 skin distance that accounts for particle movement between neighbor list rebuilds.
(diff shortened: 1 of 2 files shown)


sivasathyaseeelan (approved):
Thanks for catching this and fixing it @nhz2. Good spot on the `Base.hash` correlation, and swapping in `PhiloxRNG.randn_f32` is exactly the right call. I like that the new covariance test locks in the fix. Symmetry and momentum conservation are preserved, and since `PhiloxRNG` already runs in the Langevin kernels there is no GPU concern. LGTM.

src/interactions/dpd.jl
```diff
@@ -60,6 +61,7 @@ skin distance that accounts for particle movement between neighbor list rebuilds
     r_c::R = 1.0
     dt::D = 0.01
     use_neighbors::Bool = false
+    key::UInt64 = 0x6c6a2e796c6c6f4d # reinterpret(UInt64, b"Molly.jl")[1]
```
jgreener64: In general we want different random numbers each time, so this could be `rand(UInt64)`.

src/interactions/dpd.jl
```diff
@@ -117,7 +117,7 @@ end
     f_D = inter.γ * w_D * rdotv
 
     # Random: pairwise-correlated stochastic force
-    ξ = dpd_gaussian(atom_i.index, atom_j.index, step_n, Val(A))
+    ξ = A(dpd_gaussian(atom_i.index%UInt32, atom_j.index%UInt32, step_n%UInt64, inter.key))
```
jgreener64: Maybe have an inner function that does the `%` operations, that way testing the `dpd_gaussian` function is cleaner.

jgreener64 (commented):
Looks good, just a couple of comments.

nhz2:
Thanks for the suggestion. I added in the random seeding and fixed some of the tests that were flaky.

jgreener64:
Great, thanks.

PR #277, opened 2026-07-22
Replace copies with swap in `simulate!`

ejmeitz:
In a couple places we would do

```
a .= b
b .= a
```

instead of `a,b = b,a`. This PR fixes all those places and removes a copy from every single iteration of VV. Quick benchmark showed like 5% performance improvement which should apply to all simulations that use VV.


jgreener64:
Great, thanks.

PR #279, opened 2026-07-28
Avoids using Dispersion Correction when Custom Nonbonded forces are used

allexandre97:
Fixes a bug where dispersion correction was still being loaded despite not being defined in XML files using `CustomNonbondedForce`. The bug happened because a mistaken indexing in the `ff_param_array` when the XML was being read and loaded.


jgreener64 (commented):
Looks good, merge when ready.

PR #281, opened 2026-08-03
Add native ANI-2x neural-network potential (energy + analytic forces, CPU/Metal/CUDA)

sivasathyaseeelan:
Adds the **ANI-2x** neural-network interatomic potential to Molly as a native, `AtomsCalculators`-compatible interaction, with **analytic energies and forces on CPU and GPU** (Apple Metal + CUDA via KernelAbstractions), validated against TorchANI.

This is the **trimmed, review-focused split of #260** (per Dr. Joe's suggestion): it contains **only the implementation, tests, and docs**. The benchmark scripts, figures, and writeup stay on the `feature/mlip-ani-infrastructure` branch (#260) so this PR is easy to review and merge.

## How it works (ML flow)

ANI-2x writes the total energy as a sum of per-atom contributions, `E = Σ_i E_i`. Each atom is described by an **Atomic Environment Vector (AEV)**, a fixed-length fingerprint of its local chemistry built from Behler–Parrinello symmetry functions, which is fed to a small **element-specific neural network** that outputs that atom's energy. ANI-2x averages an **ensemble of 8** such networks. Forces are the analytic gradient `F = -∂E/∂r`, computed by backpropagating through the networks and the AEV (no autodiff on the GPU kernels).

```mermaid
flowchart TD
    IN["Molly System<br/>coordinates + atomic numbers"] --> SP["Species indices<br/>H C N O S F Cl"]
    IN --> NL["Neighbour list<br/>(GPU: CSR built on-device)"]

    subgraph FWD["Forward pass: energy"]
      direction TB
      AEV["AEV per atom, length 1008<br/>radial G_R (112 = 7 species × 16 shifts)<br/>+ angular G_A (896 = 28 pairs × 32)<br/>symmetry functions with cosine cutoff f_C"]
      AEV --> NN["Per-element Lux MLP<br/>1008 → E_i, celu01 activations"]
      NN --> ENS["Ensemble of 8 networks<br/>(averaged)"]
      ENS --> SUM["E = Σ_i E_i + self-energy shift<br/>Hartree → eV"]
    end

    subgraph BWD["Backward pass: analytic forces"]
      direction TB
      VJP["NN VJP: ∂E/∂G_i<br/>manual backward through Dense + celu01"]
      VJP --> BK["Backward AEV kernels<br/>radial + angular → ∂E/∂r"]
      BK --> FRC["F = -∂E/∂r · (Hartree→eV)<br/>ensemble-averaged"]
    end

    SP --> AEV
    NL --> AEV
    SUM --> EOUT(["AtomsCalculators.potential_energy → eV"])
    AEV -. reuse forward AEV .-> VJP
    FRC --> FOUT(["AtomsCalculators.forces! → eV/Å"])
```

All of the numerically heavy steps (AEV forward, AEV backward, the neighbour-list CSR build) are **KernelAbstractions** kernels, so the exact same code runs on **CPU, Apple Metal, and CUDA**. On GPU the AEV forward/backward use a write-reduced (one-workgroup-per-atom) scheme and the neighbour-list CSR is built on-device, so there is no host round-trip.

## What's here

- **Implementation**
  - `src/interactions/ml_potentials.jl`: the `ANIPotential` type + `AtomsCalculators` energy/forces interface, plus the scalar AEV helpers; `src/types.jl` / `src/Molly.jl` wiring.
  - `ext/MollyLuxExt.jl`: AEV radial/angular symmetry functions, per-element Lux MLPs, the staged energy path, and the analytic backward pass for forces (NN VJP → backward AEV kernels). Loaded via the `MollyLuxExt` extension (weakdeps `Lux` + `HDF5`).
- **Tests** (`test/ml_potentials.jl`): energy/AEV correctness vs TorchANI on N₂, H₂O, and a 6mrr protein slice; analytic forces vs finite-difference and TorchANI; Newton's-third-law momentum conservation (ΣF ≈ 0); NVE trajectory energy conservation. Gated on `data/ani_reference/ani2x.h5`; `test/torchani_reference.py` regenerates the reference fixtures.
- **Docs**: a worked ANI-2x example in `docs/src/examples.md` + an `docs/src/api.md` entry.

## What's excluded (on #260's branch, not here)

All `benchmark/ani*.jl` scripts, `benchmark/images/*.png`, `benchmark/ani_benchmarks.md`, and the ANI additions to `benchmark/Project.toml`.

## One thing to decide

`data/ani_reference/ani2x.h5` is the ANI-2x model weights (~55 MB) and is required for the potential to run; `data/ani_reference/6mrr_ani2x.json` is the full-protein reference (large). I've committed them so the tests run as-is, but if you'd prefer these fetched as a lazy `Artifact`/download rather than committed to the repo, I'm happy to switch to that. Just say the word.

Supersedes the reviewable core of #260.


docs/src/api.md
```diff
@@ -8,6 +8,7 @@ Package extensions are used in order to reduce the number of dependencies:
 - To use [`visualize`](@ref), call `using GLMakie`.
 - To use [`ASECalculator`](@ref), call `using PythonCall`.
 - To use [`rdf`](@ref), call `using KernelDensity`.
+- To use [`ANIPotential`](@ref), call `using Lux, HDF5`. Energy and forces both work with just those two packages: forces use an analytic path that runs on CPU or GPU, built on KernelAbstractions (a core Molly dependency, so no extra `using` is needed).
```
jgreener64: Second sentence not needed.

src/types.jl
```diff
@@ -23,7 +23,9 @@ export
     masses,
     charges,
     MollyCalculator,
-    ASECalculator
+    ASECalculator,
+    AbstractMLPotential,
```
jgreener64: These exports can go in ml_potentials.jl, but don't export `AbstractMLPotential`.

test/runtests.jl
```diff
@@ -143,3 +143,12 @@ end
 if GROUP in ("All", "Gradients")
     include("gradients.jl")
 end
+
+const ani_h5 = joinpath(@__DIR__, "..", "data", "ani_reference", "ani2x.h5")
+if GROUP in ("All", "NotGradients") && isfile(ani_h5)
+    using Lux, HDF5, KernelAbstractions
+    include("ml_potentials.jl")
+elseif GROUP in ("All", "NotGradients")
+    @warn "Skipping ML potential tests — ani2x.h5 not found. " *
```
jgreener64: Error if this file is not available.

test/ml_potentials.jl
```diff
@@ -0,0 +1,840 @@
...
+# Run after generating reference data:
+#   python test/torchani_reference.py --skip-protein
+# Then:
+#   julia --project -e 'using Molly, Lux, HDF5, KernelAbstractions; include("test/ml_potentials.jl")'
+# KernelAbstractions is required: the ANI forces path (AtomsCalculators.forces! →
+# compute_ani_forces_ka) is the KA analytic backward pass (no Enzyme/finite-difference paths).
+
+using Molly
+using Lux, HDF5, KernelAbstractions
+using StaticArrays, Unitful, LinearAlgebra
```
jgreener64: Move these imports to runtests.jl. Also, StaticArrays and Unitful are not required as they are re-exported from Molly.

test/ml_potentials.jl
```diff
@@ -0,0 +1,840 @@
...
+            x     = randn(Float32, aev_len, 1)   # (aev_length, batch=1)
+            y, _  = Lux.apply(model, x, ps, st)
+            @test size(y) == (1, 1)
+        end
+        println("All element sub-networks produce shape (1,1) output ✓")
+    end
+end
+
+# ============================================================================
+# Test 4: Energy sign-check for N₂ dimer (quick sanity, no reference needed)
```
jgreener64: This test doesn't seem to add much.

test/ml_potentials.jl
```diff
@@ -0,0 +1,840 @@
...
+        drift = abs(ustrip(u"eV", E_final - E0)) / abs(ustrip(u"eV", E0))
+        @test drift < 0.05   # < 5% relative drift over 50 steps at 0.1 fs
+        @test norm(sys.coords[1] - sys.coords[2]) > 0.5u"Å"  # no explosion
+        println("Initial total energy: ", round(typeof(E0),      E0,      sigdigits=8))
+        println("Final   total energy: ", round(typeof(E_final), E_final, sigdigits=8))
+        println("Relative energy drift: ", round(drift * 100, sigdigits=3), " %")
+    end
+end
+
+# (Former Test 9 "Enzyme AD forces vs TorchANI" removed: forces now come from the single
```
jgreener64: No reference to previous designs in the code.

test/ml_potentials.jl
```diff
@@ -0,0 +1,840 @@
...
+
+        println("H₂O force on O:  ", round.(fs_val[1], sigdigits=6), " eV/Å")
+        println("H₂O force on H1: ", round.(fs_val[2], sigdigits=6), " eV/Å")
+        println("H₂O force on H2: ", round.(fs_val[3], sigdigits=6), " eV/Å")
+        dev = maximum(maximum(abs.(fs_val[i] .- f_ref[i])) for i in 1:3)
+        println("Max force deviation: ", round(dev, sigdigits=4), " eV/Å")
+    end
+end
+
+# (Former Test 12 "6mrr protein energy, first 50 atoms" removed: the full-protein Test 13 below
```
jgreener64: Remove comment.

test/ml_potentials.jl
```diff
@@ -0,0 +1,840 @@
...
+        t = @elapsed potential_energy(sys)
+        println("  time (warm):  $(round(t, sigdigits=3)) s  ($n atoms)")
+    end
+end
+
+# ============================================================================
+# Test 14: KernelAbstractions AEV kernel — CPU and (optionally) GPU backend
+# ============================================================================
+
+if isdefined(@__MODULE__, :KernelAbstractions) || @isdefined(KernelAbstractions)
```
jgreener64: This is always defined.

test/ml_potentials.jl
```diff
@@ -0,0 +1,840 @@
...
+            println("Test15: batched energy = ", E_batched, " | per-atom ref = ", E_ref,
+                    " | diff = ", abs(E_batched - E_ref))
+        end
+    end
+end
+
+# ============================================================================
+# Test 16: KA neighbour-list AEV kernel matches the all-pairs kernel and the
+#          scalar CPU path on a multi-element system. Runs on the KA CPU backend
+#          (no GPU needed); the same kernel runs on Metal/CUDA/ROCm.
```
jgreener64, line 624: Test on GPU backends too using the same `for AT in array_list` strategy as elsewhere in the tests. We can't rely on the kernel executing the same on CPU and GPU.

test/Project.toml
```diff
@@ -12,14 +12,20 @@ DelimitedFiles = "8bb1440f-4735-579b-a4ab-409b98df4dab"
...
+JSON3 = "0f8b85d8-7281-11e9-16c2-39a750bddbf1"
+KernelAbstractions = "63c18a36-062a-441e-b654-da1e3ab1ce7c"
 KernelDensity = "5ab0869b-81aa-558d-bb23-cbf5423bbe9b"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
+Lux = "b2108857-7c20-44ae-9111-449ecde12c47"
 Measurements = "eff96d63-e80a-5855-80a2-b1b0885c5ab7"
 Metal = "dde4c033-4e86-420c-a63e-0dd931031962"
 Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
 SimpleCrystals = "64031d72-e220-11ed-1a7e-43a2532b2fa8"
+StaticArrays = "90137ffa-7385-5640-81b9-e52037218182"
```
jgreener64: StaticArrays and Unitful not required.

docs/src/examples.md
```diff
@@ -518,6 +518,42 @@ save("polymer_angle.png", f)
...
+    atoms_data = [AtomData(element="O"), AtomData(element="H"), AtomData(element="H")],
+    general_inters = (ani=pot,),
+    force_units    = u"eV/Å",
+    energy_units   = u"eV",
+)
+
+potential_energy(sys)
+forces(sys)
+```
+For larger or periodic systems, attach a neighbour finder such as [`DistanceNeighborFinder`](@ref) (or `CellListMapNeighborFinder`) — the AEV then uses the minimum-image convention and only the passed-in neighbours.
```
jgreener64: Describe briefly how to put this example on the GPU, and say that `eligible` should be all true.

src/interactions/ml_potentials.jl
```diff
@@ -0,0 +1,148 @@
+# Machine-learning interatomic potentials (core definitions).
```
jgreener64, line 1: Most of the comments in this file are not necessary, trim them down.

ext/MollyLuxExt.jl
```diff
@@ -0,0 +1,1796 @@
...
+
+module MollyLuxExt
+
+using Molly
+# celu01 / cosine_cutoff are internal (unexported) Molly helpers; import them explicitly so this
+# extension can use them by bare name without them cluttering the public Molly namespace.
+using Molly: from_device, to_device, vector, celu01, cosine_cutoff
+import AtomsCalculators
+using Lux, HDF5
+using KernelAbstractions   # strong Molly dependency; the ANI GPU kernels live in molly_lux_ka.jl
```
jgreener64: Out of date comment?

ext/MollyLuxExt.jl
```diff
@@ -0,0 +1,1796 @@
...
+#   angular_aev!   → [ANI-1] Eq. (4)  angular symmetry function G^A
+#   ani_energy_single / ani_raw_energy → [ANI-1] Eq. (1) energy sum + ensemble average
+# The element NN architecture is in build_element_model; per-atom self-energies are an
+# additive reference shift (Hartree). This extension provides the AEV + energy and the analytic
+# on-device forces (AtomsCalculators.forces! → compute_ani_forces_ka).
+
+module MollyLuxExt
+
+using Molly
+# celu01 / cosine_cutoff are internal (unexported) Molly helpers; import them explicitly so this
```
jgreener64: Unnecessary comment.

ext/MollyLuxExt.jl
```diff
@@ -0,0 +1,1796 @@
...
+using Lux, HDF5
+using KernelAbstractions   # strong Molly dependency; the ANI GPU kernels live in molly_lux_ka.jl
+using StaticArrays, Unitful, Random, LinearAlgebra
+
+# ============================================================================
+# Shared infrastructure
+# ============================================================================
+
+# Strip units from a coordinate SVector array and return plain Float SVectors.
+function strip_coords(coords::AbstractVector{SVector{D,T}}) where {D,T}
```
jgreener64: Similar functions are in ml_potentials.jl, move them all here? Is `strip_coords` used?

docs/src/examples.md
jgreener64: Give the `neighbor_finder =` line required here.

ext/MollyLuxExt.jl
```diff
@@ -0,0 +1,1796 @@
...
+        idx[pj] = i
+    end
+end
+
+# Device-side counterpart of `nl_to_csr`: build the CSR (offsets, indices) entirely on the compute
+# backend so no host-side O(total_pairs) counting sort is needed (that host loop was the largest
+# single cost of a GPU energy/forces call — ~21 ms at 15,954 atoms on the RTX 5080 host). Bulk-upload
+# the pair list, then histogram + prefix-sum + atomic scatter on device.
+function nl_to_csr_device(neighbors, n_atoms::Int, ka_backend)
+    list   = neighbors.list
```
jgreener64: The length of this list is `length(neighbors)`, not `length(list)`, and the list needs slicing up to this length.

ext/MollyLuxExt.jl
```diff
@@ -0,0 +1,1796 @@
...
+    # Lux.setup are discarded and replaced by `ps` loaded from HDF5, so the RNG seed is
+    # irrelevant to the result — a fixed Xoshiro(0) just makes construction deterministic.
+    _, st = Lux.setup(Xoshiro(0), model)
+    return model, ps, st
+end
+
+# Read an HDF5 dataset as a Vector{T}, handling scalar datasets gracefully.
+h5vec(grp, name, T) = T.(vcat(read(grp[name])))
+
+function Molly.ANIPotential(path::String;
```
jgreener64: `::AbstractString`, and possibly elsewhere.

ext/MollyLuxExt.jl
```diff
@@ -0,0 +1,1796 @@
...
+# In-place coordinate stripping into a pre-allocated buffer — zero allocations.
+function strip_coords_into!(out::AbstractVector{SVector{D,TF}},
+                              coords::AbstractVector{SVector{D,T}}) where {D, TF, T}
+    if unit(first(coords)[1]) == NoUnits
+        @inbounds for i in eachindex(coords)
+            out[i] = SVector{D,TF}(coords[i])
+        end
+    else
+        @inbounds for i in eachindex(coords)
+            out[i] = SVector{D,TF}(ustrip.(u"Å", coords[i]))
```
jgreener64, line 63: Molly convention is to treat unitless coordinates as nm, make sure this is consistent throughout.

ext/MollyLuxExt.jl
```diff
@@ -0,0 +1,1796 @@
...
+
+# The ANI forces path. Strips coords to Å (staying on their device — Array → KA CPU,
+# GPU array → Metal/CUDA), computes forces via the analytic backward, and accumulates into `fs`
+# in the system's units.
+function AtomsCalculators.forces!(fs, sys::System{D, AT, T}, inter::ANIPotential;
+                                  kwargs...) where {D, AT, T}
+    nbrs    = get(kwargs, :neighbors, nothing)
+    n_sp    = length(inter.species_map)
+    sp      = Int32[inter.species_map[ad.element] for ad in sys.atoms_data]
+    coords  = Molly.ustrip_vec.(u"Å", sys.coords)       # Å, unitless; stays on coords' device
```
jgreener64: This errors in the unitless case? Fix and add a unitless test, assuming nm.

src/interactions/ml_potentials.jl
```diff
@@ -0,0 +1,148 @@
...
+"""
+struct ANIPotential{M, PV, SV, SP, P, SE, D, F, E} <: AbstractMLPotential
+    model::M          # NamedTuple of per-element Lux.Chain sub-networks (shared architecture)
+    ps_vec::PV        # Vector of per-element parameter NamedTuples, one per ensemble member
+    st_vec::SV        # Vector of per-element state NamedTuples, one per ensemble member
+    species_map::SP   # Dict{String,Int}: element → 1-based index
+    aev_params::P     # NamedTuple: η_R, r_s_R, r_c_R, η_A, r_s_A (ShfA, Å), θ_s (ShfZ, rad), ζ, r_c_A
+    self_energies::SE # Vector: atomic self-energy per species (Hartree)
+    cutoff::D         # max(r_c_R, r_c_A), plain Float (Å)
+    force_units::F
```
jgreener64: Are the units fields ever used?

ext/MollyLuxExt.jl
```diff
@@ -0,0 +1,1796 @@
...
+    # Per-species atom-index groups; uploaded to the compute device for the AEV gather.
+    groups  = [Int32.(findall(==(s), sp_host)) for s in 1:n_species]
+    idx_dev = map(groups) do g
+        (on_gpu && !isempty(g)) || return g
+        d = KernelAbstractions.allocate(ka_backend, Int32, length(g)); copyto!(d, g); d
+    end
+
+    n_ens = length(pot.ps_vec)
+    ps_dev_all, st_dev_all = ani_nn_dev_params(pot, dev)             # cached on-device weights
+    # The per-element NN outputs are Float32, but the total energy sums O(10^4) per-atom terms of
```
jgreener64, line 1597: Is this also followed on the CPU path?

jgreener64 (commented):
Looks good, here are some comments.

The h5 file should be a lazy artifact. The forces for the 6mrr json could be stored in a compressed format? Or added to the artifact.

src/interactions/ml_potentials.jl
```diff
@@ -0,0 +1,148 @@
...
+# Convert an ANI force SVector (eV/Å, unitless) to the system's force units and device.
+function ani_force_to_units(fi::SVector{D,T}, force_units, ::Type{AT}) where {D, T, AT}
+    f_unit = if force_units == NoUnits
+        fi
+    elseif dimension(force_units) == u"𝐋 * 𝐌 * 𝐍^-1 * 𝐓^-2"
+        uconvert.(force_units, fi .* (Unitful.Na * u"eV/Å"))
+    else
+        uconvert.(force_units, fi .* u"eV/Å")
+    end
+    return to_device(f_unit, AT)
```
jgreener64: `to_device` not needed.

ext/MollyLuxExt.jl
```diff
@@ -0,0 +1,1796 @@
...
+    bwd_nb  = isnothing(csr) ? :auto : csr
+    dcoords = aev_vjp_ka(dEdAEV, coords, species, pot.aev_params, n_species;
+        backend=ka_backend, neighbors=bwd_nb, boundary=boundary)   # ∂E_Ha/∂r, (3, n_atoms)
+    # F = -∂E/∂r · (Ha→eV). `T` is a static type parameter from the coords signature, so the
+    # per-atom SVector build is type-stable. (Deriving it as `eltype(eltype(coords))` — a runtime
+    # value — made this loop dynamically dispatched, ~100 ms for 16k atoms on its own.)
+    Ha   = T(Molly.HARTREE_TO_EV)
+    fmat = Array(dcoords)
+    na   = size(fmat, 2)
+    return [SVector{3,T}(-fmat[1,i]*Ha, -fmat[2,i]*Ha, -fmat[3,i]*Ha) for i in 1:na]
```
jgreener64: Does this do an extra host-device transfer on GPU?

src/interactions/ml_potentials.jl
```diff
@@ -0,0 +1,148 @@
...
+    r < r_c ? T(0.5) * (one(T) + cos(T(π) * r / r_c)) : zero(T)
+end
+
+# celu01(x)
+#
+# CELU activation with α=0.1 — the nonlinearity between Dense layers of each ANI element
+# network (matches the TorchANI ANI-2x architecture). `celu01(x) = x` for `x ≥ 0`,
+# `0.1*(exp(x/0.1) - 1)` otherwise. A core-Molly function (Molly.celu01) so AD backends can
+# register rules without depending on MollyLuxExt. Internal helper (not part of the public API).
+@noinline celu01(x::T) where T = x >= zero(T) ? x : T(0.1) * (exp(x / T(0.1)) - one(T))
```
jgreener64: `@noinline` no longer required?

ext/MollyLuxExt.jl
```diff
@@ -0,0 +1,1796 @@
...
+end
+
+# The ANI forces path. Strips coords to Å (staying on their device — Array → KA CPU,
+# GPU array → Metal/CUDA), computes forces via the analytic backward, and accumulates into `fs`
+# in the system's units.
+function AtomsCalculators.forces!(fs, sys::System{D, AT, T}, inter::ANIPotential;
```
jgreener64, line 1789: Use the `n_threads` kwarg passed here, and to `potential_energy`.

test/ml_potentials.jl
```diff
@@ -0,0 +1,840 @@
...
+            idx_to_elem = Dict(v => k for (k, v) in pot.species_map)
+            E_ha = 0.0
+            for i in 1:n
+                s   = species[i]; sym = Symbol(idx_to_elem[s])
+                out, _ = Lux.apply(getfield(pot.model, sym),
+                                   reshape(Float32.(@view aevs_serial[i, :]), :, 1),
+                                   getfield(pot.ps_vec[1], sym), getfield(pot.st_vec[1], sym))
+                E_ha += Float64(out[1]) + pot.self_energies[s]
+            end
+            E_ref = E_ha * 27.211396132
```
jgreener64: Use the Molly constant.

src/interactions/ml_potentials.jl
```diff
@@ -0,0 +1,148 @@
...
+#
+# See ext/MollyLuxExt.jl for the ANI method overview and per-equation references
+# (ANI-1: Smith et al., Chem. Sci. 2017; ANI-2x: Devereux et al., JCTC 2020).
+
+# Base type for ML interatomic potentials. Kept as a shared supertype for current and
+# future ML potentials (ANIPotential, ...); useful for dispatch/`isa` checks.
+abstract type AbstractMLPotential end
+
+# ANI energies are produced in Hartree; MD in Molly is typically in eV.
+const HARTREE_TO_EV = 27.211396132
```
jgreener64: `ustrip(u"eV", 1u"Eh_au")`?

src/interactions/ml_potentials.jl
```diff
@@ -0,0 +1,148 @@
...
+    unit(b.side_lengths[1]) == NoUnits ? b : CubicBoundary(ustrip.(u"Å", b.side_lengths))
+
+function strip_boundary(b::TriclinicBoundary{D, T, C, A}) where {D, T, C, A}
+    unit(b.basis_vectors[1][1]) == NoUnits && return b
+    bv = SVector(ntuple(i -> ustrip.(u"Å", b.basis_vectors[i]), 3))
+    return TriclinicBoundary(bv; approx_images=A)
+end
+
+"""
+    ANIPotential(path; force_units, energy_units, T, ensemble_idx)
```
jgreener64: List supported elements, say `atoms_data` is required.

`ensemble_idx` should be made one-indexed to match Julia.

src/interactions/ml_potentials.jl
```diff
@@ -0,0 +1,148 @@
...
+
+Load an ANI neural network potential from an HDF5 file exported by
+`test/torchani_reference.py`. Requires `Lux` and `HDF5` to be loaded.
+
+By default all ensemble members are loaded and energies are averaged.
+Pass `ensemble_idx` to load only a specific member (wrapped in a length-1 vector).
+
+Note: the ANI-2x weights are `Float32`, so the energy/force paths run in `Float32`
```
jgreener64, line 43: Just checking this is true, since `T` is used in the kernels.

test/Project.toml
```diff
@@ -12,14 +12,20 @@ DelimitedFiles = "8bb1440f-4735-579b-a4ab-409b98df4dab"
 Enzyme = "7da242da-08ed-463a-9acd-ee780be4f1d9"
 FiniteDifferences = "26cc04aa-876d-5657-8c51-4c34ba976000"
 GLMakie = "e9467ef8-e4e7-5192-8a1a-b1aee30e663a"
+HDF5 = "f67ccb44-e63f-5c2f-98bd-6dc0ccc4ba2f"
```
jgreener64: Not in alphabetical order, also add compat bounds for new entries.

ext/MollyLuxExt.jl
```diff
@@ -0,0 +1,1796 @@
...
+                                    split, aev_len)
+        end
+    end
+    return @view buf.aevs[1:n_atoms, :]
+end
+
+# Public interface: allocating version (used by potential_energy and for standalone use).
+# Same AEV equations as the buffered path: radial block = [ANI-1] Eq. 3 (G^R),
+# angular block = [ANI-1] Eq. 4 (G^A), both using the f_C cutoff of [ANI-1] Eq. 2.
+function Molly.compute_aevs(coords::AbstractVector{SVector{D,T}},
```
jgreener64, line 434: Code duplication? Can this wrap another function?

docs/src/api.md
sivasathyaseeelan: Done, trimmed to just the `using Lux, HDF5` note.

src/types.jl
sivasathyaseeelan: Done. Moved the exports to ml_potentials.jl and no longer export `AbstractMLPotential`.

test/runtests.jl
sivasathyaseeelan: The h5 is now a lazily-downloaded artifact, so it is always fetched (a download failure errors clearly) and the file-missing branch is gone.

test/ml_potentials.jl
sivasathyaseeelan: Done. Imports moved to runtests.jl; dropped StaticArrays/Unitful since Molly re-exports them.

test/ml_potentials.jl
sivasathyaseeelan: Trimmed it to a focused energy-vs-TorchANI check (dropped the trivial sign assertion). Happy to remove it entirely if you prefer, since Test 13 also validates energy vs TorchANI.

test/ml_potentials.jl
sivasathyaseeelan: Removed.

test/ml_potentials.jl
sivasathyaseeelan: Removed.

test/ml_potentials.jl
sivasathyaseeelan: Removed the always-true KernelAbstractions guards.

test/ml_potentials.jl
sivasathyaseeelan: Added `for AT in array_list` to the KA AEV-kernel test and the neighbour-list kernel comparisons. Locally only Array runs; the GPU backends run on CI.

test/Project.toml
sivasathyaseeelan: Removed StaticArrays and Unitful from the test deps.

docs/src/examples.md
sivasathyaseeelan: Added a short GPU section (build the System with a GPU array type for the coordinates) and noted that `eligible` should be all true.

src/interactions/ml_potentials.jl
sivasathyaseeelan: Trimmed the comments in this file.

ext/MollyLuxExt.jl
sivasathyaseeelan: Removed (it referenced a file that does not exist).

ext/MollyLuxExt.jl
sivasathyaseeelan: Removed.

ext/MollyLuxExt.jl
sivasathyaseeelan: `strip_coords` was unused, so I removed it. Moved the unit-plumbing helpers (strip_boundary, ani_energy_to_units, ani_force_to_units) into the extension alongside the others.

docs/src/examples.md
sivasathyaseeelan: Added the concrete `neighbor_finder = DistanceNeighborFinder(eligible = trues(n, n), dist_cutoff = ...)` line.

ext/MollyLuxExt.jl
sivasathyaseeelan: Fixed. It now uses `length(neighbors)` and copies only that valid prefix of the list to the device.

ext/MollyLuxExt.jl
sivasathyaseeelan: Done (`::AbstractString`). This also needed the core fallback constructor to take an untyped `path`, so the extension method stays strictly more specific (a same-signature overwrite is an error during precompilation in Julia 1.12).

ext/MollyLuxExt.jl
sivasathyaseeelan: Done. Unitless coordinates and boundaries are treated as nm at the entry points; the kernels always receive Angstrom.

ext/MollyLuxExt.jl
sivasathyaseeelan: Fixed: it no longer errors, unitless coordinates are read as nm, and I added a unitless-nm test.

src/interactions/ml_potentials.jl
sivasathyaseeelan: They were not used (energy/forces read `sys.energy_units` / `sys.force_units`), so I removed the fields.

ext/MollyLuxExt.jl
sivasathyaseeelan: The CPU path now also accumulates the energy in Float64.

src/interactions/ml_potentials.jl
sivasathyaseeelan: Removed.

ext/MollyLuxExt.jl
sivasathyaseeelan: Yes: for a GPU system the forces are assembled on the host and re-uploaded, so there is a host round-trip. It is an optimization rather than a correctness issue, so I left it as a follow-up, but I can move the unit conversion on-device if you would like.

src/interactions/ml_potentials.jl
sivasathyaseeelan: Removed; the tests pass without it.

ext/MollyLuxExt.jl
sivasathyaseeelan: Done. `potential_energy` and `forces!` now read `n_threads` and thread it through the CPU AEV loop / host assembly.

test/ml_potentials.jl
sivasathyaseeelan: Done, uses `Molly.HARTREE_TO_EV`.

src/interactions/ml_potentials.jl
sivasathyaseeelan: Done: `const HARTREE_TO_EV = ustrip(u"eV", 1u"Eh_au")`.

src/interactions/ml_potentials.jl
sivasathyaseeelan: The docstring now lists the supported elements and notes that `atoms_data` is required, and `ensemble_idx` is one-indexed (1:8).

src/interactions/ml_potentials.jl
sivasathyaseeelan: Confirmed: the ANI-2x weights are Float32, so the energy/force paths run in Float32 internally regardless of the coordinate type.

test/Project.toml
sivasathyaseeelan: Sorted alphabetically and added compat bounds for HDF5, KernelAbstractions and Lux.

ext/MollyLuxExt.jl
sivasathyaseeelan: Done. The allocating `compute_aevs` now wraps the buffered path (`compute_aevs_buf!`, serial), so the per-atom loop lives in one place.

sivasathyaseeelan:
Thanks for the review! I have pushed two commits addressing all the comments (inline replies on each thread):

- `f5c7806d` — code, tests and docs
- `b07c73a8` — the lazy artifact

**Lazy artifact (main comment):** `ani2x.h5` (55 MB) and the large `6mrr_ani2x.json` are no longer committed. They are packaged as a lazily-downloaded artifact (`ani2x` in `Artifacts.toml`) fetched on first use, exposed via the exported `ani2x_data_dir()`; the tests and the docs example read the weights from there. The tarball is currently hosted on a GitHub release of my fork — happy to move it under the JuliaMolSim org before merge.

**Highlights:** unitless coordinates now follow the nm convention at the entry points (fixing the case that errored) while the kernels stay Angstrom-native; fixed the `nl_to_csr_device` length bug; `HARTREE_TO_EV = ustrip(u"eV", 1u"Eh_au")`; removed the unused units fields; one-indexed `ensemble_idx`; removed the always-true KernelAbstractions guards; the AEV-kernel tests now run over `array_list` for GPU coverage; CPU energy now accumulates in Float64.

All ANI tests pass locally. Two small things I flagged in the threads: I kept a slimmed N₂ energy-vs-TorchANI test rather than deleting it outright (happy to remove it), and the GPU forces path still has a host round-trip which I left as a follow-up optimization.

sivasathyaseeelan:
> The h5 file should be a lazy artifact. The forces for the 6mrr json could be stored in a compressed format? Or added to the artifact.

Both done in `b07c73a8`:

- `ani2x.h5` is now a lazy artifact (`ani2x` in `Artifacts.toml`, `lazy = true`), fetched on first use and exposed via the exported `ani2x_data_dir()`.
- `6mrr_ani2x.json` is added to the same artifact, so it is no longer committed either. It rides inside the gzip-compressed tarball, so it is effectively stored compressed as well (~1.5 MB JSON, and the whole artifact tarball is ~52 MB vs the 55 MB uncommitted).

The tests and the docs example now read both from `ani2x_data_dir()`. The tarball is on a release of my fork for now (`ani2x-data-v1`); happy to move it under the JuliaMolSim org before merge, which is just a re-upload plus updating the URL and sha256 in `Artifacts.toml`.

test/ml_potentials.jl
```diff
@@ -0,0 +1,836 @@
...
+        dev = maximum(maximum(abs.(fs_val[i] .- f_ref[i])) for i in 1:3)
+        println("Max force deviation: ", round(dev, sigdigits=4), " eV/Å")
+    end
+end
+
+# ============================================================================
+# Test 13: 6mrr full protein — all 15,954 atoms, cross-validate with TorchANI
+# ============================================================================
+
+@testset "ANIPotential: 6mrr full protein energy (all atoms)" begin
```
jgreener64, line 448: Probably worth doing this test on the GPU too. Also, it would be good to run a short (10-step?) simulation on CPU and GPU and check that the coordinates at the end are the same.

test/ml_potentials.jl
```diff
@@ -0,0 +1,836 @@
...
+    pdb_path = joinpath(@__DIR__, "..", "data", "6mrr_equil.pdb")
+    ref_path = joinpath(ANI_DIR, "6mrr_ani2x.json")
+    h5_path  = joinpath(ANI_DIR, "ani2x.h5")
+    if !isfile(pdb_path) || !isfile(h5_path)
+        @warn "6mrr_equil.pdb or ani2x.h5 not found — skipping"
```
jgreener64, line 453: I think the file should always be present at this point? It should always be there during testing if not.

Artifacts.toml
```diff
@@ -0,0 +1,7 @@
+[ani2x]
+git-tree-sha1 = "0d49f491bd9ec50b1c39aad84bb2bab74a62471d"
+lazy = true
+
+    [[ani2x.download]]
+    sha256 = "9027c79156ea0de5b5435990c6aa903b4de991350416fdb847ede91964761ae2"
+    url = "https://github.com/sivasathyaseeelan/Molly.jl/releases/download/ani2x-data-v1/ani2x_data.tar.gz"
```
jgreener64, line 7: Just to check, the large tarball will not be added to the repo if it was added then removed in this PR?

ext/MollyLuxExt.jl
```diff
@@ -0,0 +1,1819 @@
...
+            hi = c * base + min(c, rem)
+            lo <= hi && aev_chunk!(buf, lo:hi, buf.scratch[c], coords, species_indices,
+                                    use_nl, boundary, p, n_species, r_c_R, r_c_A, r_c_max,
+                                    split, aev_len)
+        end
+    end
+    return @view buf.aevs[1:n_atoms, :]
+end
+
+# Public interface: allocating version (used as the reference AEV in tests and for standalone
```
jgreener64, line 431: This should have a docstring if it is the public path. Does this make more sense as a public path than `compute_aevs_ka`, where the suffix is an implementation detail?

src/interactions/ml_potentials.jl
```diff
@@ -0,0 +1,93 @@
...
+"""
+function ani2x_data_dir end
+
+# Compute atomic environment vectors (AEVs) for all atoms, returning an
+# (n_atoms, aev_length) matrix. Implementation is in ext/MollyLuxExt.jl.
+function compute_aevs end
+
+# GPU-portable AEV computation with KernelAbstractions (implementation in ext/MollyLuxExt.jl).
+function compute_aevs_ka(args...; kwargs...)
+    error("compute_aevs_ka requires KernelAbstractions, Lux, and HDF5 to be loaded: " *
```
jgreener64, line 79: KernelAbstractions not required as it is always present.

ext/MollyLuxExt.jl
```diff
@@ -0,0 +1,1819 @@
...
+# Radial AEV — radial symmetry function G^R, [ANI-1] Eq. (3):
+#   G^R_{η,R_s} = Σ_{j≠i} exp(−η·(R_ij − R_s)²) · f_C(R_ij)
+# One element per (η_R, r_s_R) pair, computed separately per neighbour species (the
+# `base` offset selects the species block). ANI-2x uses one η_R and 16 shifts R_s.
+# The 0.25 prefactor follows the TorchANI reference implementation (not in the paper).
+# Writes into G (a pre-allocated view of the AEV matrix row);
+# nbr_coords/nbr_species hold the first n_nbr neighbours.
+function radial_aev!(G::AbstractVector{T}, coord_i::SVector{D,T},
+                      nbr_coords, nbr_species, n_nbr::Int,
+                      η_R, r_s_R, r_c_R::T, n_species::Int) where {D,T}
```
jgreener64, line 133: `n_species` not used?

docs/src/examples.md
```diff
@@ -518,6 +518,57 @@ save("polymer_angle.png", f)
...
+    force_units    = u"eV/Å",
+    energy_units   = u"eV",
+)
+
+potential_energy(sys)
+forces(sys)
+```
+For larger or periodic systems, attach a neighbour finder so the AEV uses the minimum-image convention and only the passed-in neighbours. When all pairs are eligible, set `eligible` to all true:
+```julia
+n = length(coords)
```
jgreener64, line 549: This errors.

test/ml_potentials.jl
```diff
@@ -0,0 +1,836 @@
...
+
+        # Assign random velocities at 300 K (seeded for reproducibility)
+        random_velocities!(sys, 300.0u"K"; rng=MersenneTwister(42))
+
+        E0     = total_energy(sys)
+        simulate!(sys, VelocityVerlet(dt=0.1u"fs"), 50)
+        E_final = total_energy(sys)
+
+        # NVE: total energy should be approximately conserved. This is a fast in-suite smoke test
+        # (a stringent, longer conservation test lives in benchmark/ani_trajectory.jl: 3000 steps
```
jgreener64, line 349: Referencing a file that doesn't exist.

test/torchani_reference.py
```diff
@@ -0,0 +1,429 @@
...
+    with h5py.File(output, "w") as h5:
+        # ---- AEV hyperparameters ----
+        ag = h5.create_group("aev_params")
+        ag.create_dataset("Rcr",     data=float(aev.Rcr))
+        ag.create_dataset("Rca",     data=float(aev.Rca))
+        ag.create_dataset("EtaR",    data=aev.EtaR.squeeze().numpy())
+        ag.create_dataset("ShfR",    data=aev.ShfR.squeeze().numpy())
+        ag.create_dataset("EtaA",    data=aev.EtaA.squeeze().numpy())
+        ag.create_dataset("Zeta",    data=float(aev.Zeta.item()))
+        ag.create_dataset("ShfA",    data=aev.ShfA.squeeze().numpy())   # angular shift (θ_s)
```
jgreener64, line 190: Labels the wrong way.

test/ml_potentials.jl
```diff
@@ -0,0 +1,836 @@
...
+        @test_broken false
+    else
+        pot = ANIPotential(h5_path; ensemble_idx=1)
+        valid_elements = Set(keys(pot.species_map))
+
+        # Parse all atoms from PDB
+        coords_list = SVector{3,Float64}[]
+        elem_list   = String[]
+        open(pdb_path) do f
+            for line in eachline(f)
```
jgreener64, line 463: Use BioStructures for this parsing.

test/ml_potentials.jl
```diff
@@ -0,0 +1,836 @@
...
+                x = parse(Float64, line[31:38])
+                y = parse(Float64, line[39:46])
+                z = parse(Float64, line[47:54])
+                push!(coords_list, SVector(x, y, z))
+                push!(elem_list, elem)
+            end
+        end
+
+        n = length(elem_list)
+        println("6mrr: loaded $n valid ANI-2x atoms")
```
jgreener64, line 477: No `println` in tests.

ext/MollyLuxExt.jl
```diff
@@ -0,0 +1,1819 @@
...
+# ============================================================================
+# The ANI parameters are in Å, so coordinates and boundaries are converted to unitless Å
+# before the kernels. Following the Molly convention, unitless input is treated as nm.
+
+const NM_TO_ANGSTROM = 10
+
+# In-place conversion of coordinates to unitless Å in a pre-allocated buffer (zero allocations).
+function coords_to_angstrom_into!(out::AbstractVector{SVector{D,TF}},
+                                  coords::AbstractVector{SVector{D,T}}) where {D, TF, T}
+    if unit(first(coords)[1]) == NoUnits   # unitless Molly coords are nm
```
jgreener64, line 57: Doesn't this error on GPU? Have you run the tests on GPU? They are not run on CI.

ext/MollyLuxExt.jl
```diff
@@ -0,0 +1,1819 @@
...
+                    use_nl, boundary, p, n_species, r_c_R, r_c_A, r_c_max, split, aev_len)
+    else
+        base = n_atoms ÷ nchunks
+        rem  = n_atoms % nchunks
+        # `:static` fixes the chunk→iteration mapping so chunk c always uses scratch[c]
+        # (its own buffer, no sharing). `nchunks <= length(buf.scratch) == nthreads`, so each
+        # chunk has a private scratch. It is not about load balancing — the chunks are
+        # equal-sized. (Plain @threads would also be correct since the scratch is per-chunk, but
+        # :static avoids the dynamic-scheduler overhead for this fixed, uniform partition.)
+        Threads.@threads :static for c in 1:nchunks
```
jgreener64, line 420: Best not to use `:static` unless you can show a performance gain. `@spawn` is an alternative.

ext/MollyLuxExt.jl
```diff
@@ -0,0 +1,1819 @@
...
+    n_atoms    :: Int,
+    n_eta_R    :: Int,
+    n_shf_R    :: Int,
+    n_eta_A    :: Int,
+    n_shf_r    :: Int,
+    n_th       :: Int,
+    split      :: Int,
+    ::Val{L},
+) where {L}
+    acc = @localmem Float32 (L,)
```
jgreener64, line 925: Is there a discrepancy between CPU and GPU paths here? A Float64 GPU system gets Float32 AEVs, whereas the same system on CPU gets Float64.

ext/MollyLuxExt.jl
```diff
@@ -0,0 +1,1819 @@
...
+    if unit(b.basis_vectors[1][1]) == NoUnits
+        bv = SVector(ntuple(i -> b.basis_vectors[i] .* NM_TO_ANGSTROM, 3))
+    else
+        bv = SVector(ntuple(i -> ustrip.(u"Å", b.basis_vectors[i]), 3))
+    end
+    return TriclinicBoundary(bv; approx_images=A)
+end
+
+# Convert an ANI energy (eV, unitless) to the system's energy units.
+function ani_energy_to_units(E_eV, energy_units)
```
jgreener64, line 89: Molly convention is kJ/mol in the no units case, and kJ/mol/nm for force.

test/ml_potentials.jl
```diff
@@ -0,0 +1,836 @@
...
+        aevs_n2_cpu = Molly.compute_aevs(coords_n2, species_n2, nothing, bdy, p, n_sp)
+
+        water_ref  = JSON3.read(read(joinpath(REF_DIR, "water_ani2x.json"), String))
+        coords_w   = [SVector{3,Float32}(row...) for row in water_ref.coordinates_A]
+        species_w  = [pot.species_map[s] for s in ["O","H","H"]]
+        aevs_w_cpu = Molly.compute_aevs(coords_w, species_w, nothing, bdy, p, n_sp)
+
+        # The KA kernel must match the scalar CPU path on every backend — CPU and each GPU in
+        # array_list. GPU kernels can diverge from the CPU path, so exercise them all.
+        for AT in array_list
```
jgreener64, line 539: Also test the GPU path from setup, i.e. passing GPU coordinates and general interaction to `System` then calling `potential_energy` and `forces`.

ext/MollyLuxExt.jl
```diff
@@ -0,0 +1,1819 @@
...
+    end
+end
+
+# Convert a Molly NeighborList (flat half-pairs (i,j,special)) into a host CSR
+# (offsets, symmetrised indices). This is the production path: it *consumes the
+# neighbours produced by the system's neighbour finder* (CellListMapNeighborFinder on
+# CPU, DistanceNeighborFinder on GPU) rather than building its own list. `neighbors.list`
+# may be device-resident (GPU finder) — `Molly.from_device` brings the pairs to the host for the
+# O(total_pairs) counting sort; offsets/indices are then uploaded to the compute backend.
+function nl_to_csr(neighbors, n_atoms::Int)
```
jgreener64, line 1312: Isn't this similar to `fill_csr!` above? Remove duplication if possible.

ext/MollyLuxExt.jl
```diff
@@ -0,0 +1,1819 @@
...
+# `vector` return the raw displacement. Metal is Float32-only, so the element type must match.
+boundary_for_kernel(::Nothing, ::Val{D}, ::Type{T}) where {D, T} =
+    CubicBoundary(SVector{D, T}(ntuple(_ -> T(Inf), D)))
+boundary_for_kernel(b, ::Val{D}, ::Type{T}) where {D, T} =
+    convert_boundary_eltype(ustrip_boundary(b), T)
+
+# Low-level unit stripping: convert a unit-carrying boundary to Å; a unitless boundary is
+# assumed to already be in Å. (The nm→Å convention for unitless input is applied only at the
+# high-level entry points via `strip_boundary`, so the kernels always receive Å.)
+ustrip_boundary(b::CubicBoundary) =
```
jgreener64, line 683: Isn't this similar to `strip_boundary` but assumes Å? Remove duplication and assume nm.

jgreener64 (commented):
Almost there, hopefully this is the last review round.

There is still quite a bit of code duplication. The AEV equations exist in four similar places, and the angular backward in two. Try to reuse code via functions where possible.

test/ml_potentials.jl
sivasathyaseeelan: Right — with the h5 as an artifact (always fetched) and the PDB committed, the files are always present here, so I removed the isfile gate in this test.

Artifacts.toml
sivasathyaseeelan: Correct: the tarball is a GitHub release asset, not committed to the repo. The h5 blob was added and then removed across commits in this PR, so it is absent from the branch tip; a squash-merge drops it from history entirely. Happy to squash the ANI commits if you'd prefer a clean history.

ext/MollyLuxExt.jl
sivasathyaseeelan: Added a docstring to `compute_aevs` and framed it as the public reference AEV path; `compute_aevs_ka` is now documented as the GPU-portable kernel version (the `_ka` suffix being an implementation detail).

src/interactions/ml_potentials.jl
sivasathyaseeelan: Right — dropped KernelAbstractions from the stub error messages (it is a strong Molly dependency, always present); only Lux and HDF5 gate the extension.

ext/MollyLuxExt.jl
sivasathyaseeelan: Removed the unused `n_species` argument from `radial_aev!` and its caller.

test/ml_potentials.jl
sivasathyaseeelan: Removed the comment (it referenced `benchmark/ani_trajectory.jl`, which is not part of this PR).

test/torchani_reference.py
sivasathyaseeelan: Fixed: swapped the `ShfA`/`ShfZ` comments — `ShfA` is the radial shift `r_s_A`, `ShfZ` is the angular shift `θ_s`, matching the loader.

test/ml_potentials.jl
sivasathyaseeelan: Switched the PDB parsing to BioStructures, via a shared `load_6mrr_atoms` helper used by the protein tests.

test/ml_potentials.jl
sivasathyaseeelan: Removed all `println`s from the tests.

ext/MollyLuxExt.jl
sivasathyaseeelan: Changed to a plain `Threads.@threads` (each chunk uses its own scratch set, so the loop is race-free without `:static`, and I did not have a measured gain to justify it).

ext/MollyLuxExt.jl
sivasathyaseeelan: Fixed: the no-units convention now returns kJ/mol for energy and kJ/mol/nm for force.

ext/MollyLuxExt.jl
sivasathyaseeelan: Deduplicated: `nl_to_csr` now allocates the arrays and delegates to `fill_csr!` for the shared host CSR build.

ext/MollyLuxExt.jl
sivasathyaseeelan: Removed `ustrip_boundary`. The single nm-aware `strip_boundary` runs once at the entry points (`potential_energy`/`forces!`); `boundary_for_kernel` now only matches the element type, since the boundary reaching it has already been stripped to unitless Å.

test/ml_potentials.jl
sivasathyaseeelan: Added Test 20: it builds the System with GPU coordinates and the ANI interaction, checks energy and forces against the CPU path, and runs a 10-step CPU-vs-GPU simulation and checks the final coordinates match. Verified on an RTX 5080 (CUDA): the ANI test suite passes on both `Array` and `CuArray`.

ext/MollyLuxExt.jl
sivasathyaseeelan: You were right. `potential_energy` went through the host CPU buffered path (which scalar-indexes the coordinates) for every system, so it errored on GPU. It now routes GPU systems to the on-device energy, and `coords_to_angstrom` detects units by element type rather than indexing. I ran the ANI test suite on an RTX 5080 (CUDA) — `array_list = (Array, CuArray)`, `CUDA.functional() = true`, all 19 testsets pass. (CI has no GPU, so these run on the GPU box.)

ext/MollyLuxExt.jl
sivasathyaseeelan: The AEV and energy now run in Float32 on both paths (the ANI-2x model is Float32): the CPU energy buffer is Float32 regardless of the coordinate type, matching the GPU kernel's Float32 accumulator. Verified consistent on an RTX 5080 (Test 20 checks GPU energy/forces against CPU).

test/ml_potentials.jl
sivasathyaseeelan: Added as Test 20 and verified on an RTX 5080: build the System with GPU coordinates and the ANI interaction, then `potential_energy` and `forces` match the CPU path, and a short CPU/GPU simulation reaches the same coordinates.

docs/src/examples.md
sivasathyaseeelan: Fixed: the snippet now moves the atoms, coordinates and velocities to the device together. The System-level GPU path (build with GPU arrays, then `potential_energy`/`forces`) is verified on an RTX 5080.

sivasathyaseeelan:
Addressed the second review round (commits `83bb8203` + `36e385b8`), replied inline on each thread.

Highlights:
- **Units**: the no-units convention now returns kJ/mol (energy) and kJ/mol/nm (force); removed the `ustrip_boundary`/`strip_boundary` duplication.
- **GPU path**: `potential_energy` now routes GPU systems to the on-device energy (the host path scalar-indexed the coordinates and errored on GPU), and the unit check is by element type rather than indexing. The AEV + energy run in Float32 on both paths.
- **Tests**: added Test 20 (build the System with GPU coordinates + the ANI interaction, then check energy, forces and a short CPU-vs-GPU simulation); switched PDB parsing to BioStructures; removed the always-present gates and all `println`s.
- **Cleanups**: `compute_aevs` docstring, `nl_to_csr`→`fill_csr!` dedup, `@threads` instead of `:static`, unused-arg and stub-message fixes, torchani label + docs GPU snippet.

I ran the ANI test suite on an **RTX 5080 (CUDA)** to answer the "have you run on GPU?" question: `array_list = (Array, CuArray)`, `CUDA.functional() = true`, and all 19 testsets pass. (CI has no GPU, so these run on the GPU box.)

jgreener64:
Thanks for all your work on this. Doing a squash merge to avoid putting the large file into the repo history.

PR #283, opened 2026-08-18
Native Allegro equivariant potential (energy + analytic forces, CPU/GPU)

sivasathyaseeelan:
## Summary

A native-Julia **Allegro** interatomic potential (Musaelian et al. 2023), the equivariant-GNN counterpart to the merged native ANI-2x (#281). Allegro is strictly local — its energy is a sum over directed edges within a cutoff — and this PR builds it end to end: the O(3)-equivariant machinery (which has no mature Julia implementation), the energy, analytic forces, and GPU support.

Everything is validated against [e3nn](https://github.com/e3nn/e3nn) (used offline only, as an oracle — Molly itself has **no Python runtime dependency**, matching the ANI approach).

## What's here

**Equivariant primitives** — new core module `src/equivariant/` (depends only on StaticArrays + KernelAbstractions, so it compiles and tests with a bare `using Molly`):
- `irreps.jl` — `Irrep`/`Irreps` (channel-major layout) + `TensorProductPaths` (allowed CG couplings).
- `spherical_harmonics.jl` — component-normalized real spherical harmonics (l ≤ 2) with an analytic Jacobian, **pinned to e3nn to ~1e-15**.
- `radial_basis.jl` — Bessel basis + DimeNet envelope, with derivatives.
- `clebsch_gordan.jl` — sparse real Clebsch-Gordan coefficients derived from complex CG; **match e3nn's `o3.wigner_3j` bit-for-bit for every coupling with l ≤ 2**.
- `tensor_product.jl` / `linear.jl` — weighted CG tensor product and equivariant linear layer, each with its analytic VJP.

**Model + wiring:**
- `src/equivariant/allegro_model.jl` — the strictly-local Allegro energy forward (two-body Bessel/SH embedding → N tensor-product layers → per-edge readout) and the analytic backward (`F = -∂E/∂r`) composing the primitive VJPs; `ΣF = 0` by construction.
- `ext/MollyAllegroExt.jl` — loads a model from HDF5 and implements `AtomsCalculators.potential_energy` / `forces!` for a Molly `System` (units, minimum-image, GPU-backed systems). `AllegroPotential` wraps the model.
- `test/allegro_reference.py` — offline e3nn generator: convention references + a small (random, seeded) model with reference energies/forces.

## Validation (all passing)

- `test/equivariant.jl` — **480 checks**: SH normalization/equivariance, finite-difference gradient checks of every primitive, exact tensor-product invariance/equivariance under random SO(3), the tensor-product/linear VJPs, and the e3nn convention pins (SH + Wigner-3j).
- `test/allegro_potentials.jl` — **17 checks**: energy reproduces the e3nn reference (rtol 1e-4) and is rotation-invariant through the calculator; analytic forces match the reference (~1e-9), `ΣF ≈ 0` (~1e-16), and match finite differences of both the core energy and the calculator. A guarded GPU-consistency testset runs for CUDA in CI.
- Validated on **Apple Metal**: energy CPU==GPU, forces match to ~5e-7 (Float32).

## Milestones in this PR

1. Equivariant primitives (with equivariance + finite-diff gradient tests).
2. Pinned to e3nn conventions (SH axis order/normalization, Wigner-3j signs).
3. CPU energy (model + `potential_energy`), validated vs the e3nn reference.
4. CPU analytic forces (`forces!`), validated vs reference + finite-diff, `ΣF=0`.
5. GPU-backed systems run correctly (host round-trip; Metal-validated, CUDA-ready).
6. Docs example + scaling benchmark.

## Known follow-ups

- **Native on-device GPU kernels** (the AEV-style write-reduced tensor product): today GPU systems take a host round-trip for correctness/portability; on-device kernels are the performance step.
- **Neighbour-list path**: the CPU reference is O(N²) all-pairs within the cutoff; a CSR neighbour-list path (reusing the ANI machinery) is the next optimization.
- **Weights artifact + trained model**: `allegro_data_dir()` is stubbed pending a hosted trained Allegro checkpoint; loading from a local HDF5 works today. l ≤ 2 currently; l = 3 is a straightforward extension.


docs/src/api.md
```diff
@@ -9,6 +9,7 @@ Package extensions are used in order to reduce the number of dependencies:
 - To use [`ASECalculator`](@ref), call `using PythonCall`.
 - To use [`rdf`](@ref), call `using KernelDensity`.
 - To use [`ANIPotential`](@ref), call `using Lux, HDF5`.
+- [`AllegroPotential`](@ref), a native equivariant (Allegro-style) neural network potential, computes energy and analytic forces on the CPU (`using Lux, HDF5`). It is built on O(3)-equivariant primitives (real spherical harmonics, Clebsch-Gordan tensor products, equivariant linear layers) that live in core Molly and are pinned to e3nn's conventions; these and the forces are validated by `test/equivariant.jl` and `test/allegro_potentials.jl`. GPU kernels and a weights artifact are follow-ups.
```
jgreener64: No need for the long text.

ext/MollyAllegroExt.jl
```diff
@@ -0,0 +1,123 @@
+# Extension implementing the native Allegro equivariant potential's model loading and
```
jgreener64: Am I right in thinking the only weak dependency here is HDF5? In that case this should be MollyHDF5Ext.jl, but only the function that uses HDF5 needs to be in the extension.

Project.toml
```diff
@@ -53,6 +53,7 @@ PythonCall = "6099a3de-0909-46bc-b1f4-468b9a2dfc0d"
 MollyCUDAExt = "CUDA"
 MollyEnzymeExt = "Enzyme"
 MollyGLMakieExt = ["GLMakie", "Colors"]
+MollyAllegroExt = ["Lux", "HDF5"]
```
jgreener64: Lux not required?

ext/MollyAllegroExt.jl
```diff
@@ -0,0 +1,123 @@
...
+        return Molly.AllegroPotential(model, species_map, T(rc), Ref{Any}(nothing))
+    end
+end
+
+Molly.allegro_data_dir() = error("no artifact configured yet; load AllegroPotential from a local HDF5 path")
+
+# Coordinates: Molly stores unitless coords as nm; the model uses Å (×10). Unitful coords are
+# stripped to nm first. GPU-backed coords are pulled to the host (the model forward/backward is a
+# CPU reference; native on-device kernels are a further optimization).
+function _coords_to_angstrom(coords_dev)
```
jgreener64: Can this conversion machinery be shared with ANI?

src/equivariant/allegro_model.jl
```diff
@@ -0,0 +1,271 @@
...
+forward is [`allegro_edge_energy`](@ref) / [`allegro_total_energy`](@ref).
+"""
+struct AllegroModel{T}
+    C::Int; H::Int; nb::Int; S::Int; L::Int; env_p::Int
+    r_c::T
+    feat::Irreps
+    sh::Irreps
+    paths::TensorProductPaths
+    cg::SparseCG{T}
+    emb_W1::Matrix{T}; emb_b1::Vector{T}
```
jgreener64: Separate lines, and elsewhere.

src/equivariant/allegro_model.jl
```diff
@@ -0,0 +1,271 @@
...
+    return AllegroModel{T}(C, H, nb, S, L, env_p, T(r_c), feat, sh, paths, cg,
+                           T.(weights.emb_W1), T.(weights.emb_b1), T.(weights.emb_W2), T.(weights.emb_b2),
+                           T.(weights.init_w), T.(weights.init_b0),
+                           layers, T.(weights.out_W), T.(weights.out_b))
+end
+
+@inline _dense(W, b, x) = W * x .+ b
+
+# derivative of SiLU: d/dx[x·σ(x)] = σ(x)·(1 + x·(1−σ(x)))
+@inline function _silu_grad(x::T) where T
```
jgreener64: No leading underscore for function names.

test/allegro_potentials.jl
```diff
@@ -0,0 +1,116 @@
+# Tests for the native Allegro potential (energy + analytic forces). Guarded on the reference files
```
jgreener64: This can go in the same file as the ANI tests.

test/allegro_reference.py
```diff
@@ -0,0 +1,252 @@
+#!/usr/bin/env python3
```
jgreener64, line 1: This seems to be a custom implementation in Python? Is there a library you can call into for this? Otherwise we don't know that this is correct.

jgreener64 (commented):
Looks like a good start. As you say, the next steps are to reuse the ANI neighbour machinery and to write the GPU kernels.

It would also be good to run a trajectory on 6mrr with a pre-trained model and check it looks okay.

docs/src/api.md
sivasathyaseeelan: Trimmed to a single line matching the other extension bullets.

ext/MollyAllegroExt.jl
sivasathyaseeelan: Done. Renamed to MollyHDF5Ext (gated on HDF5 only). Only the HDF5 weight loader stays in the extension; the AtomsCalculators energy/forces wiring moved to core Molly.

Project.toml
sivasathyaseeelan: Yes, Lux is not needed. The extension is now MollyHDF5Ext = "HDF5".

ext/MollyAllegroExt.jl
sivasathyaseeelan: Yes. Moved coords_to_angstrom, strip_boundary and the eV-to-units conversion (ml_energy_to_units / ml_force_to_units) into core Molly. Both ANIPotential and AllegroPotential use them now, and MollyLuxExt imports them instead of keeping its own copies.

src/equivariant/allegro_model.jl
sivasathyaseeelan: Done, split the multi-statement lines onto separate lines across the equivariant files.

src/equivariant/allegro_model.jl
sivasathyaseeelan: Done, removed the leading underscores from the internal function names.

test/allegro_potentials.jl
sivasathyaseeelan: Merged into test/ml_potentials.jl and removed the separate file.

test/allegro_reference.py
sivasathyaseeelan: The equivariant operations do come from a library. The spherical harmonics use e3nn's o3.spherical_harmonics and the tensor-product coefficients use o3.wigner_3j, and test/equivariant.jl pins the Julia primitives to these bit-for-bit. Only the surrounding dense MLPs are plain numpy, and I have documented this in the script header. Matching a trained mir-group/allegro checkpoint end to end (loading its weights and running a 6mrr trajectory, as you suggested) is the planned next step.

sivasathyaseeelan:
Thanks for the review. Addressed all the comments and rebased on the latest main:

- The extension is now HDF5-only (`MollyHDF5Ext`), holding just the weight loader; the energy/forces `AtomsCalculators` wiring moved to core Molly (no Lux dependency).
- The unit/coordinate conversion (`coords_to_angstrom`, `strip_boundary`, `ml_energy_to_units`, `ml_force_to_units`) is now shared with ANIPotential in core.
- Style fixes (separate lines, no leading underscores), Allegro tests merged into `test/ml_potentials.jl`, docs trimmed.
- On the reference: the equivariant operations come from e3nn (`o3.spherical_harmonics`, `o3.wigner_3j`) and the Julia primitives are pinned to them bit-for-bit in `test/equivariant.jl`; validating against a trained mir-group/allegro checkpoint end to end is the planned next step.

jgreener64:
Thanks for the changes.

It seems that the implemented model is a pair potential, not Allegro. `allegro_edge_energy` sees one edge, and every tensor product is taken with that edge's own spherical harmonics. The edge energy is therefore a rotation-invariant function of a single vector, i.e. a function of `|r_ij|` and the species only.

Fixing this and comparing to the reference values seems like the next step.

sivasathyaseeelan:
@jgreener64 thanks for the review, you were right that the previous version was a pair potential. Every layer's tensor product used the edge's own spherical harmonics, so each edge energy was a function of only (|r_ij|, species). I have fixed this and pushed the changes.

### What changed

1. **The model is now genuinely many-body.** Each layer couples an edge to the central atom's environment, a sum over all of that atom's neighbours, following the Allegro paper (Musaelian et al. 2023):
   `Env_i = (1/avg_nn) * Σ_{k in N(i)} envMLP(x_ik)[l,c] * Y(r̂_ik)`
   and the tensor product is now `V_ij ⊗ Env_i` (feat ⊗ feat, uvu), not `V_ij ⊗ Y_ij`. The density trick keeps it to one environment sum per atom, then one tensor product per edge.

2. **Analytic forces.** The finite difference placeholder is replaced by a full hand written reverse pass (taped forward, backward through every layer and the environment pooling, environment adjoint scattered back to all neighbours, then the spherical harmonic Jacobian and the |r| chain). This mirrors the native ANI force path.

3. **Reference and tests.** `test/allegro_reference.py` now generates a many-body reference, committed to `data/allegro_reference/` so the Allegro tests run in CI (they were being skipped before). Added a non-additivity test that proves the model is not a pair potential (`E_full ≠ Σ isolated-pair energies`).

4. **Rebased on latest master** (the new ParallelTestRunner test layout).

5. **Benchmark suite** in the same style as the ANI benchmarks (`benchmark/allegro*.jl` + `benchmark/allegro_benchmarks.md`, with figures).

### Validation

- Energy matches the many-body reference to about 1e-16.
- Analytic forces match central finite differences to about 3e-10 and the reference forces to about 4e-10, with ΣF about 2e-16.
- Rotation, translation and permutation invariant energy, and the non-additivity test confirms it is genuinely many-body.
- Equivariant primitives 480/480.
- Analytic forces are 29x to 109x faster than finite differences over 16 to 64 atoms, the gap widening with system size.

One thing still open: the reference is a self consistent numpy implementation of the same many-body architecture rather than a bit for bit match to the real nequip-allegro package. I can take that on next if you would like it for this PR.

Would appreciate another look when you have time. Thanks!

jgreener64:
Looks good. It's important that the reference is against something external though, ideally just calling the existing package from Python. Re-implementing everything in Python to test against raises the question of how we know that implementation is correct.

> It would also be good to run a trajectory on 6mrr with a pre-trained model and check it looks okay.

This would also be useful (you can send it to me on Slack).

sivasathyaseeelan:
@jgreener64 thanks for your previous comment.

**External reference:** now generated by calling the real `nequip-allegro` package (`test/allegro_package_reference.py`), replacing the numpy port. Molly drives the same package model through `ASECalculator` and reproduces its energy and forces to about 1e-6 (`test/allegro_ase_validation.jl`).

**6mrr trajectory:** added native GPU forces (`compute_allegro_forces_ka`, CUDA and Metal), verified against the CPU backward (1e-15) and finite differences under PBC (1e-9). A 250 fs NVE run on the full system (15,954 atoms) stays finite. Energy still drifts because the reference weights are untrained, not a trained potential. A small model I trained on SPICE is not MD-stable yet, so a physical trajectory needs more training. I will send the DCD and the drift plot on Slack.

PR #285, opened 2026-08-19
Add MetaDynamics

ywitzky:
Add enhanced sampling via MetaDynamics simulations implemented as a general force field.

MetaDynamics computes collective variables at fixed intervals and adds a Gaussian potential to the memory of the visited cvs with a frequency according to _deposit_interval_. The memory is implemented as subtype of _AbstractMetaDynamicsMemory_:
- _ListHills_: Lists of cv-positions, height and variance. Exact method that needs summing over all previous hills in vicinity.
-  _GridHills_: Discretized cv-space spanning grid_min:bin_width:grid_max. Not exact but O(1) evaluation cost
- Can be extended by none equidistant grids or adaptively refined grids for less space requirements for higher dimensions

The bias force is evaluated at each step from the time dependent memory bias according for each cv. Tempering is implemented as a subtype of _AbstractTempering_. 
- _NoTempering_: Heights of gaussians is constant with time. This can easily lead to biases that are far larger than the free energy potential differences.
- _WellTemperedTempering_:  Heights of the gaussians decay with fixed rate to limit the maximal bias.
- interface for _tempering_height_ allows to implement cv-, bias-, memory- and height-dependent schemes if wanted.

I intend to add more CVs that are commonly used in [plumed](https://www.plumed.org/doc-v2.10/user-doc/html/_colvar.html) to explore the conformational ensemble of a protein.


jgreener64:
Looks interesting. If this is ready for review could you write a bit about what the code does and the ideas behind the design?

ywitzky:
@jgreener64 
I have added the description. Let me know if you want more or different information.

Most of the interface are my best guess of what might be interesting for more elaborate bias schemes. We could change them.

Fully offloading for the gpu is planned for future improvements but is probably not yet supported by the code.

jgreener64:
Looks promising. Am I right in thinking that every call to `potential_energy` and `forces` adds to the bias? In the past I have thought about a `visited` keyword argument to these functions (and friends like `total_energy`, `forces_virial`). This flag is default `true` with the behaviour in this PR, i.e. changing the bias. But it is set to `false` in contexts where you only want to observe the system rather than changing it, like logging. Maybe you could consider designs like this and make the relevant changes.

It would also be useful if @allexandre97 could review once he is back from holiday.

ywitzky:
Not exactly. The bias is increased only every _deposit_interval_'th function call to allow the system to relax before increasing the bias to rapidly. 

We could use a keyword like you proposed. Currently __deposit_interval__ relies on an internal counter which is increased whenever it is called, such that it is not exactly the time step. I like that Idea.

PR #287, opened 2026-09-08
GPU-native collective variables and biased-simulation overhead reduction

ywitzky:
## Motivation

`BiasPotential`-driven simulations (metadynamics-style biased MD, using `CalcDist`/`CalcRg`/
`CalcRMSD`/etc. as collective variables) were, at the start of this work, both **slow relative to
CPU at realistic system sizes** and **expensive relative to an unbiased simulation** on GPU:
attaching 3-10 `BiasPotential`s to a `Langevin` simulation cost 2.6x-5.2x the wall time of the same
simulation with no bias. Two separate problems caused this, and this branch fixes both:

1. **CV computation was not offloaded to GPU at all.** It ran on CPU every step regardless of the
   rest of the simulation being GPU-resident, forcing a host/device round trip per CV per step --
   a cost that scaled with the CV's atom-group size instead of staying flat.
2. **Per-step host/launch overhead dominated wall time**, independent of any single kernel's cost.
   Profiling showed most of a biased step's cost was plain host-side Julia dispatch and CUDA API overhead, not device compute. Therefore, and CUDA graph
   capture (`use_cuda_graph=true`) was implemented to capture the serial call of force kernels and replay them to reduce launch overhead.

### 1. GPU-resident CV evaluation, minimizing launch count (`src/bias/cv.jl`)

`CalcMinDist`/`CalcMaxDist`/`CalcCMDist`/`CalcRg`/`CalcRMSD` gained genuine GPU implementations
(`calculate_cv!`/`cv_gradient!`), designed from the start around keeping the number of kernel
launches per call fixed and small rather than scaling with the CV's atom-group size -- the design
goal throughout was minimizing launch count, since launch overhead, not device compute, is what
dominates at the group sizes these CVs are actually used at.

Each CV type gets a persistent scratch struct (`MinMaxScratch`/`CMDistScratch`/`RgScratch`/
`RmsdScratch`, all in `cv.jl`) sized once, using one-off host-side type-sample probing to get the
right Unitful element types (mass-weighted position, mass-weighted squared-deviation, etc.)
without guessing at unit formulas. `BiasPotential` (`bias.jl`) allocates these lazily
(`ensure_bias_dist_scratch!`) and every downstream index array is uploaded to the device exactly
once (`upload_idx`) instead of being re-uploaded from a `@view coords[cv.atom_inds]` on every call.

Net effect: `calculate_cv!`/`cv_gradient!` for every CV type is now bounded-launch-count,
GPU-resident work instead of forcing a fallback to CPU -- GPU now outperforms CPU at realistic
group sizes.

### 2. Keeping per-CV launch counts to the minimum the dependency structure allows
   (`src/bias/bias.jl`, `cv.jl`)

Each CV's value+gradient path is a small, fixed number of launches, kept to the minimum a genuine
sequential dependency forces -- steps that don't actually depend on each other are combined into
one kernel rather than issued separately:
- `bias_apply_force_kernel!` (superseded by the batched kernel below) combines `fs_svec =
  d_bias_buf .* grad; fs -= fs_svec` into 1 launch instead of 2.
- CMDist's value+gradient path: 2-3 launches.
- RMSD's value+gradient path: 3-4 launches (`gather_kernel!` + `isum_reduce` + `finalize` [+
  `grad_write`]), reusing the host-computed Kabsch rotation rather than recomputing it.
- Rg needs 4 (value) / 5 (gradient) launches -- center-of-mass must be fully resolved before the
  sum-of-squared-deviations reduction can start, a genuine multi-stage dependency.

### 3. Deferred, batched NaN/Inf checking (`src/bias/bias.jl`)

`BiasPotential`'s finite-value guard used to sync to host every step (illegal mid-capture, too).
Now a kernel records the first bad step into a device buffer with no sync; a periodic host readback
(every `finite_check_every` steps, default 20) checks it and errors with the recorded step if
needed, batched across all attached biases into one sync instead of one per bias. Gated by
`defer_finite_check`, on only for `use_cuda_graph=true` step loops -- everywhere else keeps the old
immediate check.

### 4. Batched multi-bias captured-path tail (`src/bias/bias.jl`, `src/force.jl`)

For `n_bias` simultaneous `BiasPotential`s, several per-bias steps that have no bias-type- or
CV-type-specific dispatch were fused into one launch covering all biases at once, using a
compile-time tuple-recursion pattern (not a runtime loop, to stay type-stable over the
heterogeneous per-bias buffer/bias-type tuples):
- `bias_batched_gradient_kernel!`: every bias's `bias_gradient` (LinearBias/SquareBias/
  FlatBottomSquareBias/PeriodicFlatBottomBias dispatch) in one launch.
- `bias_batched_apply_kernel!`: every bias's `fs_svec = d_bias_buf .* grad; fs -= fs_svec` in one
  launch.
- `bias_batched_finite_kernel!`: every bias's finite check on the one array the tail itself
  produces, in one launch.

`force.jl`'s `general_inters` dispatch loop gained a compile-time-stable partition
(`split_biases`) that pulls `BiasPotential`s out for this batched treatment while leaving any other
`general_inters` type on its original, unbatched, per-inter call, unchanged (no such type exists
in this codebase today, but the loop no longer assumes `general_inters` is bias-only). Each bias's
own `cv_gradient!` remains a separate, per-bias call -- different CV types need genuinely different
kernels, so that part doesn't batch.

### 5. CuGraph capture mechanism (`ext/MollyCUDAExt.jl`, `src/force.jl`, `src/simulators.jl`)

CUDA.jl's own `@captured` re-records the graph and calls `cuGraphExecUpdate` on *every* call, never
degrading to a pure replay -- ~4x more expensive than a bare `cuGraphLaunch` at realistic bias
counts. `captured_forces_once!` (`MollyCUDAExt.jl`) captures once and replays via plain
`cuGraphLaunch` for the rest of the `simulate!` call. This is safe because:
- Topology-changing steps (virial, tile refresh) already route around capture via
  `needs_vir`/`is_tile_refresh_step`; a tile refresh also invalidates the cached graph, forcing
  recapture next eligible step.
- `warmup_cuda_graph_capture!` (`force.jl`) runs two warm-up `forces!` calls before the step loop
  starts, materializing every `BiasPotential` buffer and JIT-compiling every captured-path kernel
  (both illegal mid-capture) -- unconditionally, regardless of whether the first real step needs a
  virial.

Two graphs are cached per `System` (`graph_exec_no_check`/`graph_exec_with_check`, with/without the
finite-check kernel), since the finite-check kernel's `step_n` argument would otherwise freeze at
capture time -- the with-check graph uses a fixed sentinel step, so its error message reports
"somewhere since the last finite check" rather than an exact step.

Per-step routing (plain vs. captured, deferred finite-check) is centralized in `forces_step!`
(`force.jl`), dispatched on `Val(use_cuda_graph)` so `use_cuda_graph=false` compiles to a bare
`forces!` call with no added overhead. 
- **`use_cuda_graph` is support for simulators:**
`Langevin`, `VelocityVerlet`, `DPDVelocityVerlet`, `StormerVerlet`, `NoseHoover` and `OverdampedLangevin`
 via this same machinery, leaving any conditional post-coupling force recompute
(box-rescale-triggered, same topology risk as a tile refresh) uncaptured;
-  **not yet supported for**: `MTSIntegrator`/`MTSLangevinIntegrator`/`ReplicaExchangeMD`
- **(probably) not reasonable:** `SteepestDescentMinimizer`, `MetropolisMonteCarlo`.

## Net measured impact

Benchmark with Langevin, `use_cuda_graph=true`,`n_atoms=800`, 200 steps, ~10 atoms per CV:
overhead vs an unbiased simulation of the same system: 

| n_bias | uncaptured | captured |
|---|---|---|
| 3  | ~2.6x (160% extra) | **1.06x (6% extra)** |
| 5  | ~3.3x (230% extra) | **1.28x (28% extra)** |
| 10 | ~5.2x (420% extra) | **1.95x (95% extra)** |

The remaining gap at higher `n_bias` is attributable to each bias's own `cv_gradient!` launches (CV-type-specific,
not batchable across biases) and the fixed per-`cuGraphLaunch` floor, not to anything left
unaddressed by this branch's fixes. Benchmark depends on the CVs and their number of sucessive kernel launches.

CVs scale reasonable with number of atoms and tend not to be the limiting issue for n_atoms>2000. 

## Testing

- `test/basic.jl`/`test/bias.jl`/`test/gpu_optimizations.jl`, extended, GPU-backed, all passing:
  - 68 "Collective variables", 36 "Bias potentials", 3 "Bias correction and GPU"
  - 16 "Biased simulation", 21 "BiasPotential persistent buffers"
  - new "GPU Optimizations" testset: `use_cuda_graph` (re-)capture-after-tile-refresh behavior
    across the newly-supported simulators

  
## Documentation
- `docs/src/documentation.md` gained a "GPU offload for biased simulations" section covering when
  to offload CV/bias calculation to GPU, `use_cuda_graph` legality requirements, and warm-up/
  invalidation behavior, aimed at users rather than reviewers of this diff.
- Ad hoc correctness scripts used during development (not upstreamed into `test/`): CPU-vs-GPU
  numerical agreement for MinDist/CMDist/Rg/RMSD at scale; batched-vs-per-bias force agreement (max
  abs difference at floating-point noise level, ~5.7e-14).

## Known gaps / not addressed by this branch

- **CalcRMSD is not compatible with `use_cuda_graph=true`** (`check_cuda_graph_legality`,
  `simulators.jl`): its Kabsch alignment needs a host SVD every call, which can't be captured. Same
  for any `BiasPotential` using `correction=:pbc`: `unwrap_molecules` is GPU-native (no host sync,
  see `_gpu_unwrap_fractional`, `spatial.jl`) but still allocates via broadcast every call, which
  is equally illegal inside a captured region. CalcMinDist/CalcMaxDist ARE supported -- their
  persistent-scratch path finds the extremal pair fully device-resident, no host sync. These are
  pre-existing constraints, not regressions, but are now more visible since capture is faster,
  cheaper, and available on more simulators.
- **`calculate_virial_dist!`/`calculate_virial!` for CV types still use the older `@view
  coords[cv.atom_inds]` re-upload pattern and bare `sum()` host syncs** -- lower priority since
  virial steps are periodic and already excluded from the captured region, but not yet brought up
  to the same standard as the value/gradient paths.
- **No batching across same-CV-type biases' own `cv_gradient!` calls** (e.g. N `CalcCMDist`
  biases each still launch their own `cmdist_reduce_kernel!`/etc. separately) -- scoped out as a
  large, CV-type-by-CV-type undertaking with a smaller expected payoff than the changes in this
  branch, given `cv_gradient!`'s launches are a minority of per-step cost after items 3-5 above.
- **CPU/GPU-mixed bias execution** (some `BiasPotential`s on CPU, others on GPU, in the same
  `System`) is not supported -- tracked as a separate flexibility feature, not a performance gap.

## Beyond this branch

- **CUDA graph capture, more broadly.** The current mechanism captures only the per-step force
  evaluation. NVIDIA's own guidance (see
  [this GROMACS write-up](https://developer.nvidia.com/blog/a-guide-to-cuda-graphs-in-gromacs-2023/))
  and our own numbers agree that uncaptured force evaluation typically costs 3-20x more than a
  captured replay at 3-10 `BiasPotential`s -- so there's a case for capturing whole `simulate!`
  step loops, not just the force call, which would need the hot loop refactored into a single
  function with explicit handling for every invalidating/untrackable event (tile refresh,
  coupling-triggered recompute, periodic finite-checks). A secondary optimization once that exists:
  tuning how often those non-capturable steps are allowed to occur, to reduce both the number of
  distinct graphs recorded and the number of uncaptured-path launches. AMD's equivalent is [rocGRAPH](https://rocm.docs.amd.com/projects/rocGRAPH/en/latest/), which I couldnt test.
- **Persistent scratch for `unwrap_molecules`.** Its GPU path (`_gpu_unwrap_fractional`,
  `spatial.jl`) is already fully device-resident (parallel pointer-doubling, no host sync) but
  still allocates a new array on every broadcasted intermediate each call, which is why
  `correction=:pbc` biases can't use `use_cuda_graph=true` today. Following the same
  pre-sized-persistent-scratch pattern already used for `MinMaxScratch`/`CMDistScratch`/
  `RgScratch`/`RmsdScratch` (`cv.jl`) would remove that allocation and let `:pbc`-correction
  biases join the captured path too -- scoped out of this branch since it's a similar-shaped but
  separate piece of work per CV/correction type.
- **Test-output suppression.** `test/runtests.jl` now wraps `using AMDGPU`/`CUDA`/`Metal`/`oneAPI`
  in `Suppressor.@suppress` (this machine has neither ROCm, Metal, nor oneAPI hardware, so those
  packages' `__init__` discovery routines print substantial error/warning noise otherwise). Easy to
  revert if a reviewer prefers to see that output; kept because it makes local `Pkg.test` runs
  much easier to scan for actual failures.


ywitzky:
The documentation check fails, as I propose docstrings for functions that arent exported. This makes sense as no User should worry about that functionality. I was wondering if the _dev_ documentation should or could contain more information for developers to make it easier to contribute. 

I can convert the docstrings to simple comments if the consensus is that these docstrings are unnecessary.

A few tests in _/tests/gpu_optimisations.jl_ are currently uncommented by accident as they passed on in my test and slowed down the testing. They will be added once I know how bad the test coverage is. The codecov report is probably worse than it actually is, as the gpu code doesnt seem to be executed.

jgreener64:
Looks interesting and there is definitely useful stuff here. As it is, the change is too big to easily reason about which part is worthwhile and speeds things up.

I would suggest making this PR just the GPU-native molecule unwrapping and buffered CV API/removal of the host round trips. Then a follow-up PR could consider the fused per-CV kernels and CUDA graph capture, if they show value during benchmarking. The latter two are a higher maintenance burden, so we would want to make sure they are worth it.

You can also use the buffers objects in src/force.jl rather than putting mutable per-system state in `BiasPotential`.

> I can convert the docstrings to simple comments if the consensus is that these docstrings are unnecessary.

Yes this is what we do for non-exported things.

> The codecov report is probably worse than it actually is, as the gpu code doesnt seem to be executed.

It is not run on CI, but do make sure to run the tests with a GPU locally to check everything passes.

ywitzky:
Ok. I will split it. Some of the GPU kernels perform worse than the CPU kernels for reasonable group sizes due to the limit compute needed for them and the overhead of launching 3-4 kernels on the GPU without CuGraph.

CuGraph increases the maintenance but it should improve performance for small systems.

PR #288, opened 2026-09-18
Alchemical setup + lambda-scaled functions + HREMD on GPU

ERovers:
# Overview

Code that will support free energy calculations including absolute solvation free energy, absolute binding free energy and relative binding free energy. While softcore potentials and annihilation lambda schedulers were present, this pull requests includes an unifying way of setting up absolute FE systems and relative FE systems, along with lambda scaled functions for all the interactions possibly found in the current forcefields supported. The systems set-ups are based on a reference system and lambda scheduler. The OpenFELambdaScheduler defaults will result in similar energies and forces to OpenFE implementation (included in the tests). **Molly (with this PR) will have more capabilities than OpenFE with the possibility to test different softcores, variety of functional forms for Coulomb and LJ, option to add new lambda schedulers, and an option to switch between single and dual topology.**

Also, REMD is updated to be able to run on multiple GPUs and thereby speeding up the simulation time. Similar implementations could potentially be possible for TSS and AWH.

Since there is a lot of commit history, a squash merge might be better to preserve memory.

# Implementation details
## Alchemical set up and lambda-scaled functions
- **Alchemical systems** (`src/free_energy/alchemical_system.jl`)
  - `AbsoluteFESystem(sys, global_λ, mapping)` for absolute free energies, `RelativeFESystem(sysA,
    sysB, global_λ, mapping, core_mapAB)` for relative ones.
  - Atoms carry a role: environment, core, core-insert, core-delete, insert or delete. Roles are
    mixed per interaction, which decides how each term is coupled.
  - Dual topology (energy scaling) and single topology (parameter scaling, OpenFE-style).
  - Rebuilds PME, the Ewald exclusions and the dispersion correction for the λ-dependent charges,
    so the three Ewald terms and the tail always agree with each other.
- **Lambda schedulers** (`alchemy.jl` , `scale_dual.jl` / `scale_single.jl`)
  - `DefaultLambdaScheduler`, `LinearLambdaScheduler`, `GROMACSLambdaABFEScheduler`,
    `GROMACSLambdaRBFEScheduler`, `OpenFEScheduler`, `NAMDLambdaScheduler`,
    `QuartersLambdaScheduler`, `EleScaledLambdaScheduler`, all exported and documented.
  - Each defines how sterics and electrostatics are staged along λ; shared options cover dual or
    single topology, per-atom or per-state parameter mixing, 1-4 handling, and whether the
    alchemical atoms keep their own Lennard-Jones (`intraLJ`).
- **Soft cores**: Beutler and Gapsys for Lennard-Jones and Coulomb, plus directly scaled Coulomb,
  each in the plain, reaction-field and Ewald forms.
- **λ-aware interactions**: every bonded term (bonds, angles, Urey-Bradley, torsions including CMAP,
  cosine angle, Ryckaert-Bellemans, Morse) and every non-bonded term gained a λ counterpart, built
  automatically when an alchemical system is created.
- **Two-grid PME** (`PME_λ`): the GROMACS convention, where the mesh mixes two charge grids with the
  electrostatic coupling, next to the OpenFE/OpenMM convention of scaling each atom's charge.
- **λ-scaled virial**: Determining whether interaction has or has not been scaled by lambda yet, if not it will be scaled by lambda and otherwise will be left as is.
- **λ-dependent dispersion correction** (`LJDispersionCorrectionλ`) with the matching virial scaling
  (`virial_scaling.jl`), so pressure coupling stays correct across λ.
- **Documentation**: Added to the free energy documentation the different free energy protocols, the different system setups and how to match OpenFE settings. 
- **Tests**: A `RelativeFESystem` is built for TYK2 with ejm31 and ejm52 and the energies and forces are compared to OpenFE (OpenMM) over a range of lambda values (`test/free_energy.jl`). **NOTE**: OpenFE's LJ softcore had to be adjusted to truly match the softcore describe in [Gapsys et al. 2012](https://pubs.acs.org/jctcce/article/8/7/2373/1208704/New-Soft-Core-Potential-Function-for-Molecular). PR will be opened on the OpenFE github, for now results are included in the data files.
- **Test case**: An `AbsoluteFESystem` was built for benzene in solvent and in vacuum and ran with AWH, TSS and HREMD to calculate the hydration/solvation free energy. The values are compared to experimental value of -3.67±0.05 kJ/mol and predictions match closely with -3.763 kJ/mol for AWH, -3.901±0.176 kJ/mol for TSS, and -4.186 kJ/mol for HREMD. The set-up and protocols are described in the documentation `free_energy.md` and included as runnable scripts in `scripts/`. 

## Replica exchange on multiple GPUs

Master's `simulate_remd!` runs the replicas on CPU threads. This branch adds a GPU path and makes
both backends behave the same.

- **One worker process per GPU**, each advancing a block of thermodynamic states on its own device,
  so a ladder of 20 states runs on as many GPUs as are available. The benzene Hamiltonian replica
  exchange example takes 56 min on two GPUs, against 163 min for AWH and 191 min for TSS on one.
- **The state-keyed contract**, shared by CPU and GPU: state `k` integrates the replica currently in
  it, with that state's interactions, integrator, neighbour finder and loggers, and writes the
  coordinates, velocities and box back to that replica. Loggers therefore follow a thermodynamic
  state rather than a wandering configuration.
- **Systems are sent to the workers once**, not on every call. Later calls transfer only
  coordinates, velocities and boxes, which is what makes frequent sampling affordable.
- **Repeated calls resume** from `current_step`, so a run can be split into chunks to collect
  samples without restarting the step count or re-logging.
- **The exchange logger** records the accepted exchanges and the state-to-replica map after every
  cycle, which gives per-pair acceptance and the replica exchange history plot.
- **Documentation**: Hamiltonian replica exchange with MBAR is a documented workflow: `ReplicaSystem` built
  from one `ThermoState` per λ, sampled configurations scored in every state and solved with MBAR. A test case is created calculating the absolute solvation free energy of benzene. The test case is implemented for HREMD, AWH and TSS and matched to experimental values.
  
## Other small changes or additions.
- Improper torsion mapping was adjusted to previous master code to better match OpenMM assigning indexes. Not sure if new master code had fixed this bug already.
- `TrajectoryWriter` gained `suppress_warn` to suppress the warnings that it is appending to a file, which is helpful during REMD simulations if the trajectorywriter needs to be reinitialized on the GPU processes.
 - Added Amber14 forcefield files and tests in comparison to OpenMM on TYK2-ejm31 system (`test/protein.jl`).
 - Added a calcAngle in the CV.jl.


Project.toml
```diff
@@ -1,19 +1,21 @@
...
 authors = ["Joe G Greener <jgreener@hotmail.co.uk>"]
+version = "0.23.3"
 
 [deps]
 AcceleratedKernels = "6a4ca0a5-0e36-4168-a932-d9be78d558f1"
 Atomix = "a9b6321e-bd34-4604-b9c9-b65b8de01458"
 AtomsBase = "a963bdd2-2df7-4f54-a1ee-49d51e6be12a"
 AtomsCalculators = "a3e0e189-c65a-42c1-833c-339540406eb1"
 BioStructures = "de9282ab-8554-53be-b2d6-f6c222edabfc"
+CUDA = "052768ef-5323-5732-b1bb-66c8b64840ba"
```
jgreener64: This can stay as a weakdep.

data/force_fields/benzene.xml
```diff
@@ -0,0 +1,137 @@
+<ForceField>
```
jgreener64, line 1: Which force field generated this? Could add a metadata XML entry at the top.

src/bias/cv.jl
```diff
@@ -219,6 +220,71 @@ function calculate_cv(cv::CalcDist, coords, atoms, boundary, args...; kwargs...)
     return dist_val
 end
 
+"""
```
jgreener64, line 223: Add a test to `"Collective variables"` in test/bias.jl and `"CV gradients"` in test/gradients_2.jl.

src/free_energy/alchemical_system.jl
```diff
@@ -0,0 +1,739 @@
...
+    the potential is scaled directly by λ.
+"""
+function AbsoluteFESystem(sys::System, global_λ, mapping; 
+                        temp = 298.0u"K", 
+                        units=true,
+                        scheduler=DefaultLambdaScheduler(dual=true),
+                        loggers=(),
+                        array_type=Array,
+                        float_type=Float32,
+                        LJsoftcore="gapsys",
```
jgreener64, line 71: Symbols here like `:gapsys` would be more consistent with arguments elsewhere, also applies to other functions.

src/free_energy/alchemical_system.jl
```diff
@@ -0,0 +1,739 @@
...
+- `temp=298.0u"K"`: the temperature used to generate random velocities.
+- `units=true`: whether to use Unitful quantities.
+- `scheduler=DefaultLambdaScheduler(dual=true)`: the lambda scheduler that turns `global_λ`
+    into the couplings of the steric, electrostatic and bonded interactions, for example
+    [`DefaultLambdaScheduler`](@ref), [`GROMACSLambdaABFEScheduler`](@ref) or
+    [`OpenFEScheduler`](@ref). Absolute systems require dual topology (`dual=true`).
+- `loggers=()`: the loggers that record properties of interest during a simulation.
+- `array_type=Array`: the array type for the simulation, for example use `CuArray` or
+    `ROCArray` for GPU support.
+- `float_type=Float32`: the float type of the system. If the type of `global_λ` differs, the
```
jgreener64, line 56: This be inferred from the system, i.e. `System{<:Any, <:Any, T}` and `where T`. Also for the relative struct.

src/free_energy/alchemical_system.jl
```diff
@@ -0,0 +1,739 @@
...
+    Coords       = []
+    Interactions = []
+    Boundary = deepcopy(sys.boundary)
+
+    # Add all the atoms with new numbering (save a mapping to adjust the interactions list later)
+    # Add first core, then unique and then environment atoms
+    counter = 1
+    res_n = 0
+    res_number = 0
+    chain = ""
```
jgreener64, line 109: Not used? Is the comment accurate?

src/free_energy/alchemical_system.jl
```diff
@@ -0,0 +1,739 @@
...
+    end
+
+    special = falses(n_atoms, n_atoms)
+    for (i, l) in zip(Interactions[3].is, Interactions[3].ls)
+        special[i, l] = true
+        special[l, i] = true
+    end
+
+    if AT <:AbstractGPUArray
+        nf = GPUNeighborFinder(eligible=to_device(eligible, AT), special=to_device(special, AT), n_steps_reorder=10,
```
jgreener64, line 651: This is only available for CUDA, see the use of `uses_gpu_neighbor_finder` in src/setup.jl for how to setup a `DistanceNeighborFinder` in non-CUDA cases.

src/free_energy/virial_scaling.jl
```diff
@@ -0,0 +1,18 @@
+# Factor on a specific interaction's virial contribution. The default of 1 is correct whenever
```
jgreener64, line 1: Can this small file go at the end of another?

src/analysis.jl
```diff
@@ -173,3 +174,36 @@ Returns a list of distance bin centers and a list of the corresponding values
 of ``g(r)``, which tends to 1 at large distances for a uniform system.
 """
 function rdf end
+
+
+"""
+    read_trajectory(sys::System, traj_file; units=true)
```
jgreener64, line 180: There is similar code to to this in src/trajectory.jl, could you use or extend that?

src/loggers.jl
```diff
@@ -794,7 +794,7 @@ end
 """
     TrajectoryWriter(n_steps, filepath; format="", correction=:pbc, atom_inds=[],
                      excluded_res=String[], write_velocities=false, write_boundary=true,
-                     overwrite=false)
+                     overwrite=false, suppress_warn=false)
```
jgreener64, line 797: `append_warn=true` might be clearer which warning is suppressed.

src/simulators.jl
```diff
@@ -108,7 +108,9 @@ function check_array_nans(svec_arrays, labels, step_n)
         err_msg = "NaNs found at the end of step $step_n:"
         for (svec_array, label) in zip(svec_arrays, labels)
             c = count(isnan_svec, svec_array)
+            idx = findall(x -> isnan_svec(x), svec_array)
             err_msg *= "\n    $label - $c out of $(length(svec_array)) contain a NaN"
+            err_msg *= "\n    indexes: $idx"
```
jgreener64, line 113: This risks an unclear output if there are thousands of NaNs. You could only print the first ten then say "and more".

src/simulators.jl
```diff
@@ -2300,6 +2424,121 @@ end
...
+        return rep_id_proc, n_blocks
+    end
+
+    if n_proc < k
+        @warn("Number of processes ($n_proc) less than the number of replicas ($k), some replicas will not be simulated in parallel, but sequentially")
+    elseif n_proc > k
+        @warn("Number of processes ($n_proc) greater than the number of replicas ($k), some processes will be idle during the simulation, 
+        consider reducing the number of processes to match the number of replicas for more efficient simulation")
+    end
+    println("Attach GPUs to workers")
```
jgreener64, line 2504: Use `@info` if printing is required.

src/simulators.jl
```diff
@@ -2300,6 +2424,121 @@ end
...
+    # For composite structs (like TrajectoryWriter), map recursively over all fields
+    new_vals = ntuple(i -> reinit_c_pointers(getfield(val, i)), fieldcount(T))
+    
+    # Re-invoke the positional constructor for the parent struct
+    return T(new_vals...)
+end
+
+# Writers are rebuilt on the workers at every call and keep appending to their own file, so
+#   the file is never deleted and the append warning is not repeated
+function process_logger(logger)
```
jgreener64, line 2469: This can be done with dispatch, i.e. a default `process_logger` that returns `logger` and then a specialised method. It also could live in src/loggers.jl, and be called something more descriptive.

src/units.jl
```diff
@@ -98,13 +98,21 @@ function check_other_units(atoms, boundary, sys_units::NamedTuple)
     end
 
     for at in from_device(atoms)
-        if hasproperty(at, :σ) && at.σ != 0.0u"nm" && unit(at.σ) != sys_units[:length]
-            throw(ArgumentError("Atom σ has $(unit(at.σ)) units but length unit of coords " *
-                                "was $(sys_units[:length])"))
+        if hasproperty(at, :σ)
+            for σ in at.σ
```
jgreener64, line 102: I don't think this multiple parameter pattern is used any more?

src/types.jl
```diff
@@ -463,6 +464,142 @@ function Base.show(io::IO, sil::T) where T <: SpecificInteractionList
...
+    end
+    return interactions_final
+end
+
+function Base.append!(il1::InteractionList1Atoms{I, T, D}, il2::InteractionList1Atoms{I, T, D}) where {I, T, D}
+    return InteractionList1Atoms(
+        append!(il1.is,il2.is),
+        append!(il1.inters,il2.inters),
+        append!(il1.types,il2.types),
+        nothing,
```
jgreener64, line 488: Keep the data from `il1`.

src/types.jl
```diff
@@ -1639,15 +1786,84 @@ function AtomsBase.atomic_number(sys::ReplicaSystem)
...
+
+    # 3. Construct a new System instance with the copied fields
+    new_sys = ccall(:jl_new_structv, Any, (Any, Ptr{Any}, UInt32), 
+                    typeof(sys), reference_array(field_copies), length(field_copies))
+    
+    dict[sys] = new_sys
+    return new_sys::typeof(sys)
+end
+
+function to_device(t::Tuple)
```
jgreener64, line 1859: Move these up to the other methods for these functions.

src/types.jl
```diff
@@ -463,6 +464,142 @@ function Base.show(io::IO, sil::T) where T <: SpecificInteractionList
...
+        append!(il1.js,il2.js),
+        append!(il1.ks,il2.ks),
+        append!(il1.ls,il2.ls),
+        append!(il1.inters,il2.inters),
+        append!(il1.types,il2.types),
+        nothing
+    )
+end
+
+function Base.append!(il1::InteractionList5Atoms{I, T, D}, il2::InteractionList5Atoms{I, T, D}) where {I, T, D}
```
jgreener64, line 525: Use dispatch rather than branching here; have a specialised version of this function in the CMAP file.

src/mixing.jl
```diff
@@ -118,14 +148,19 @@ function ϵ_mixing(me::MixingException, atom_i, atom_j, args...)
...
     default = λ_mixing(me.mixing, atom_i, atom_j, args...)
     return get_pair(me.exceptions, atom_i.atom_type, atom_j.atom_type, default)
 end
 
-
 struct MinimumMixing end
 
-function λ_mixing(m::MinimumMixing, a, b, args...)
-    return min(one(a.λ), min(a.λ, b.λ))
+@inline function λ_mixing(m::MinimumMixing, lambdas::Tuple{T, Vararg{T}}, args...) where T
```
jgreener64, line 158: Is it okay for performance to still pass the two atoms through to this function fully, rather than just the lambda values? This would allow users to define lambda mixing functions that depend on other atom parameters.

test/reference/openfe_comparison.py
```diff
@@ -0,0 +1,309 @@
...
+        )
+
+complex_dag = transformation_complex.create()
+solvent_dag = transformation_solvent.create()
+
+#### Step 6 ####
+complex_unit = list(complex_dag.protocol_units)[0]
+results = complex_unit.run(dry=True, verbose=True)
+import pickle
+with open(os.path.join(out_dir, "openfe_tyk2_system.pkl"), "wb") as file:
```
jgreener64, line 195: Link to artifact?

data/openmm_tyk2/openfe/A_db.json
```diff
@@ -0,0 +1,8 @@
+{
```
jgreener64, line 1: The data files are still quite large, 40 MB in total. Could you add openmm_tyk2 to the artifact? See test/ml_potentials.jl for how to use an artifact in the tests.

src/force_field.jl
```diff
@@ -241,19 +241,13 @@ end
 
 # Impropers: lookup with 6-permutation scan and cache
 function find_improper_match(t1::AbstractString, t2::AbstractString, t3::AbstractString,
-                             t4::AbstractString; resolver::TorsionResolver{T, E},
+                             t4::AbstractString, indexes, atom_type_of, resnum_of, template_id_of, 
+                             element_of; resolver::TorsionResolver{T, E},
                              type_to_class::Dict{String, String}) where {T, E}
```
jgreener64, line 246: `atom_types` is another argument here.

src/force_field.jl
```diff
@@ -280,21 +275,178 @@ function find_improper_match(t1::AbstractString, t2::AbstractString, t3::Abstrac
...
+                    # We need the two peripheral atoms in positions 2 and 3, and the remaining
+                    #   peripheral in 4
+                    a1 = src_atoms[bestperm[2]]
+                    a2 = src_atoms[bestperm[3]]
+                    a4 = src_atoms[bestperm[4]]
+
+                    # Elements and masses for tie-break
+                    e_a1 = Symbol(element_of[a1])
+                    e_a2 = Symbol(element_of[a2])
+                    m_a1 = force_field.atom_types[atom_type_of[a1]].mass
```
jgreener64, line 392: No `force_field.`, also on line 431.

src/setup.jl
```diff
@@ -337,25 +337,20 @@ function resolve_proper_torsion(ff::MolecularForceField, t1::AbstractString, t2:
...
 function resolve_improper_torsion(ff::MolecularForceField, t1::AbstractString, t2::AbstractString,
-                                  t3::AbstractString, t4::AbstractString)
+                                  t3::AbstractString, t4::AbstractString, indexes, atom_type_of, 
+                                  resnum_of, template_id_of, element_of)
     # Resolver scans all 6 permutations internally and caches the winner
-    p = find_improper_match(t1, t2, t3, t4; resolver=ff.torsion_resolver,
-                                                type_to_class=ff.type_to_class)
+    p = find_improper_match(t1, t2, t3, t4, indexes, atom_type_of, resnum_of, template_id_of, 
+                            element_of; resolver=ff.torsion_resolver, 
+                            type_to_class=ff.type_to_class)
```
jgreener64, line 345: `atom_types=ff.atom_types`.

src/setup.jl
```diff
@@ -704,6 +699,8 @@ function System(coord_file::AbstractString,
     atom_type_of = Vector{String}(undef, n_atoms)
     charge_of = Vector{Union{T, Missing}}(undef, n_atoms)
     element_of = Vector{String}(undef, n_atoms)
+    template_id_of = Vector{Integer}(undef, n_atoms)
```
jgreener64, line 702: `Vector{Int}` and below.

src/interactions/urey_bradley.jl
```diff
@@ -66,3 +66,77 @@ end
...
+    kangle::KA
+    θ0::A
+    kbond::KB
+    r0::D
+    λ_mixing::LM = MinimumMixing()
+    scheduler::SCH = DefaultLambdaScheduler()
+end
+
+function Base.zero(::UreyBradleyλ{KA, A, KB, D, LM, SCH}) where {KA, A, KB, D, LM, SCH}
+    return UreyBradleyλ(kangle=zero(KA), θ0=zero(A), kbond=zero(KB), r0=zero(D))
```
jgreener64, line 81: More consistent with existing code to pass through the mixing/scheduler properties to the zeroed case. Also for `+`, which can take the properties from `a1`. This is the case for a number of the interactions.

src/free_energy/alchemical_system.jl
```diff
@@ -0,0 +1,739 @@
...
+        c = sys_coords[i]
+        if i in mapping
+            push!(Atoms, Atom(index=counter, atom_type=a.atom_type, mass=a.mass, charge=a.charge, σ=a.σ, ϵ=a.ϵ, 
+                                λ=FT(global_λ), alch_role=DeleteRole))
+        else
+            push!(Atoms, Atom(index=counter, atom_type=a.atom_type, mass=a.mass, charge=a.charge, σ=a.σ, ϵ=a.ϵ, 
+                                λ=FT(1.0), alch_role=EnvRole))
+        end
+
+        push!(Data, AtomData(atom_type=d.atom_type, atom_name=d.atom_name, res_number=res_number,
```
jgreener64, line 123: `push!(Data, d)`?

src/free_energy/alchemical_system.jl
```diff
@@ -0,0 +1,739 @@
...
+    else
+        topology = nothing
+    end
+
+    sys_final = System(
+        atoms=to_device(Atoms, AT),
+        coords=to_device(Coords, AT),
+        atoms_data=Data,
+        boundary=Boundary,
+        topology=topology,
```
jgreener64, line 224: Pass the virtual sites.

src/free_energy/alchemical_system.jl
```diff
@@ -0,0 +1,739 @@
...
+        dA = sysA.atoms_data[i]
+        cA = sysA.coords[i]
+        aB = sysB.atoms[core_mapAB[i]]
+        dB = sysB.atoms_data[core_mapAB[i]]
+        cB = sysB.coords[core_mapAB[i]]
+        if scheduler.dual
+            push!(Atoms, Atom(index=counter, atom_type=aA.atom_type, mass=aA.mass, charge=aA.charge, σ=aA.σ, ϵ=aA.ϵ, 
+                                                λ=FT(global_λ), alch_role=CoreDRole))
+            if chain!=d.chain_id
+                chain = d.chain_id
```
jgreener64, line 348: `d` doesn't exist? I guess this path isn't tested as it would have errored.

src/free_energy/alchemical_system.jl
```diff
@@ -0,0 +1,739 @@
...
+    end
+
+    # Add unique atoms from system B
+    for i in mapping["unique_B"]
+        a = sysB.atoms[i]
+        d = sysB.atoms_data[i]
+        c = sysB.coords[i]
+        if scheduler.dual
+            push!(Atoms, Atom(index=counter, atom_type=a.atom_type, mass=a.mass, charge=a.charge, σ=a.σ, ϵ=a.ϵ, 
+                                                λ=T(global_λ), alch_role=InsertRole))
```
jgreener64, line 430: `FT`?

src/free_energy/scale_dual.jl
```diff
@@ -0,0 +1,291 @@
...
+        return λ, one(λ), one(λ)
+    elseif role == DeleteRole
+        λ = T(λ < (T(1.0) / T(3.0)) ? T(0.0) : (λ - (T(1.0) / T(3.0))) * (T(3.0) / T(2.0)))
+        return (1-λ), one(λ), one(λ)
+    elseif role == CoreIRole
+        return λ, one(λ), one(λ)
+    elseif role == CoreDRole
+        return (1-λ), one(λ), one(λ)
+    else
+        return one(λ), one(λ), λ
```
jgreener64, line 201: Differs from `scale_elec`, just checking that is intentional?

src/interactions/cmap_torsion.jl
```diff
@@ -344,3 +347,124 @@ end
...
+
+# λ version of `CMAPTorsion` for alchemical systems, built by `to_lambda_function`.
+@kwdef struct CMAPTorsionλ{I,L,LM,SCH}
+    index::I
+    size::I
+    λ_mixing::LM = MinimumMixing()
+    scheduler::SCH = DefaultLambdaScheduler()
+end
+
+Base.zero(::CMAPTorsionλ) = CMAPTorsionλ(index=0, size=0)
```
jgreener64, line 359: Pass through other properties.

src/interactions/cmap_torsion.jl
```diff
@@ -344,3 +347,124 @@ end
     pe = da*pe + ((data[idx,4]*db + data[idx,3])*db + data[idx,2])*db + data[idx,1]
     return pe
 end
+
+# λ version of `CMAPTorsion` for alchemical systems, built by `to_lambda_function`.
+@kwdef struct CMAPTorsionλ{I,L,LM,SCH}
```
jgreener64, line 352: No `L`? This can't have been constructed by the tests.

src/free_energy/alchemical_system.jl
```diff
@@ -0,0 +1,739 @@
...
+    Interactions = merge(Interactions)
+
+    # Pairwise Interactions
+    n_atoms = length(Coords)
+    eligible = trues(n_atoms, n_atoms)
+    for i in 1:n_atoms
+        eligible[i, i] = false
+    end
+
+    for (i, j) in zip(Interactions[1].is, Interactions[1].js)
```
jgreener64, line 628: Select by type not by position.

src/types.jl
```diff
@@ -463,6 +464,142 @@ function Base.show(io::IO, sil::T) where T <: SpecificInteractionList
...
+    tmp_inters = il1.inters
+    cmaptorsion = typeof(il1.inters[1])
+    matrix = typeof(il1.data)
+    for inter in il2.inters
+        if isa(inter, CMAPTorsion)
+            push!(tmp_inters, cmaptorsion((4*tmp_inters[end].size*tmp_inters[end].size)+tmp_inters[end].index, inter.size, inter.λ, inter.res_num, inter.res_id))
+        elseif isa(inter, CMAPTorsion_L)
```
jgreener64, line 532: `CMAPTorsion_L` is undefined, this code is not run anywhere?

src/interactions/lennard_jones.jl
```diff
@@ -955,3 +1168,138 @@ end
...
+        l1.σ14_mixed + l2.σ14_mixed,
+        l1.ϵ14_mixed + l2.ϵ14_mixed,
+        l1.weight_14 + l2.weight_14,
+        l1.α + l2.α,
+        l1.λ_mixing,
+        l1.scheduler,
+    )
+end
+
+function to_lambda_function(inter::LennardJones14, ::GapsysSoftCore; α=0.85, λ_mixing=MinimumMixing(),
```
jgreener64, line 1206: Is this called anywhere?

src/free_energy/alchemy.jl
```diff
@@ -1,122 +1,292 @@
...
+    QuartersLambdaScheduler,
+    EleScaledLambdaScheduler
+
+# Alchemical Roles
 const AlchemicalRole = Int32
 
-const CoreRole::AlchemicalRole   = Int32(0)
-const InsertRole::AlchemicalRole = Int32(1)
-const DeleteRole::AlchemicalRole = Int32(2)
+const EnvRole::AlchemicalRole    = Int32(0)
```
jgreener64, line 14: The meaning of 0 has changed from `CoreRole` to `EnvRole`. Possibly okay but the default for an `Atom` is still `CoreRole`, is this correct? And just to check, the intention is to ignore lambda for `CoreRole` atoms with dual topology? A table outlining the different roles in the docs would be useful.

src/free_energy/alchemy.jl
```diff
@@ -1,122 +1,292 @@
...
+Lambda scheduler for relative free energy calculations, following GROMACS.
+
+The electrostatics and sterics are scaled linearly, as in [`LinearLambdaScheduler`](@ref). With
+[`RelativeFESystem`](@ref), PME is replaced by the two grid scheme of GROMACS, where the grids of
+the two end states are mixed with the electrostatic coupling.
+
+See [`DefaultLambdaScheduler`](@ref) for the keyword arguments.
+"""
+@kwdef struct GROMACSLambdaRBFEScheduler
+    dual::Bool = true
```
jgreener64, line 120: The schedule types all carry similar fields. Maybe one type with a field to distinguish them, e.g. an empty `type=GROMACSLambdaRBFEScheduler()`, would be cleaner.

src/free_energy/alchemical_system.jl
```diff
@@ -0,0 +1,739 @@
...
+                        temp = 298.0u"K", 
+                        units=true,
+                        scheduler=DefaultLambdaScheduler(dual=true),
+                        loggers=(),
+                        array_type=Array,
+                        float_type=Float32,
+                        LJsoftcore="gapsys",
+                        Csoftcore="gapsys"
+                        )
+    # Collect correct datatypes and fill in missing gaps in the mappings
```
jgreener64, line 74: There is code re-used here with the relative version, can anything be brought out into separate functions? Or just use one internal big function with branching?

src/free_energy/alchemical_system.jl
```diff
@@ -0,0 +1,739 @@
...
+The mapped atoms are fully coupled at `global_λ = 0` and fully decoupled at `global_λ = 1`.
+The atom order of `sys` is kept. Interactions are replaced by their λ-dependent versions and
+random velocities are generated at `temp`.
+
+# Arguments
+- `sys`: the [`System`](@ref) containing the atoms to decouple.
+- `global_λ`: the global λ of the system, from 0 (coupled) to 1 (decoupled).
+- `mapping`: a vector with the indices of the atoms to decouple.
+- `temp=298.0u"K"`: the temperature used to generate random velocities.
+- `units=true`: whether to use Unitful quantities.
```
jgreener64, line 48: Infer this from the system.

src/interactions/periodic_torsion.jl
```diff
@@ -125,3 +126,158 @@ end
...
+    return PeriodicTorsionλ(periodicities=tuplejoin(periodicities_A, periodicities_B), 
+                            phases=tuplejoin(phases_A, phases_B), ks=tuplejoin(ks_A,ks_B), proper=interB.proper, 
+                            λ_mixing=λ_mixing, scheduler=scheduler)
+end
+
+
+function update_lambda_function(existing_lambda::PeriodicTorsionλ, interB::PeriodicTorsion)
+    return PeriodicTorsionλ(periodicities=tuplejoin(existing_lambda.periodicities[1:(end/2)], interB.periodicities), 
+                          phases=tuplejoin(existing_lambda.phases[1:(end/2)], interB.phases), 
+                          ks=tuplejoin(existing_lambda.ks[1:(end/2)], interB.ks),
```
jgreener64, line 223: `1:(length(...) ÷ 2)` avoids a `Float64` range.

test/free_energy.jl
```diff
@@ -0,0 +1,654 @@
...
+    new_coords = wrap_coords.(new_coords, (sys.boundary,))
+
+    test_sys = System(sys,
+                    coords=new_coords,)
+    place_virtual_sites!(test_sys)
+    neighbors = find_neighbors(test_sys)
+
+    # --- Basic Benchmark ---
+    bench_result = @benchmark potential_energy($test_sys, $neighbors; n_threads=1)
+    # @test bench_result.allocs <= 11
```
jgreener64, line 120: Remove if not passing.

test/free_energy.jl
```diff
@@ -0,0 +1,654 @@
+@testset "OpenFE TYK2 comparison" begin
```
jgreener64, line 1: This file repeats the same block five times. Call in a `for` loop and only read the coordinates once.

jgreener64 (commented):
Looks great, this is a lot of work and adds loads of functionality. I left many comments, but only a few will take significant time.

A few general points:
- Many combinations are now possible, e.g. {absolute, relative} × {dual, single} × {8 schedulers} × {4 LJ soft cores} × {4 Coulomb soft cores}, but only one is tested thoroughly. It would be good to add at least end-state tests for other combinations, i.e. test that `potential_energy(sysA) == potential_energy(RelativeFESystem(sysA, sysA, λ, ...))` for lambda 0 and 1.
- A few of the interactions copy blocks of code to the lambda versions. In general we don't want to repeat the physics like that for maintainability reasons, so code can be extracted into a function and called twice.
- Constraints don't seem to work, they are not forwarded in `RelativeFESystem` (and neither is topology). Also, rebuilding the `eligible` and `special` matrices doesn't take account of the constraints being removed from the bond lists. It would be good to have a test with constraints, even just an end-point one, and to test that the `eligible` matrices and constraints match the original system after alchemical setup.
- It would be good to do a GPU benchmark on a non-alchemical system on master v this branch, just to check that performance did not regress.

Project.toml
ERovers: Reverted back to masters `Project.toml`

data/force_fields/benzene.xml
ERovers: Generated with openff-2.2.1 have added it in the info section of the xml file.
Commit: `295facd7`

src/bias/cv.jl
ERovers: Done, added a test similar to `CalcTorsion` for `CalcAngle`
Commit: `cacd23ad`

src/free_energy/alchemical_system.jl
ERovers: Agreed, have adjusted it and included a test to error to let user know if any option is not available and which are available.
Commit: `b80a0dfd`

src/free_energy/alchemical_system.jl
ERovers: Yes, done
Commit: `b80a0dfd`

src/free_energy/alchemical_system.jl
ERovers: Yess, have copied similar logic from `setup.jl`, but taking sysA as reference for which neighbourfinder.
Commit: `273fa7` and `b80a0dfd`

src/free_energy/virial_scaling.jl
ERovers: Have added it to `alchemy.jl` at the end and set up the virial scaling up based on whether an specific inter lists interaction is bonded `AlchemicalBondedInteraction` or pairwise interaction where the pairwise interactions are skipped, since these are always lambda scaled and therefore don't have to be scaled twice.
Commit: `e0803aae`

src/analysis.jl
ERovers: Have removed it and are using `EnsembleSystem` and `read_frame!` in the examples now.
Commit: `e0803aae`

src/loggers.jl
ERovers: Changed to `append_warn=true`
Commit: `e0803aae`

src/simulators.jl
ERovers: Done
Commit: `e0803aae`

src/units.jl
ERovers: It's still used for the single topology, which carries a tuple of parameters.

src/types.jl
ERovers: Done
Commit: `05e6df31`

src/types.jl
ERovers: Done, it was still refering to older version of `CMAP_torsion_L` that I needed for other project.
Commit: `05e6df31`

src/mixing.jl
ERovers: Don't think would affect performance, since atoms are Structs. Have done this and updated `MinimumMixing()` to extract the lambdas before mixing.
Commit: `a043c76e`

src/force_field.jl
ERovers: Done
Commit: `05e6df31`

src/force_field.jl
ERovers: Done
Commit: `05e6df31`

src/setup.jl
ERovers: Done
Commit: `05e6df31`

src/setup.jl
ERovers: Done
Commit: `05e6df31`

src/interactions/urey_bradley.jl
ERovers: Great catch, hadn't updated the not frequently used functions yet, is now done.
Commit: `fe0fcffc`

src/free_energy/alchemical_system.jl
ERovers: Yeah fair
Commit: `b80a0dfd`

src/free_energy/alchemical_system.jl
ERovers: Commit: `b80a0dfd`

src/free_energy/scale_dual.jl
ERovers: Yeah that was a mistake
Commit: `6683d9c0`

src/types.jl
ERovers: Was my old function for other project is no removed from this branch.
Commit: `05e6df31`

src/interactions/lennard_jones.jl
ERovers: Yes, it should be hit when a reference system is used that has been parametrized with CHARMM forcefield.
But have adjusted the functions and added more softcore options since. 
Commit: `23a53580`

src/free_energy/alchemy.jl
ERovers: Yes, the default should be EnvRole, because indeed CoreRole is an atom that changes parameters. In single topology this means the atom struct has a tuple, in dual topology this means there is an atom (with sys A parameters) and virtual atom (with sys B parameters). I have adjusted anywhere where initially CoreRole was used to describe environment atoms. I have also included a table in the documentation.
Commits: `05e6df31`, `b4856972`, `72ecddbe`

src/free_energy/alchemy.jl
ERovers: Yess makes sense, have implemented that.
Commit: `6683d9c0`

src/free_energy/alchemical_system.jl
ERovers: Yes, I have split up what was possible to split up, but there are quite big differences, which were left in the functions itself. Let me know if this is okay, or whether to it one function with dispatching or branching.
Commit: `b80a0dfd`

src/free_energy/alchemical_system.jl
ERovers: Done
Commit: `b80a0dfd`

src/interactions/periodic_torsion.jl
ERovers: Done
Commit: `fe0fcffc`

src/simulators.jl
ERovers: Done
Commit: `e0803aae` and `f10df0`

src/simulators.jl
ERovers: Yes great idea, have removed that here and added to `loggers.jl`.
Commit: `e0803aae' and 'f10df0'

src/types.jl
ERovers: Done, and adjusted the `to_device(::Tuple)` since it was missing `AT`.
Commit: `e0803aae` and `f10df0`

src/free_energy/alchemical_system.jl
ERovers: Hadn't adjusted it as I was focussing on the `AbsoluteFESystem`, but now correctly implemented `dA`/`dB` for chain ids.
Commit: `b80a0dfd`

src/interactions/cmap_torsion.jl
ERovers: CMAPTorsion is completely redone to minimize the code (based on the general comment), so think this is solved now.
Commit: `fe0fcffc`

src/interactions/cmap_torsion.jl
ERovers: CMAPTorsion had still a lot things from the other project, hence there is some stale code, should have been updated now, see comment above.
Commit: `fe0fcffc`

src/free_energy/alchemical_system.jl
ERovers: Redid completely, rather than taking it from the interaction list, taking the excluded and special lists of neighbourfinder of both system A and B and iterate through those and add to matrix. Then add the exclusion between alchemical A and alchemical B. Had to go this route since the constraints in system A and system B stripped the bonds from the specific_inter_lists, so can't use the same logic as `setup.jl`.
Commit: `b80a0dfd`

